PR #12, opened 2017-06-02
v0.2.0

piotr-oles:
**Changes**:
* tsconfig.json and tslint.json path are not printed anymore.
* `watch` option is not used on 'build' mode
* Handle case with no options object (`new ForkTsCheckerWebpacPlugin()`)
* Basic integration tests (along  units)

**Breaking changes**:
  * tslint is not enabled by default - you have to set `tslint: true` or `tslint: './path/to/tslint.json'` to enable it.
  * `blockEmit` option is removed - it choose automatically - blocks always on 'build' mode, never on 'watch' mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ npm-debug.log*
 # Coverage directory used by tools like istanbul
 coverage
 
+# Tmp directory for integration test
+tmp
+
 # Dependency directories
 node_modules
 jspm_packages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## v0.2.0
+ * tsconfig.json and tslint.json path are not printed anymore.
+ * `watch` option is not used on 'build' mode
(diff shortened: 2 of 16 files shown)


johnnyreilly:
This is brilliant, well done! Just curious, what sets `this.isWatching` to `true` or `false`? Read the code but it's not clear to me. 

Thanks for all your work by the way - I really appreciate it! I guess I should go through the examples on ts-loader and strip out the `block-emit` stuff?

johnnyreilly:
Made changes with https://github.com/TypeStrong/ts-loader/commit/96a8113a696cdda02ebe1607c07b0233ae89e5cd

I've also "promoted" fork-ts-checker to near the start of the readme.  I figure people are interested in fast builds 😉

piotr-oles:
`isWatching` is set to `false` on event `run`, and set to `true` on event `watch-run` :) Unfortunately, I can't find any flag in webpack`s compiler and compilation object that could tell me if webpack is watching or not :/
Thanks for promotion, I really appreciate this :)

Issue #17, opened 2017-06-06
Thoughts on "isolatedModules": true and ts-loader


johnnyreilly:
Hey!

An idea just occurred to me which might improve performance still when using this plugin with ts-loader. 

I've just been pondering usage of `"isolatedModules": true,` in relation to using ts-loader alongside fork-ts-checker-webpack-plugin.  Essentially using this option will speed up compilation times.  Which is nice.

However, if we set this option in the tsconfig.json then presumably fork-ts-checker-webpack-plugin will be impacted by this and effectively cease to typecheck.  So that's not good.  

Would you consider / would it make sense to have fork-ts-checker-webpack-plugin set `isolatedModules` to `false` regardless of what it says in the `tsconfig.json`?  If that were the case then we could advise people using ts-loader with fork-ts-checker-webpack-plugin to use `"isolatedModules": true,` in their `tsconfig.json` for an extra perf boost.  

What do you think?

cc @jamesbrantly

PS this is a badly documented compiler flag - there's a nice blog post on it though: http://weblogs.thinktecture.com/thomas/2016/05/tired-of-waiting-for-typescript-compilation.html

piotr-oles:
Hi! :)
Thanks for this proposal, I think it's a good idea. Could you create PR for this? I think you will have to add some checks in `lib/IncrementalChecker.js:29`, it shouldn't be hard.

johnnyreilly:
Sure - as soon as I get a moment I'll take a look!

piotr-oles:
Released with 0.2.2 ;)

johnnyreilly:
Sweet!

odensc:
What does `transpileOnly` in ts-loader do different from `isolatedModules`? Or are they meant to work together?

johnnyreilly:
@odensc - meant to work together. See the readme of ts-loader for `transpileOnly` details. Essentially it should just give a slight perf boost to ts-loader.

odensc:
Ah, thanks.

I updated to `0.2.2` and enabled `isolatedModules` - however I'm getting errors such as:

```ts
ERROR at /home/oden/Dev/JS/fttv/node_modules/@types/babel-core/index.d.ts(16,19):
TS1205: Cannot re-export a type when the '--isolatedModules' flag is provided.
```

If `isolatedModules` is overriden to false, shouldn't I not be getting errors related to it being enabled?

johnnyreilly:
[...]

PR #18, opened 2017-06-07
Made isolatedModules always false

johnnyreilly:
so it can be set to true in the tsconfig.json without affecting typechecking.  I think it's as simple as this - do correct me if I'm wrong!

Resolves https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/17

BTW your build is *nix specific (the `NODE_ENV=test ` in the `package.json`).  I haven't changed it but you could make it cross platform easily with [cross-env](https://www.npmjs.com/package/cross-env)

### DIFF
--- a/lib/IncrementalChecker.js
+++ b/lib/IncrementalChecker.js
@@ -28,7 +28,8 @@ module.exports = IncrementalChecker;
 
 IncrementalChecker.loadProgramConfig = function (configFile) {
   return ts.parseJsonConfigFileContent(
-    ts.readConfigFile(configFile, ts.sys.readFile).config,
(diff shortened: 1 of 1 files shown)


piotr-oles (approved):

piotr-oles:
Well done, thank you for your work :)

I will add cross-env but not sure when - currently I'm a little bit busy :/

johnnyreilly:
> currently I'm a little bit busy :/

Believe me I understand 😄 BTW I've started the blog post - about half written so far.  I'm planning to do one post that first covers using fork-ts-checker-webpack-plugin and then moves on to add happypack into the mix as well. (again using fork-ts-checker-webpack-plugin)

PR #28, opened 2017-07-08
Service is killed when webpack watch is done

### LINKED ISSUE #27: karma-webpack causes false-positive "watch" mode which causes ForkTSCheckerWebpackPlugin to never exit
We use karma-webpack to run our tests. It accepts a `singleRun` argument in its options object but even when set to true it uses `webpack-dev-middleware` which always starts webpack via a call to `watch()`. This produces a false-positive where [this line](https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/lib/index.js#L132) thinks webpack is watching (because the `watch-run` event *is* emitted) when it's effectively not. The end result is even after the tests are finished ForkTSCheckerWebpackPlugin stays open because it ignores the `done` event and never receives an `exit` event.

Now, I think the clean solution is that that `singleRun` argument for karma-webpack should *not* end up making a call to `watch()` for webpack. That being said, that sounds like a fairly intensive refactor to some fairly frequently used code. Would it be possible to have a `singleRun` option that's accepted by ForkTsCheckerWebpackPlugin that either A) ignores the `isWatching` flag if `singleRun: true` or B) causes the `isWatching` flag to always stay `false` when `singleRun` is `true`? I can get a PR together really shortly if this is amenable as this is basically a 2 line change either way.

Love the plugin so far as it's improved the speed of our build a ton. Thanks!

trash:
This fixes #27 by making sure the service is killed when the watch
task emits the "watch-close" event.
This fix only works with webpack >= 2.3.0 otherwise it's effectively a
noop because the event will never be emitted.

### DIFF
--- a/test/integration/index.spec.js
+++ b/test/integration/index.spec.js
@@ -8,8 +8,11 @@ var ForkTsCheckerWebpackPlugin = require('../../lib/index');
 
 describe('[INTEGRATION] index', function () {
   this.timeout(30000);
+  var plugin;
 
   function createCompiler(options) {
+    plugin = new ForkTsCheckerWebpackPlugin(Object.assign({}, options, { silent: true }));
+
     return webpack({
       context: path.resolve(__dirname, './project'),
       entry: './src/index.ts',
@@ -29,9 +32,18 @@ describe('[INTEGRATION] index', function () {
         ]
       },
       plugins: [
-        new ForkTsCheckerWebpackPlugin(Object.assign({}, options, { silent: true }))
+        plugin
       ]
     });
+  }
+
+  /**
+   * Implicitly check whether killService was called by checking that
+   * the service property was set to undefined.
+   * @returns [boolean] true if killService was called
+   */
+  function killServiceWasCalled() {
+    return plugin.service === undefined;
   }
 
   it('should allow to pass no options', function () {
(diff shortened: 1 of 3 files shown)


test/integration/index.spec.js
```diff
@@ -71,6 +82,18 @@ describe('[INTEGRATION] index', function () {
     });
   });
 
+  it('kills the service when the watch is done', function (done) {
```
trash, line 123: This test successfully fails without the addition of the `watch-close` callback

johnnyreilly:
Great to see this! Do you know why the test is failing?

piotr-oles:
The test fails because of timeouts... It looks like travis is very busy now. I will increase timeouts in the another pr to make build less non-deterministic

piotr-oles:
Please, pull from the master branch - I have fixed problem with tests timeout

piotr-oles:
We should also increase webpack dependency version to 2.3.0 - lower versions are not fully compatible with this plugin - it will work but with the bug. And user still can install it and use with lower version but npm/yarn will show him a warning - it looks like proper behavior :)
Files to change: `package.json` and `README.md`.

johnnyreilly:
Sounds good!

trash:
Rebased against master, updated the webpack peer dependency version, and it looks like the CI is passing.

piotr-oles:
Thank you for your work, I will publish it as v0.2.7 :)

trash:
@piotr-oles Would it be possible to get 0.2.7 published today? I was hoping to get an issue relying on this change resolved before the weekend. Thanks so much if that's possible :)

piotr-oles:
Sorry, I am not be able to deploy today. I will do it tomorrow I hope ;)

piotr-oles:
@trash - Just released as v0.2.7 :)

johnnyreilly:
Great!

trash:
Fantastic. Thanks!

PR #29, opened 2017-07-08
Fix - Make CI build more deterministic

piotr-oles:
## Changes
 * Move test/unit/index.spec.js to test/integration/index.spec.js - it was not an unit test
 * Increase integration tests timeout to 30 seconds (sometimes travis is very busy)
 * Remove unused NODE_ENV=test - make it more compatible with Windows platform


piotr-oles:
Should fix timeout issue in #28

Issue #25, opened 2017-06-29
Support for create-react-app, typed css-modules and some other stuff...


zinserjan:
Thank you for your work on this plugin. This approach improves webpack build times immense, especially incremental times are a lot faster (even in blocking mode).

But there are some issues at the moment that keeps us from using this plugin in the typescript fork of CRA:

 - All type checking errors/warnings needs to be forwarded to webpack’s compile lifecycle, cause CRA controls the terminal output (auto clear) and shows errors in the browser.
	-> watch mode needs to block webpack compilation until type checking is done
- File watcher should be configured by default to get the performance improvements 
- Displayed type errors should look similar to ts-loader

Additionally to these, I would also like to see support for typed css modules. The main issue here is that the type checking can only be done after the css files were processed cause type definitions for css files will be created on the fly.

I tried to fork this plugin and fix these „issues“. Adding the blocking of the emit phase back in watch mode was quite easy, but adding support for typed css modules was more complex. That’s why I started to work on a POC for this and ended up with [a new plugin](https://github.com/zinserjan/ts-checker-webpack-plugin), written in TypeScript and tested by Jest. 

The approach is the same like your plugin, but there are some differences:

1. The type checking starts after all files were built by webpack (seal hook), but the compilation still needs to be packed together and some other magic needs to be done to optimize the bundle. 
2. All files are cached by default until they will be invalidated by webpack. This is possible cause we have the information which files were (re-)built, due to first point. 
3. Webpack’s built in file watcher is used to watch additional files like type definitions. So theres no need for a secondary file watcher with it’s own configuration.
4. Only files that are used by webpack will be type checked & linted. This is also possible cause of 1.
5. Compilation is always blocked until type checking is done. In my test project type checking was in the most cases faster than webpack on incremental builds.
6. There is no support of type checking cancellation, cause hacks like [checking for file existence](https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/lib/CancellationToken.js#L40) are too expensive due to the I/O usage (didn’t checked if this was really called every 10ms). This is also superfluous due to the blocking. 
7. There are different error formatters for type checking errors (like ts-loader, code frame, …)
8. There is only a single thread for type checking. In my test cases I never saw a benefit on incremental times with more than 1 thread. Question is how many modules needs to be checked to see a benefit for multiple processes here? And the other question is, can we configure this automatically so that the user just defines a maximal number of processes and we optimize the used processes under the hood.
[...]

PR #31, opened 2017-07-08
Add diagnostics/lints formatters

piotr-oles:
I was inspired by #25 - so here we have support for formatters - to make this plugin more extensible and customizable :)


johnnyreilly:
Nice!

PR #38, opened 2017-08-11
Fix log message when tslint is on.

matt-tock:
I just noticed this log message is incorrectly flipped. Here's the fix.


piotr-oles:
Thank you :)

matt-tock:
No problem. ;)

Issue #39, opened 2017-08-14
Add flag to allow fork-ts-checker-webpack-plugin to check for syntax errors (HappyPack / thread-loader)


johnnyreilly:
Excuse the minimal post; on my phone. An issue was logged against ts-loader; details can be found here:

https://github.com/TypeStrong/ts-loader/issues/602

Essentially this comes down to ts-loader being unable to report errors to webpack when used in `happyPackMode` ie ts-loader is used with HappyPack or thread-loader to split work over multiple instances

johnnyreilly:
copied from linked issue in ts-loader:

I think the issue is here: https://github.com/TypeStrong/ts-loader/blob/master/src/index.ts#L214

```ts
    // _module.errors is not available inside happypack - see https://github.com/TypeStrong/ts-loader/issues/336
    if (!instance.loaderOptions.happyPackMode) {
        registerWebpackErrors(
            loader._module.errors,
            formatErrors(diagnostics, instance.loaderOptions, instance.compiler, { module: loader._module })
        );
    }
```

In `happyPackMode` we cannot register transpile errors with webpack and so this info is lost. BTW the same issue would presumably apply if using thread-loader instead of happypack.  Same principle; ts-loader cannot register errors with webpack.

There's nothing that we can do from a ts-loader perspective here.  

>  I think it should not be hard to add flag to fork-ts-checker plugin to check syntax errors.

If you could look at this that would be awesome!

abirmingham:
Currently we are seeing an issue where a number of errors are not reported by ts-loader and are not reported by fork-ts-checker-webpack-plugin. This is fairly concerning :)

Examples include:
```js
// Note the missing comma below:
const bar = [
    'a',
    'b'
    'c'
];

// Note the semi-colons below:
const biz = {
    a: 1,
    b: 2;
    c: 3;
    d: 4
};

// Note the missing '=>' below:
export const foo = async (bar:number):Promise<void> {
   ...
}
```

@johnnyreilly would this issue cause the behavior we're seeing? Thanks!

johnnyreilly:
I think so yes @abirmingham. If you drop usage of thread-loader / cache-loader and stop using happypack mode for the time being the issues will be raised as you'd hope.

In the ts-loader issue @piotr-oles said:

[...]

PR #43, opened 2017-08-30
Speculative fix for #39

johnnyreilly:
See details in https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/39

### DIFF
--- a/test/integration/index.spec.js
+++ b/test/integration/index.spec.js
@@ -191,12 +191,12 @@ describe('[INTEGRATION] index', function () {
     var compiler = createCompiler({ checkSyntacticErrors: false }, true);
     
     compiler.run(function(error, stats) {
-      expect(stats.compilation.errors.length).to.be.equal(2);
+      expect(stats.compilation.errors.length).to.be.equal(1);
       callback();
     });
   });
 
-  it('should not find syntactic errors when checkSyntacticErrors is true', function (callback) {
+  it('should find syntactic errors when checkSyntacticErrors is true', function (callback) {
     var compiler = createCompiler({ checkSyntacticErrors: true }, true);
     
     compiler.run(function(error, stats) {
--- a/test/integration/index.spec.js
+++ b/test/integration/index.spec.js
@@ -193,7 +193,7 @@ describe('[INTEGRATION] index', function () {
     var compiler = createCompiler({ checkSyntacticErrors: false }, true);
     
     compiler.run(function(error, stats) {
-      expect(stats.compilation.errors.length).to.be.equal(3);
+      expect(stats.compilation.errors.length).to.be.equal(2);
       callback();
     });
   });
@@ -202,7 +202,7 @@ describe('[INTEGRATION] index', function () {
     var compiler = createCompiler({ checkSyntacticErrors: true }, true);
     
     compiler.run(function(error, stats) {
-      expect(stats.compilation.errors.length).to.be.equal(3);
+      expect(stats.compilation.errors.length).to.be.equal(2);
       callback();
     });
   });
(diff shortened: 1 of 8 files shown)


piotr-oles:
Great, thank you for your work! Just one thing is missing - could you add an integration test for it? :)

johnnyreilly:
I have an idea for an integration test for this but it looks like tests don't run on Windows? Is that right?

piotr-oles:
Oh, it's possible, I didn't run tests on Windows :/ Could you fix it? I think `NODE_ENV=test` is not used so you can remove it :) Also you can remove `node_modules/.bin/` prefix because npm/yarn will use this path automatically :)

johnnyreilly:
No cigar :sad:  I've committed my speculative test - let's see what happens

johnnyreilly:
The only way I can run this is by committing and letting Travis run it. The logic in my test is not quite correct - can you take a look please?  About to have to dash out the door

test/integration/index.spec.js
```diff
@@ -185,4 +188,22 @@ describe('[INTEGRATION] index', function () {
       createCompiler({ tslint: true });
     }).to.not.throw.error;
   });
+
+  it('should not find syntactic errors when checkSyntacticErrors is false', function (callback) {
+    var compiler = createCompiler({ checkSyntacticErrors: false }, true);
+    
+    compiler.run(function(error, stats) {
+      expect(stats.compilation.errors.length).to.be.equal(2);
```
piotr-oles: I think it should be 1 :)

test/integration/index.spec.js
```diff
@@ -185,4 +186,22 @@ describe('[INTEGRATION] index', function () {
...
+      callback();
+    });
+  });
+
+  it('should not find syntactic errors when checkSyntacticErrors is true', function (callback) {
```
piotr-oles: It should be `should find` not `should not find` :)

piotr-oles (changes requested):

johnnyreilly:
Yeah I'd appreciate some help with the test please 🌷  I've got to work now and not being able to run this on Windows is killing me.  The test I have in place *should* work but doesn't. And somehow I seem to have broken the `should find the same errors on multi-process mode` test but I can't see why.

The actual functionality of the change is fine AFAIK - I'm just failing to implement a test correctly.

johnnyreilly:
Just implemented the changes requested.  I need help please. I'm leaving the house now and will hopefully be online again in about 3 hours....

test/integration/index.spec.js
johnnyreilly: done

test/integration/index.spec.js
johnnyreilly: quite right!

piotr-oles:
Oki, I will fix tests :)

johnnyreilly:
Thanks - I appreciate it 🌷

johnnyreilly:
Taken from the issue, @voodoorider found this:

> @johnnyreilly It worked, however after linking your fork in `thread-loader` example I discovered that the syntactical errors are reported even without providing the `checkSyntacticErrors` flag.

This makes me think that the flag is always being set to `true` regardless.  This would explain why one of my tests is failing also (and perhaps the other one as well?)

I may have misunderstood how the options get passed through. I'm away from my laptop and can't run the code; I have tried tweaking just in browser.  If someone else gets the chance to take a look that'd be awesome.

johnnyreilly:
OK it's just the pre-existing  `should find the same errors on multi-process mode` test that errors now.  Other problems fixed.

johnnyreilly:
OK I think we're green.  Feel free to merge!

piotr-oles (approved):

piotr-oles:
Thank you @johnnyreilly! Sorry I didn't help you with tests, I was very busy :/

johnnyreilly:
No worries - got there in the end! Do you know when it will be released?

Issue #40, opened 2017-08-23
Allow deferring plugin execution so loaders have time to generate files


Obi-Dann:
Hi,

We recently started using https://github.com/Jimdo/typings-for-css-modules-loader. It creates `d.ts` files for `css-modules` providing us with type checking and intellisense on css class names.
The problem is that the `d.ts` files are typically generated after `fork-ts-checker-webpack-plugin` is started. It means, the plugin either throws an error that a file is missing or checks an old version of `d.ts` file.

I suggest to add an async plugin hook `fork-ts-checker-service-before-start` so I can defer `fork-ts-checker-webpack-plugin` start up from another plugin.
If it sounds alright, I can create a PR implementing it.

I guess another option would be to add a flag `startAfterFinishModules` or `startAfterCompile` so the plugin will start type checking on compilation `finish-modules` or compiler `after-compile`

piotr-oles:
Hi :)
The first option sounds more elastic for me, so please implement this one if it's not a problem. 
PR's are always welcome :)

Obi-Dann:
Awesome, thanks, I've created a PR https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/44

piotr-oles:
Merged! I will release it as `0.2.8` :)

PR #44, opened 2017-08-31
Adding `fork-ts-checker-service-before-start` plugin to allow delaying service-start

Obi-Dann:
See #40 for details


piotr-oles:
Great work, thank you! :)

Issue #45, opened 2017-09-03
Port to TypeScript?


johnnyreilly:
Hello!

I've always wondered why this isn't actually written in TypeScript (given that's what it provides checking for).  I'm currently hacking locally on porting this to TypeScript.  For the most part I just want to see how hard it would be to do (not hard I suspect).  

Would you be interested in a PR for this assuming all goes well?

johnnyreilly:
Also, there might me a minor perf gains as I can kill some `bind`s (notoriously slow)

piotr-oles:
I wrote it in a pure javascript to keep things simple :) It's easier to debug and develop small plugin in javascript :) But as it becomes bigger and currently is stable (I think), rewriting it to TypeScript could be a good step forward to maintain growing complexity of the plugin :)

PR #46, opened 2017-09-03
Initial TypeScript port

johnnyreilly:
This is not ready for merging yet but I thought you'd like to see the first steps on #45... (Also I'd like to see if the tests still pass - still can't run on Windows!)


johnnyreilly:
The tests pass on first attempt - hah!  Check me out 😉 

I did not expect it to work first time.  Quite pleased it did.  I'd planned to remove the `lib` directory and leave only the `src` directory committed. (The `lib` directory is generated from the `src` directory now.)  What do you think?

piotr-oles:
You can remove it but you have to add `prepublish` script in `package.json` to build lib before release. Also you will have to update tests to compile before run and add `lib` to `.gitignore` :)

johnnyreilly:
Cool - doing it

johnnyreilly:
Done

johnnyreilly:
OK - I've made it `noImplicitAny: true` compliant (can't stand an `any` 😄).  Feel free to merge!

src/NormalizedMessage.ts
```diff
@@ -0,0 +1,212 @@
...
+      return 1;
+    }
+
+    return (
+      NormalizedMessage.compareTypes(messageA.getType(), messageB.getType()) ||
+      NormalizedMessage.compareOptionalStrings(messageA.getFile(), messageB.getFile()) ||
+      NormalizedMessage.compareSeverities(messageA.getSeverity(), messageB.getSeverity()) ||
+      NormalizedMessage.compareNumbers(messageA.getLine(), messageB.getLine()) ||
+      NormalizedMessage.compareNumbers(messageA.getCharacter(), messageB.getCharacter()) ||
+      // code can be string (lint failure) or number (typescript error) - should the following line cater for this in some way?
```
johnnyreilly, line 92: I think this might be a bug?  I've left it as is just in case.

johnnyreilly: Line 93 I mean - 92 is the comment I put in relating to this...

piotr-oles:
Ok, I've checked it and looks good :) I will not release it now because it doesn't add a new functionality. I plan to add errors limit option to avoid out of memory on big projects refactoring, so it will be released then :)

johnnyreilly:
Great!

PR #48, opened 2017-09-08
Fix clicking on URLs in the terminal (for VS Code users)

johnnyreilly:
**EDIT: I misunderstood the original problem as I did this PR.  I've left most of what I've written in place but please do read all the way to the end.  It will all make a little more sense then I promise 😄**

The motivation behind this PR has very little to do with getting reported errors closer to ts-loader's "style".  (I honestly have no opinion on the style as is.) This PR essentially flows from my addiction to using VS Code as my editor.  Let me explain.  At present the formatted output of an error looks like this:

```
ERROR at C:/work/fx-online/src/FxOnline.App/src/containers/deposits/account/GiveNotice/index.tsx(190,44):
no-use-before-declare: variable 'form' used before declaration
```

Which is great in terms of information.  However; there's a gotcha.  When using the integrated terminal in VS Code I like being able to click on a file in the terminal and navigate straight there.  The existing formatting in the `defaultFormatter` puts the parens and error location information in with the path itself.  And VS Code doesn't interpret it as a link. :cry: 

~~The change I've made moves the parens onto the next line.  This is more like what ts-loader does and also the `codeframeFormatter` too.  The new output looks like this:~~

~~ERROR in C:/work/fx-online/src/FxOnline.App/src/containers/deposits/account/GiveNotice/index.tsx
(190,44): no-use-before-declare: variable 'form' used before declaration~~

**EDIT: I misunderstood the reason this wasn't "clickable" - read on and you'll see more.**

This is clickable in VS Code :+1:  I've also changed " at " to " in " as I thought it was a little clearer.  If you prefer " at " I'm more than happy to revert that.  

My team is running with this fork locally now (`yarn link`ing our way to glory)

What do you think about this?  I'd really like to get this merged.  More generally I think there could be some work done to allow people to supply their own formatters but in a future PR perhaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,9 @@ If `false`, disables built-in colors in logger messages. Default: `true`.
 * **logger** `object`:
 Logger instance. It should be object that implements method: `error`, `warn`, `info`. Default: `console`.
 
-* **formatter** `'default' | 'codeframe' | Function`:
+* **formatter** `'default' | 'codeframe' | ((message: NormalizedMessage, useColors: boolean) => string)`:
 Formatter for diagnostics and lints. By default uses `default` formatter. You can also pass your own formatter as a function
-(see `lib/NormalizedMessage.js` and `lib/formatter/` for api reference).
+(see `src/NormalizedMessage.js` and `src/formatter/` for api reference).
 
 * **formatterOptions** `object`:
 Options passed to formatters (currently only `codeframe` - see [available options](https://www.npmjs.com/package/babel-code-frame#options))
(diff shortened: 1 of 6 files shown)


johnnyreilly:
FWIW I realise now that it's possible to supply your own formatter to the plugin (which is awesome).  I hadn't grokked that before and so I've updated the documentation to make it a little clearer.

johnnyreilly:
It's just occurred to me that the colouring in the formatters might be what's causing the problem. Let me experiment and report back before you merge this...

johnnyreilly:
Yeah - the actual issue with not being able to click the URLs was the colours.  The formatting used to switch colours partway through the message; as it moved from the filename to the error line / column.  That's hard to understand; this should clarify it:

```
ERROR at the/path/index.tsx(190,44):
                            ^ Colour change here!
```

This colour change screwed up the link clickability which is what got me looking at this in the first place.  

Anyway, the fix was simple; just keep the same colour for the filename, line number and column number.  This fixes the issue completely; files open when you click on them **and** to the exact position of the lint / error in question!  🎆 🍾 

So I think this is good to merge now - sorry for my original misdiagnosis of the issue.

johnnyreilly:
Details of a workaround for those waiting on this PR being merged: https://blog.johnnyreilly.com/2017/09/fork-ts-checker-webpack-plugin-code.html

piotr-oles:
Thank you for your work :) Will be released in the v0.2.9

johnnyreilly:
Thanks @piotr-oles - I appreciate that!

PR #52, opened 2017-09-15
fixes #50

### LINKED ISSUE #50: Differences in watch mode and non-watch mode
Hi!

Thank you for the cool plugin. It would be nice, if the README could mention some differences when this plugin is used in watch mode (`webpack(options).watch`) and non-watch mode (`webpack(options).run`).

E.g. we use `new ForkTsCheckerWebpackPlugin({ silent: true })`. This will log errors in non-watch mode, but it will not log errors in watch mode. To log errors in both cases with webpacks native logging we need to set `new ForkTsCheckerWebpackPlugin({ silent: true, async: false })`. Are there more differences?

donaldpipowitch:
PR to fix #50 😄


johnnyreilly:
Nice!

piotr-oles:
Thank you for your work! :)

Issue #61, opened 2017-11-01
Parameter to define which files should be linted


dreef3:
Hi,

Thanks for a great plugin! The issue I have is the following:

All files in the `program` are passed to TSLint for linting. However the program contains not only the files of the application but also any dependencies from node_modules. In my case I have some dependencies with *.ts files which are imported by application.

And while I definitely need them for compilation & type checking I'd prefer not to run TSLint on them.

Do you think it'be possible to have a param like `skipLint` which would accept string[] of regexp for files which should not be linted? By default it'd filter out node_modules.

DorianGrey:
At least the ability to kick of the `node_modules` directory would be awesome - there are several cases where people are publishing their `ts` source files along with their transpiled counterparts to be used directly, or they referencing them in provided typings, so that they get picked up.

However, it should be documented that using to many filters here might cause this to become quite expensive.

Rough suggestion: Replace

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/39d7986fe5c9a110aa21858d873540b05c106fce/src/IncrementalChecker.ts#L181-L183

with something like
```typescript
// From configuration; these entries should be present by default
const skipped = [
        /\.d\.ts$/, // Replaces the current "endsWith" check
        /(\\|\/)node_modules(\\|\/)/ // Beware of path separator differences Windows / rest of the world
];
const filesToLint = this.files
        .keys()
        .filter(filePath =>
           !this.files.getData(filePath).linted && !skipped.some(s => s.test(filePath))
        ); 
```

mdudek:
I'm facing same issue. I need to reference some *.ts files from node_modules, but some of these files are not tslint compliant. So i can't build my project without tslint errors at this moment.

There is an option in tslint.json that could be used for excluding of files from linting, but this setting is not respected by fork-ts-checker-webpack-plugin:

`
"linterOptions": {
    "exclude": [
      "node_modules/**/*.ts"
    ]
}
`

DorianGrey:
That would be even better than a custom solution.

This option in the `tslint.json` currently seem to only be respected automatically when using the `tslint` CLI. Users of the `Linter` API have to do that manually.

johnnyreilly:
Why not submit a PR? I've had a good level of success with PRs being taken by this fine project.

DorianGrey:
Sure, I'll see if can get some spare time during the next days to work on this. Should not be that hard.

[...]

PR #62, opened 2017-11-04
Correctly handle linterOptions.exclude from tslint config, if present (#61)

DorianGrey:
Refers to #61 :

Tslint currently supports particular linter options to be set in the `tslint.json`. These are handled automatically when using the CLI, but most of them have to be handled manually when using the `Linter` API.
The parameter for manually excluding files is `linterOptions.exclude` (description can be found in the [docs](https://github.com/palantir/tslint/blob/77191671fef18cfe6fcdd5d817b006161259c2d9/docs/usage/configuration/index.md)). Excluding files is sometimes required, e.g. in case a module provides type definitions that refer to plain `ts` files or when using `ts` source files from those modules directly - in both cases, those might conform to different linter settings, and thus might cause the linting process to fail.

This PR adds some code to pick up this config entry if present, and filter out files that should not be linted according to the provided patterns. If follows the approach that `tslint` itself uses:

https://github.com/palantir/tslint/blob/5634a074f72bd82854bc5439a3a27388b05b64d3/src/runner.ts#L193-L196
https://github.com/palantir/tslint/blob/5634a074f72bd82854bc5439a3a27388b05b64d3/src/runner.ts#L223-L229

Uses the same glob matching library as tslint, the dependency is now just explicitly listed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "@types/lodash.isfunction": "^3.0.3",
     "@types/lodash.isstring": "^4.0.3",
     "@types/lodash.startswith": "^4.2.3",
+    "@types/minimatch": "^3.0.1",
     "@types/node": "^8.0.26",
     "@types/webpack": "^3.0.10",
     "chai": "^3.5.0",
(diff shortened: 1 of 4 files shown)


johnnyreilly (approved):
This looks good to me @piotroles - what do you think to merging and releasing this?

johnnyreilly:
^ @piotr-oles I meant 😄

piotr-oles (approved):

piotr-oles:
@DorianGrey - thank you for your work, it will be released in a next version :)

johnnyreilly:
Awesome! Do you know when you're planning to release?

PR #64, opened 2017-11-07
Fix deploying

piotr-oles:

PR #77, opened 2017-11-29
Adds .vue functionality

prograhammer:
Hi guys, 

This PR adds ability to work with `.vue` files (see https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/70).  

For a quick test, I created a `vue` branch (out of the `feature/vue` branch) with the built `lib` folder included. So you can install like this:  

`npm install git://github.com/prograhammer/fork-ts-checker-webpack-plugin.git#vue --save-dev
`
1. Turn on the vue option in the plugin in your webpack config:

```
      tslint: true,
      vue: true
    })
```

2. For linting to work in `.vue` files, you need to ensure your script tag's language attribute is set
to `ts` or `tsx` (also make sure you include the `.vue` extension in all your import statements as shown below): 

```html
<script lang="ts">
import Hello from '@/components/hello.vue'

// ...

</script>
```

3. If you are testing in Webpack, (in addition to this plugin) you'll need something like this in your rules:

```
{
  test: /\.ts$/,
  loader: 'ts-loader',
  include: [resolve('src'), resolve('test')],
  options: {
    appendTsSuffixTo: [/\.vue$/],
    transpileOnly: true
  }
},
{
  test: /\.vue$/,
  loader: 'vue-loader',
  options: vueLoaderConfig
},
```
4. I also currently use [tslint-config-standard](https://github.com/blakeembrey/tslint-config-standard) so my `tslint.json` looks something like:  

```json
{
    "defaultSeverity": "error",
    "extends": [
      "tslint-config-standard"
    ]
}
```
5. Ensure your `tsconfig.json` doesn't exclude .vue files:

```
// tsconfig.json
{
  "include": [
    "src/**/*.ts",
    "src/**/*.vue"
  ],
  "exclude": [
      "node_modules"
  ],
  "compilerOptions": {
    
    // ...

    "baseUrl": ".",
    "paths": {
      "@/*": [
        "src/*"
      ]
    }
  }
}
```

If you are working in **VSCode**, you'll need extensions [Vetur](https://marketplace.visualstudio.com/items?itemName=octref.vetur) and [TSLint Vue](https://marketplace.visualstudio.com/items?itemName=prograhammer.tslint-vue) (a forked extension which I also currently maintain) and the editor will match the output you get from this `fork-ts-checker-webpack-plugin`.

Let me know what more you need or if you have any suggestions on code improvement/cleanup/etc.

Thanks!


johnnyreilly:
First of all: good work! What I say should be taken with a pinch of salt as I'm not a maintainer of this plugin.

That aside, I notice that with the current approach the code moves to always bring Vue into the mix. I'd be tempted to refactor the code to only use the Vue approach when you are definitely using the Vue functionality. Essentially treating the Vue functionality as something that can be feature flagged on or off. Do you understand what I mean? (I may not be being super clear)

prograhammer:
@johnnyreilly Yeah I was thinking about that too. Do you think a flag to turn it on via the plugin options would be the way to go?

johnnyreilly:
I think that might be an idea (unless it can be reliably inferred somehow - possibly from the presence of `'vue'` in `compiler.options.resolve.extensions`?)

src/IncrementalChecker.ts
```diff
@@ -78,8 +95,27 @@ class IncrementalChecker {
     return tslint.Configuration.loadConfigurationFromPath(configFile) as ConfigurationFile;
   }
 
+  /**
+  * Fixes a non-ts module's name so the file and source text can be properly found later.
+  * Assumes wildcard "@" is [project root]/src.
+  */
+  static resolveNonTsModuleName(moduleName: string, containingFile: string, programConfigFile: string): string {
+    if (moduleName.indexOf('@/') === 0) {
```
Toilal: Instead of hardcoded values, both webpack configuration ([resolve.alias](https://webpack.js.org/configuration/resolve/#resolve-alias), [resolve.modules](https://webpack.js.org/configuration/resolve/#resolve-modules)) and tsconfig parameters ([baseUrl and paths option](https://www.typescriptlang.org/docs/handbook/compiler-options.html)) could be exploited here.

src/IncrementalChecker.ts
```diff
@@ -9,6 +9,7 @@ import WorkSet = require('./WorkSet');
 import NormalizedMessage = require('./NormalizedMessage');
 import CancellationToken = require('./CancellationToken');
 import minimatch = require('minimatch');
+import vueParser = require('vue-parser');
```
Toilal: It would be better to use [vue-template-compiler](https://www.npmjs.com/package/vue-template-compiler) parseComponent function. Return type is SFCDescriptor, see flow typings https://github.com/vuejs/vue/blob/dev/flow/compiler.js#L172-L177

Toilal:
I don't think we can rely on `compiler.options.resolve.extensions` because most ts-loader users enable `appendTsSuffixTo` to make `*.vue` files looks like `*.ts` files. (https://github.com/TypeStrong/ts-loader#appendtsxsuffixto-regexp-default)

johnnyreilly:
Fair point - maybe an explicit flag is better

Toilal:
Maybe we can check vue-loader is available by trying to simply trying to require it ...

prograhammer:
@Toilal That's a pretty clever idea (smart detection).

How do you guys feel about me moving all of this code into a separate module called `CreateVueProgram.ts`. I'll restore the code back to the way it was before the PR and just use a condition to decide whether to call `CreateVueProgram` here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/src/IncrementalChecker.ts#L161

There's no reason to be super DRY here. We can let Vue be its own thing and go it's own path by taking care of it in it's own module. Leave the normal program as it was in IncrementalChecker before I touched it. Then I'll start applying all those suggestions you made throughout the code @Toilal.

Thoughts?
Cc @johnnyreilly

johnnyreilly:
Sounds promising!

prograhammer:
Alright guys,

Everything works great on my example project on my local machine because the project is a Vue project with `vue-loader` in it's node_modules. You can see I added the condition [here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/77/files#diff-e27b56838bc2968a5affe095bbd08efeR173). But Travis fails because I imagine there's no `vue-loader` in the project it is testing with. This is a nice approach, but what should we do about Travis? Or should we do something else instead, like check for an option from the plugin?

Update: Nevermind! **facepalm** Thanks @Toilal

src/IncrementalChecker.ts
```diff
@@ -128,6 +130,16 @@ class IncrementalChecker {
     return endsWith(filePath, '.d.ts') || linterExclusions.some(matcher => matcher.match(filePath));
   }
 
+  /**
+   * Check for vue-loader in user's project node_modules.
+   */
+  hasVueLoader(): boolean {
+    const projectPath = path.dirname(this.programConfigFile);
+    const contextResolution = Resolve.sync('vue-loader', { basedir: projectPath });
```
Toilal: It seems you still have to try/catch this line and return false on error ...

src/IncrementalChecker.ts
```diff
@@ -128,6 +130,16 @@ class IncrementalChecker {
...
+    
+    return require(contextResolution) ? true : false;
```
Toilal: It seems you still have to try/catch this line and return false on error ...

prograhammer: Oh right! Brain freeze! Of course that's the problem. LOL. Will fix...

prograhammer:
voila! @Toilal !

All checks passing now. I'll merge master into it (and fix any conflicts) after all feedback is in. Now regarding feedback, I looked over webpack resolve and so far I don't really see any API documentation around that for using it programmatically. Also, for `vue-template-compiler` there are no types defined. We just need a fast AST tree parser and that's what `vue-parser` is doing here (uses [parse5](https://github.com/inikulin/parse5) under the hood).

prograhammer:
Alright guys, there's a huge line of eager people waiting for this (there's over 1,300 people who have [TSLint Vue](https://github.com/prograhammer/vscode-tslint-vue) currently installed in their VSCode editor and those Vue folks are almost all guaranteed to be on Webpack). And there's no other option availabe for Vue developers that runs at the typechecker level. Since I've split off the code so it doesn't disrupt the original program, what can we do to fast-track this PR? I'll continue to help polish things up afterwards as well.

johnnyreilly:
@prograhammer the person who needs to feed back is @piotr-oles. This is his project and only he can merge I'm afraid. In the short term you could consider publishing the fork to a GitHub repo with the compiled `dist` included as well. We did that whilst we were waiting for 0.2.9 to be published. See here: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

Speaking of which, I must take that down now 0.2.9 has shipped.

johnnyreilly:
Background to how that came to pass here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/64

prograhammer:
Ah, gotcha. No worries!

Yeah I've got this branch here that I'll keep promoting to folks in the meantime: `npm install git://github.com/prograhammer/fork-ts-checker-webpack-plugin.git#vue --save-dev`

johnnyreilly:
Sweet!

src/IncrementalChecker.ts
```diff
@@ -134,20 +137,22 @@ class IncrementalChecker {
...
+      this.vueLoaderResolved = false;
     }
+
+    return this.vueLoaderResolved;
   }
 
   nextIteration() {
     if (!this.watcher) {
-      this.watcher = new FilesWatcher(this.watchPaths, ['.ts', '.tsx']);
+      this.watcher = new FilesWatcher(this.watchPaths, ['.ts', '.tsx', '.vue']);
```
johnnyreilly: Could the `.vue` be conditional based on whether vue is in the mix or not please?

src/IncrementalChecker.ts
```diff
@@ -128,9 +133,26 @@ class IncrementalChecker {
...
+      this.vueLoaderResolved = false;
+    }
+
+    return this.vueLoaderResolved;
+  }
+
   nextIteration() {
     if (!this.watcher) {
-      this.watcher = new FilesWatcher(this.watchPaths, ['.ts', '.tsx']);
+      this.watcher = new FilesWatcher(this.watchPaths, ['.ts', '.tsx', '.vue']);
```
johnnyreilly: Could including `vue` be conditional upon whether this is a `vue` program or not please?

src/IncrementalChecker.ts
```diff
@@ -158,7 +176,14 @@ class IncrementalChecker {
       }
     }
 
-    this.program = IncrementalChecker.createProgram(this.programConfig, this.files, this.watcher, this.program);
+    if (this.hasVueLoader()) {
+      this.programConfig = this.programConfig || VueProgram.loadProgramConfig(this.programConfigFile);
```
johnnyreilly: in this branch isn't it always the `VueProgram` you use?

johnnyreilly:
It's looking good @prograhammer - I've a few comments as you can see but it's basically good. I'm glad that this is mostly triggered only when Vue is in the mix.  

@piotr-oles would you be able to take a look at this when you get a mo?

src/IncrementalChecker.ts
prograhammer: Does it hurt to always include `.vue`? It doesn't really bother the folks not on Vue.js. Otherwise it's another condition we'll need to make here for the watcher. Thoughts?

src/IncrementalChecker.ts
prograhammer: No no, this condition is here to detect if `vue-loader` is found in the user's project and then use the VueProgram if it is. We don't need to maintain a separate branch for Vue if we have this condition. It can be merged into master.

johnnyreilly: `this.programConfig || VueProgram.loadProgramConfig(this.programConfigFile)` but wouldn't this statement always use `programConfig`? Or have I missed something?

src/IncrementalChecker.ts
johnnyreilly: 2 things really:

1. It will make a marginal speed difference for non-Vue users. I'm philosophically opposed to that :smile:
2. It bothers me on an OCD level that Vue files would be always in the mix when they needn't be. This is more subjective I grant you but makes me twitch in unhappy ways. It is my suffering.

src/IncrementalChecker.ts
prograhammer: The first pass over that line `this.programConfig` will be undefined so it will grab the return from the right side of the condition (`VueProgram.loadProgramConfig(...`). The next time it passes over that line `this.programConfig` will have a value and it will grab that instead of performing another unnecessary `VueProgram.loadProgramConfig(...`.

src/IncrementalChecker.ts
prograhammer: There's always a battle between code readability and micro-optimization (in this case the optimization may be significant). OCD kicks in for both :smile: 
But let me see of a clean way to condition that...

src/IncrementalChecker.ts
johnnyreilly: Cool - WFM!

src/IncrementalChecker.ts
johnnyreilly: Thanks man!

jthomaschewski:
Thanks for the effort @prograhammer 
I tried using your fork in a project of mine but I have problems with mixed JS/TS vue components.

When importing a `JSModule.vue` module with `<script>` from a `TSModule.vue` module with `<script lang="ts">` (or any .ts file)  I get:

```ts
ERROR in /app/TSModule.vue(2,27):
TS2306: File '/app/JSModule.vue' is not a module.
```

Same result regardless of `allowJs` setting in `tsconfig.json`.
No issues when running `tsc -p . --noEmit` or `ts-loader` with`transpileOnly: false`

Importing vue components with TS works fine from both plain `.ts` and ts-enabled `.vue` files


EDIT:
Importing such JS vue modules does work when using absolute paths, e.g `@/components/JSModule.vue`. But I would expect this to also work with relative paths as I have the following declaration file for Vue:

```ts
declare module '*.vue' {
  import Vue from 'vue'
  export default Vue
}
```

CKGrafico:
Please accept that!

prograhammer:
@jbbr A couple things to point out here:

1. You don't need that wildcard declaration (`declare module '*.vue' ...`). That defeats the purpose of this PR and TypeScript. You won't get full type information from your Vue modules if you do that.
2. You need to ensure your scripts have `lang="ts"` or `lang="tsx"` for them to be parsed. I just updated the code to allow `js` and `jsx` as well, but that doesn't seem like your problem.
3. The code works for relative paths too. Maybe something else is going on. The best way to know is to give us an example project repo on Github we can pull down and see the problem.

prograhammer:
@johnnyreilly All your feedback is all done brother, :smile_cat:

jthomaschewski:
@prograhammer thanks for the feedback.

1. I know that this is not necessary for webpack but it's required if you rely on typescript language server in IDEs. e.g in IntelliJ/WebStorm but also in VSCode (outside of .vue files/Vetur).
Typings are still useful within a vue file. E.g usage of internal props, data, methods, external services etc. is typed properly.
If there is a better way to tell typescript to recognize .vue modules - Please tell me! I'd prefer a better solution.
The microsoft typescript-vue-starter repo does also [contain this shim](https://github.com/Microsoft/TypeScript-Vue-Starter/blob/master/src/vue-shims.d.ts)

2. I have `lang="ts"` in my .vue files with TS and just a plain `<script>` tag in .vue files with JS (should be equal to `lang="js"`).
I still have problems importing `.vue`-files with `lang=js` from within `.vue`-files with `lang=ts`.

Example repos:
- Working example without ForkTSChecker: https://github.com/jbbr/typescript-fork-ts-vue-test/tree/master
- Broken example with ForkTSChecker: https://github.com/jbbr/typescript-fork-ts-vue-test/tree/fork-ts-checker-1
(Breaks simply by adding the plugin, no other changes, see https://github.com/jbbr/typescript-fork-ts-vue-test/commit/609492d35f3756f2668099b4569e14fbaa539beb )
- Broken example with absolute paths: https://github.com/jbbr/typescript-fork-ts-vue-test/tree/fork-ts-checker-2
(For some reason using absolute paths didn't helped this time. Maybe it's related to the recent changes in your fork - at least the behavior is consistent now )

prograhammer:
Hey @jbbr 

(numbered list again :smile: )

1. For your editor (VSCode), yeah in TS modules, but this is going away. See here: https://github.com/prograhammer/vscode-tslint-vue/issues/4. This is not related to this plugin though, as you mention.  

2. Pull down the latest (`npm install git://github.com/prograhammer/fork-ts-checker-webpack-plugin.git#vue --save-dev`) and you'll have `lang="js"` working. Using `<script>` without lang attribute **will NOT work**.

I'll take a look at those repos you provided later today and see what's up with your paths. Thanks.

johnnyreilly:
@piotr-oles would you be able to review this please?

jthomaschewski:
@prograhammer Thanks!

1. I think this is kind of related to this plugin:
Typescript allows type definitions which provide type information for any kind of module - without depending on the typescript or webpack module resolution. E.g I could just force typescript to accept `*.random` files without trying to resolve them:
```ts
declare module "*.random" {
  const content: number
  export default content
}
```
Now if I do an `import mod from 'test.random'` I would expect typescript to not throw an error and have the `Number` prototype available on `mod`.
Of course webpack would throw an error but for typescript the module should just exist.

This is the behavior of plain `ts-loader` with type checking but also of `fork-ts-checker` without vue as it replicates the behavior of typescript.

This PR seems to just ignore the `declare module` typings for `.vue` files so it always tries to resolve them by it's own logic (which is fine - but I would prefer it to not throw errors for missing .vue files, so having the behavior of `vue-loader+ts-loader`'s type checking) - webpack will take care of module resolution and throw "Module not found" errors.

Not sure what would be the "right" way. It's just what I would expect - maybe the behavior of `vue-loader+ts-loader` is "wrong" and this is how it should be :smile:

2. Thank you, adding `lang="js"` fixes the error indeed. Would it be possible to support `<script>` without lang attribute as well? That's the default in the official vue documentation and many people in mixed JS/TS environments are using this


btw: I don't want to delay the merge of this PR. I really appreciate the work and it does work. So even if the behavior with mixed TS/JS is not perfect, it IS working perfectly for TS-only vue projects.

prograhammer:
@jbbr Actually what you may not realize is that it's ignoring those `.vue` imports. With this PR (and if you activate it by having `vue-loader` in your project) you will have to have those vue imports exist or it will throw an error. This is ideal.

Regarding the use of `<script>` without `lang`, I look at that as the way developers could opt-out of parsing. I need to think about that a little more. Maybe consider this after the PR is accepted. Thinking about it though.

aweiu:
Waiting for merging!

aweiu:
@prograhammer If the Vue file does not contain any script tags, the error of "*.vue'is not a module." will be reported

aweiu:
@prograhammer Thanks!I have updated it just now,but the problem remains

> If the Vue file does not contain any script tags, the error of "*.vue'is not a module." will be reported

```
 error  in /Users/aweiu/Documents/工作/yun-game-admin/src/router/index.ts

(4,23): File '/Users/aweiu/Documents/工作/yun-game-admin/src/components/Developer.vue' is not a module.

 error  in /Users/aweiu/Documents/工作/yun-game-admin/src/router/index.ts

(5,21): File '/Users/aweiu/Documents/工作/yun-game-admin/src/components/Channel.vue' is not a module.
```

prograhammer:
Ah, @aweiu, for like components that are just template/style. Ok, no biggie, let me look into that. Obviously you could just put an empty script tag in there and that's probably what you are doing now. But not ideal.

(the "add parent nodes flag" update was for another small issue related to tslint)

aweiu:
@prograhammer Yes, after all, there are some static pages that do not really need script

johnnyreilly:
Hi all,

fork-ts-checker-webpack-plugin takes some time to merge and ship. Now this is open source and it’s @piotr-oles's project which he does in his spare time. We can’t insist he does any more work than he does.

But it’s a pain waiting. I recognise that. To mitigate that I've created this repo: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

It exists to be a **built** version of the fork-ts-checker-webpack-plugin with prospective changes that have not yet been merged but that people hope will. 

Hopefully we can use this as an interim step whilst we wait for merge / release by @piotr-oles.   @piotr-oles  I hope this is cool with you; I wanted to not nag you but also create a way for the new hotness to get used!  I've given @prograhammer access to that - I'll try and do the same for you.  Do say if this is an issue.

src/IncrementalChecker.ts
```diff
@@ -128,9 +133,29 @@ class IncrementalChecker {
     return endsWith(filePath, '.d.ts') || linterExclusions.some(matcher => matcher.match(filePath));
   }
 
+  /**
+   * Check for vue-loader in user's project node_modules.
+   */
+  hasVueLoader(): boolean {
+    if (this.vueLoaderResolved !== undefined) return this.vueLoaderResolved;
```
piotr-oles: Please add curly braces to follow code style in the project :) Would be good to add prettier to this project in a feature

src/IncrementalChecker.ts
```diff
@@ -128,9 +133,29 @@ class IncrementalChecker {
...
+    }
+
+    return this.vueLoaderResolved;
+  }
+
   nextIteration() {
     if (!this.watcher) {
-      this.watcher = new FilesWatcher(this.watchPaths, ['.ts', '.tsx']);
+      const extensions = ['.ts', '.tsx'];
+      if (this.hasVueLoader()) extensions.push('.vue');
```
piotr-oles: same here :)

src/VueProgram.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+  * Fixes a non-ts module's name so the file and source text can be properly found later.
+  * Essentially just resolves the module's name from a relative or wildcard path into an absolute path.
+  * Assumes wildcard "@" is [location of tsconfig]/src.
+  */
+  static resolveNonTsModuleName(moduleName: string, containingFile: string, basedir: string): string {
+    if (moduleName.indexOf('@/') === 0) {
+      moduleName = basedir + '/src' + moduleName.substr(1);
```
piotr-oles: Is this always true that `@/...` means `src/...`? Is this hardcoded vue logic or something that we can configure?

src/VueProgram.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    else if (moduleName.indexOf('./') === 0 || moduleName.indexOf('../') === 0) {
```
piotr-oles: please follow project code style and write `else if` next to the closing curly brace :)

src/IncrementalChecker.ts
```diff
@@ -128,9 +133,29 @@ class IncrementalChecker {
...
   }
 
+  /**
+   * Check for vue-loader in user's project node_modules.
+   */
+  hasVueLoader(): boolean {
+    if (this.vueLoaderResolved !== undefined) return this.vueLoaderResolved;
+    
+    try {
+      Resolve.sync('vue-loader', { basedir: this.basedir });
```
piotr-oles: do we need `Resolve` to test if vue-loader is installed? Maybe we could use built-in node functionality?
```
try {
    require.resolve("vue-loader");
    this.vueLoaderResolved = true;
} catch (e) {
    this.vueLoaderResolved = false;
}
```

piotr-oles (changes requested):

piotr-oles:
@prograhammer Thank you for your work, I really appreciate that!

I've added some comments to the PR and I will release this plugin as `0.3.0-alpha`. There will be info in README.md for vue developers that they can use that version. 

I think I will keep it as an alpha for about 1 month and if there will be no big issues I will release it as `0.3.0`. It's because it will run in many different environments and I want to be sure that updating fork-ts-checker-plugin will not break builds for non-vue developers :)

johnnyreilly:
Thanks for the review @piotr-oles - that's great!  You should now have received an invite to TypeStrong as well.  Hope you're having a lovely December 8th! :christmas_tree:  :wink:

piotr-oles:
@johnnyreilly Thank you for the invitation, I'm glad to join the TypeStrong ;)
@prograhammer Please add functional tests for it :)

prograhammer:
Hey @piotr-oles @johnnyreilly and anyone else following this PR,

I'll apply your feedback @piotr-oles and add testing. But first, I'm once again very torn. I'm encountering more benefits to using a .ts extension such as `.vue.ts` or `.v.ts` (and they just work so smoothly in Webpack in Vue). So what I'm thinking is that we offer an option in the plugin:

```
{
    // ...
    vue: true  // <-- defaults to using .vue.ts (the preferred extension)
}
```

and to change the extension:

```
{
    // ...
    vue: '.vue'  // to keep others happy (but list consequences in README)
}
```
This also means we can do away with checking for vue-loader (makes for easier testing I suppose). 

**Now what is the recommended way to pull in options** into `IncrementalChecker` or `VueProgram`? Should I put the vue option in node's `process.env` (like it is done [here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/src/index.ts#L309))?

johnnyreilly:
Cc @herringtondarkholme

johnnyreilly:
I defer to @piotr-oles when it comes to desirable architecture - what say you chap?

aweiu:
The Vue file without the script tag will report the error of "*.vue'is not a module."
For me, this is perfect if it is solved!

HerringtonDarkholme:
Thank @prograhammer for bring this! I would suggest an architecture similar to language service plugin, where custom logic can be extracted outside of core. That said, a webpack plugin that can load other ts-plugin might be desirable. @piotr-oles 

Other than that, Vue team also has a service host implementation. Hope it will be helpful. https://github.com/vuejs/vetur/blob/master/server/src/modes/script/serviceHost.ts

prograhammer:
Issues I've noticed with `.vue` extension:
1. HMR (hot module reload) seems to act a little strange. If you change a module, sometimes the data gets reverted back to the default and requires you to refresh the browser page. @aweiu Did you notice this? I may have to upload a test example project to demonstrate. But, ultimately, I think it's not a deal-breaker at the moment. (with `.vue.ts` this problem doesn't happen)
2. In `fork-ts-checker-webpack-plugin` it works great, but in VSCode editor (with Vetur) using `.vue` doesn't allow for intellisense if the `.vue` module is imported into a `.ts` module (ie. `main.ts` or `test.spec.ts`). Again, not a deal-breaker since most people will import `.vue` into another `.vue` or import `.ts` into `.vue` instead of vice-versa. @aweiu What editor/extension are you using to TSLint your project?

So I guess we will stay with `.vue` as default extension and offer option to change. I'm going to get all this work done tonight and hopefully we can push this PR through in the next few days. Thanks!

prograhammer:
Can I squash commits while I'm in PR (I've never tried during a PR)? Or is that something only done when the PR is finally merged by the owner with a `squash and merge` merge?

johnnyreilly:
@prograhammer - I don't know. I guess it doesn't matter as long as you can do it at the end..

aweiu:
@prograhammer 
Only use VSCode with  TSlint Vue plugin.
In webpack,i have closed all TSlint and waiting for this PR.

`HMR (hot module reload) seems to act a little strange. If you change a module, sometimes the data gets reverted back to the default and requires you to refresh the browser page. @aweiu Did you notice this? I may have to upload a test example project to demonstrate. But, ultimately, I think it's not a deal-breaker at the moment. (with .vue.ts this problem doesn't happen)`

I've met this problem too without any TSlint loader.So, I think that might be the result of other plugins

src/VueProgram.ts
Toilal: :+1: It should be grabbed from webpack configuration ([resolve.alias](https://webpack.js.org/configuration/resolve/#resolve-alias)).

See vue webpack template

https://github.com/vuejs-templates/webpack/blob/cc89353be444eee88e7e6480bf186a3cf5bd439f/template/build/webpack.base.conf.js#L34-L42

prograhammer:
@piotr-oles @johnnyreilly 

I've applied all the feedback (linting, unit/integration tests, etc.). Awaiting your approval!

For all who are following this PR, you'll need to turn on the vue option in the plugin:

```
      tslint: true,
      vue: true
    })
```

@aweiu I've also fixed the bug you reported. Reinstall and let me know if it works.

Thanks!

.travis.yml
```diff
@@ -4,7 +4,7 @@ deploy:
...
+    secure: J4q7IGixHsEp5of9FG/Zefit8ba6bV7T1I0yV8TBNQHfwBaTC344PBhx3jRg8Nhv04vzkzY+s2cJYIuyqlfEeZrsBcFTvPE9LowR/4cMCoFjteb/gDGtW2AFw2M5pfJ6L4PHXtigO5emWs7xeT3R8KxcQn3e2nzJcGXShc0N9vCCDe601MKJFco5svoHqx5cAmHoknIQDPatddob9VdOUvaRGZWWh7Pv8pdDHGMcUk18TL67p55vMDXL8LvtHlOJpg106P0hX8a3NGhD0Oq9jGgmS77JwAOel36gncxou8CbEXlzpNsH65gJ2/tEJ3uSUZdHYQtQCiQtNzFbqI45JTu7UXkwgQv0Wxb7fnnNsz9StYmquVNp7MDw+Ol2/txA55tBBdSCeK+hy/X7cgI4tI/ftZ98c0YQbeUe/fdaV6PMxv0lPzED2We3N2Y080ineOib1ySsjY0diN2NM6NB8+9TJgqyLzd7X3WofJxExMFMRKuApCDCU7Bjw/FlKkISzvUOMmthL91LunQZ4x+ratiYVbz2cKa7PMcP0Qm7cFnosbyQKJWjM1KTOsCnoCS7iLZHxzKKE+prOWrssbcfENvpSTmouWrGppi9D8nwXDEFNyp+pVx8b7gwQ5GR/WjXb1PY/WkjBgNRaZbHzJlJzYTMBjNjlI1Nn0s4ncWhsZE=
```
johnnyreilly: Hey @prograhammer  

How come you changed the key?

johnnyreilly:
Apart from the API key this looks good to me.  What do you think @piotr-oles?

.travis.yml
prograhammer: Oh wow, no idea? I actually recently signed up with Travis with my own GitHub so maybe that's why? Let me put the original key back...

aweiu:
@prograhammer I just tested the latest version, and it's all right now！Thank you very much!
And,can it completely replace the tslint-loader?

Toilal:
Thanks for this work, it's great ! I just tested it and it works, it now detects TypeScript type errors. 

But I have small issues with TSLint.

- TSLint reports `Expected indentation of 0 spaces but found 2.` because of indentation inside <script> tag. Here's my vue file.

```<template>
  ...
</template>
<style>
  ...
</style>
<script lang="ts">
  import Vue from 'vue'
  
  export default Vue.extend({
    ...
  })
</script>
```

Workaround is to remove indentation inside `<script>` tag, but editor wants to bring it back ... Maybe we could process script by removing indents if all lines are indented ?

- TSLint reports errors on files that are not TypeScript when imported from a TypeScript file. In my case, I have linting errors on my bootstrap theme imported from main.ts entrypoint.

```
ERROR in /home/toilal/idea-projects/vue-tree/app/styles/bootstrap-theme.scss(1,1):
no-unused-expression: unused expression, expected an assignment or function call
ERROR in /home/toilal/idea-projects/vue-tree/app/styles/bootstrap-theme.scss(1,6):
whitespace: missing whitespace
ERROR in /home/toilal/idea-projects/vue-tree/app/styles/bootstrap-theme.scss(1,7):
whitespace: missing whitespace
ERROR in /home/toilal/idea-projects/vue-tree/app/styles/bootstrap-theme.scss(1,11):
whitespace: missing whitespace
ERROR in /home/toilal/idea-projects/vue-tree/app/styles/bootstrap-theme.scss(1,12):
whitespace: missing whitespace
```

(`import './styles/bootstrap-theme.scss'` in main.ts) => *update* Workaround is to use require instead of import ! Maybe it's expected behavior finally ...

- TSLint reports `eofline: file should end with a newline` with such vue files

```
<template>
  <div id="app">
    <router-view></router-view>
  </div>
</template>
<script lang="ts" src="./App.ts"></script>
```

Workaround is to add an EOL inside `<script>` tag (Note that `app.ts` already has an EOL and the end of file).

```
<template>
  <div id="app">
    <router-view></router-view>
  </div>
</template>
<script lang="ts" src="./App.ts">
</script>
```

aweiu:
@prograhammer 
If i modify the TS file, it is ok.
But,if i modify the VUE file,then:

wx20171219-161019

I have to restart webpack to fix it

Toilal:
Also, is it possible to run tslint against vue files without running the build or dev server ?

Toilal:
@prograhammer, I just created a pull request against this branch (https://github.com/prograhammer/fork-ts-checker-webpack-plugin/pull/1) that fixes indentation related issues. 

It also use vue-template-compiler instead of vue-parser lib, as it's the vue official library to parse SFC components and it includes the "deindent" magic that fixes the issue with identation.

Toilal:
About the API key mistery, it's because the fork was created before the key changed. You should have rebased this branch to master @prograhammer.

see https://github.com/Realytics/fork-ts-checker-webpack-plugin/commit/8a77948ae7ca271f8ff38d2dc86e431173533117

prograhammer:
Hey @aweiu 
Can you upload an example test project to github that I can clone and see this problem? This is a weird one I haven't caught yet. (And thanks so much for helping test this stuff out!) :smile: 

Hey @Toilal 
1. I would have loved to go with `vue-template-compiler`, but here our the reasons I had against it:
  -  It puts in blank lines and you'll get a `no-consecutive-blank-lines` linting error. So with `vue-parser` I put in slashes for commenting out lines. I also add `tslint:disable` and `tslint:enable` if possible (if I can fit it in without changing character count). I pay particular attention in `vue-parser` to not mess up character count (which would break editor auto-fixing). You can see [here](https://github.com/prograhammer/vue-parser/blob/master/src/index.ts#L36) that I do some funky stuff to preserve character count.
  - The de-indentation in `vue-template-compiler` seems like it would throw off the character count and character line position. There's really no need to consider the script tags themselves into the indentation. Regarding editors, in VSCode editor, the plugin [TSLint-Vue](https://marketplace.visualstudio.com/items?itemName=prograhammer.tslint-vue) doesn't have a problem with this either. 
-  I also lose a particular amount of control with `vue-template-compiler` because TypeScript has proved to be picky about things. For example, to fix `is not a module` problem when importing a Vue file that has no script content I had to adjust `vue-parser` to include an empty import [here](https://github.com/prograhammer/vue-parser/blob/master/src/index.ts#L27).
2. Since this is a Webpack plugin that programmatically uses `tslint`, then unfortunately there is no way to run `tslint` directly from the command line to lint Vue files. This is something that would have to be added to `tslint` or a fork of it (which would then require more active maintenance).  

Let me know if I've missed something here @Toilal and convince me more if we need to switch to `vue-template-compiler`. :smile_cat:

Toilal:
No problem @prograhammer, I understand this better now :+1: 

I'm currently trying to add `.vue` file support directly into tslint (https://github.com/palantir/tslint/pull/3596) through a plugin system (there no such thing at the moment ...), but it's a hard work and I hit same difficulties than your work on this pull request.

Those positions MUST be exact for many reasons, but `vue-template-compiler` deindents the script, so it breaks the positions ... Problem is that `vue-parser` keeps position, but it then throws false positives errors related to invalid indentation. It seems natural to indent script source inside `<script>` tag, tslint should handle this properly ...

When using `vue-template-compiler`, reported positions have an offset compared to original `.vue` file. So currently, none of the solution really works.

I hope i'll be able to find a perfect solution about that directly in my tslint fork and it's associated plugin.

Toilal:
So in VSCode, when you automatically format the code, the script is not indented inside `<script>` tag ? I should consider this as an IntelliJ Issue then.

Examples from VueJS docs are not indented too : https://fr.vuejs.org/v2/guide/single-file-components.html

(edit: I just found an option in intellij to configure disable indentation of children of `<script>` and `<style>` tags :smile: )

prograhammer:
When I use [TSLint Vue](https://marketplace.visualstudio.com/items?itemName=prograhammer.tslint-vue) plugin, it accepts the unindented content in the `<script>` tags in Vue files (because I built it on `vue-parser` also). Personally I don't really use auto formatting (although I might use the auto-fix :bulb: for a specific linting error). I just rely on the linting as I go. But the `Format Document` command will still work on the indentation correctly:

vscode-format-indent

However, there are other issues with Vetur and TypeScript (such as the spacing before function parenthesis). See more [here](https://github.com/vuejs/vetur/issues/476). It's something I hope to work out later sometime.

aweiu:
@prograhammer 
See here: https://github.com/aweiu/fork-ts-checker-webpack-plugin-test

prograhammer:
@aweiu 
Yeah, I'm getting this problem too. It seems it happens in `fork-ts-checker-webpack-plugin` alone. If you install the plugin normally, the problem persists. Hmmm...

prograhammer:
@aweiu 
It seems to be a problem with `fork-ts-checker-webpack-plugin` and not the Vue functionality in Pull Request. Now what you can do is convert all your Vue files to `lang='ts'` and also update your `tsconfig.json`:

```
  "include": [
    "src/**/*.ts",
    "src/**/*.vue"
  ],
  "exclude": [
      "node_modules"
  ],
```
However, I realize that some projects might have a mix of regular Vue scripts and Vue with TypeScript (or may even use Vue files imported from other libraries). Maybe @piotr-oles and @johnnyreilly can help us more here. If you guys pull down @aweiu repo: https://github.com/aweiu/fork-ts-checker-webpack-plugin-test and update the `package.json` with the master of `fork-ts-checker-webpack-plugin` you can see first hand the problem. Run `npm run dev` and then make a small modification to the `App.vue` template (add some text somewhere) and notice lots of errors pop up as things break.

prograhammer:
Ok @aweiu  I think I've figured it out, will update after dinner...

CKGrafico:
Like other people I'm waiting to @prograhammer but thanks for your hard work on this :)

prograhammer:
Thanks @CKGrafico , the PR is ready to go as soon as @piotr-oles let's it through into a `fork-ts-checker-webpack-plugin beta` release so that you can install it via npm.

@aweiu I fixed that problem (was a problem with parsing non-lang-ts scripts in Vue). Reinstall, and you'll see it doesn't break when modifying Vue files. Of course, eventually you should use TypeScript in your Vue files too (`lang="ts"`). Also, I noticed you forgot two things in your test repo. See [my PR there](https://github.com/aweiu/fork-ts-checker-webpack-plugin-test/pull/1).

jvbianchi:
@prograhammer In this PR, is a way to ignore a type of file (`*.graphql` for example)?

Because I [graphql-tag/loader](https://github.com/apollographql/graphql-tag#webpack-preprocessing) to import `graphql` files.

But with this PR I'm getting errors like:
```
ERROR in /Users/user/dev/app/graphql/update.user.graphql(19,9):
TS2304: Cannot find name '$id'.
ERROR in /Users/user/dev/app/graphql/update.user.graphql(20,5):
TS2304: Cannot find name 'name'.
```

I use this typings for typescript allow me to import graphql files, but this PR seems to be ignoring it:
```ts
declare module '*.graphql' {
  import { DocumentNode } from 'graphql';

  const GraphqlFile: DocumentNode;

  export default GraphqlFile;
}
```

aweiu:
@jvbianchi Me too!I need import some '[s]css' file

src/VueProgram.ts
```diff
@@ -0,0 +1,131 @@
...
+      path.dirname(configFile)
+    );
+
+    parsed.options.allowNonTsExtensions = true;
+
+    return parsed;
+  }
+
+  /**
+   * Since 99.9% of Vue projects use the wildcard '@/*', we only search for that in tsconfig CompilerOptions.paths.
```
piotr-oles, line 34: Please, add information about vue support and how it resolves "@/*" paths in the `README.md` :)

package.json
```diff
@@ -50,8 +50,10 @@
     "@types/lodash.startswith": "^4.2.3",
     "@types/minimatch": "^3.0.1",
     "@types/node": "^8.0.26",
+    "@types/resolve": "0.0.4",
     "@types/webpack": "^3.0.10",
     "chai": "^3.5.0",
+    "css-loader": "^0.28.7",
```
piotr-oles, line 56: do we need css-loader here?

piotr-oles (changes requested):

piotr-oles:
I've changed target branch to the `next` branch - it will be released as `fork-ts-checker-webpack-plugin@next` :) Also there is a `fork-ts-checker-webpack-plugin@latest` tag - latest master build :)

prograhammer:
Hi @piotr-oles 

I'll get to your feedback and also look at the issue reported by @jvbianchi later today. 

Thanks!

prograhammer:
Alright, solved those problems. But I need another day before I can do a clean up on the fixes and then I'll push them up.

johnnyreilly:
:+1:

package.json
prograhammer: Yeah, `css-loader` is for the Vue integration test. Since Vue uses single-file-components, I wanted to include a simple css section in the example files just to ensure everything is working. Also, I added an `import exampleCss from "./example.css"` to ensure other "non-vue" imports work (as reported by @aweiu @CKGrafico @jvbianchi).

src/VueProgram.ts
prograhammer: Added `vue section` to README. :smile:

prograhammer:
@johnnyreilly @aweiu @CKGrafico @jvbianchi  Install the test branch `npm install git://github.com/prograhammer/fork-ts-checker-webpack-plugin.git#vue --save-dev` and you should see those issues resolved. 

@aweiu Remember, you'll need a `.d.ts` file for importing non-ts. 

@piotr-oles All your feedback is applied. I'm not sure why Travis failed (as all tests pass on my local machine). Can you look into that? It looks like the timeout has been exceeded by Travis. :thinking:

Punit-wingify:
@prograhammer I am getting error "Property '$store' does not exist on type comp_name" however it works without using 'fork-ts-checker-webpack-plugin'. Can you please check if you are able to use vuex inside any local component.

prograhammer:
Hi @Punit-wingify 

I can confirm in my own project that I am able to access `this.$store.state.some.foo` and it works perfectly with this PR. Make sure you have the latest Vue and Vuex. You can [see here](https://github.com/vuejs/vuex/blob/dev/types/vue.d.ts#L16) that when you add Vuex to your project the Vue type interface is extended to include `$store: Store<any>`.

Punit-wingify:
@prograhammer Okay thanks, just needed the confirmation 👍

TheLarkInn:
Anything we [at webpack] can help with furthering this along. This is super exciting and we've heard a few downstream requests come our way in regards to its status. 

Happy to help in any way possible or unblock any blockers if they exist.

johnnyreilly:
Thanks @TheLarkInn - that's always appreciated! 

I think the current status is that @prograhammer has fixed all the issues we're aware of and the only thing we're waiting on is @piotr-oles to confirm he's happy to merge.  Exciting!
  
@piotr-oles are we ready to go now?

keenwon:
@prograhammer 

After turn on the vue option, I got a error:

```js
  vue: true
})
```

```js
{
  test: /\.ts$/,
  exclude: /node_modules|vue\/src/,
  loader: 'ts-loader',
  options: {
    transpileOnly: true,
    appendTsSuffixTo: [/\.vue$/]
  }
},
{
  test: /\.vue$/,
  loader: 'vue-loader',
  options: {
    esModule: true,
    loaders: {
      ts: 'ts-loader',
      scss: 'vue-style-loader!css-loader!sass-loader'
    }
  }
}
```

```shell
Starting type checking service...
Using 1 worker with 2048MB memory limit
webpack building...
D:\Code\nei7\node_modules\fork-ts-checker-webpack-plugin\lib\service.js:22
        throw error;
        ^

TypeError: Cannot read property 'flags' of undefined
    at getBaseTypes (D:\Code\nei7\node_modules\typescript\lib\typescript.js:28172:37)
    at getSuperClass (D:\Code\nei7\node_modules\typescript\lib\typescript.js:37548:21)
    at isPropertyDeclaredInAncestorClass (D:\Code\nei7\node_modules\typescript\lib\typescript.js:37537:29)
    at checkPropertyNotUsedBeforeDeclaration (D:\Code\nei7\node_modules\typescript\lib\typescript.js:37507:21)
    at checkPropertyAccessExpressionOrQualifiedName (D:\Code\nei7\node_modules\typescript\lib\typescript.js:37477:13)
    at checkPropertyAccessExpression (D:\Code\nei7\node_modules\typescript\lib\typescript.js:37448:20)
    at checkExpressionWorker (D:\Code\nei7\node_modules\typescript\lib\typescript.js:40582:28)
    at checkExpression (D:\Code\nei7\node_modules\typescript\lib\typescript.js:40529:42)
    at checkExpressionCached (D:\Code\nei7\node_modules\typescript\lib\typescript.js:40406:38)
    at checkVariableLikeDeclaration (D:\Code\nei7\node_modules\typescript\lib\typescript.js:42623:43)
```

prograhammer:
Hi @keenwon 

1. Instead of passing Vue files to ts-loader from vue-loader, try updating your vue-loader options to look like this: 

```
    loaders: {
      // ts: 'ts-loader',  <-- remove this
      scss: 'vue-style-loader!css-loader!sass-loader'
    }

```
2. And then you would need to update your `ts-loader` exclude statement (try an include statement as shown in the first post).

3. Are using this build: `npm install git://github.com/prograhammer/fork-ts-checker-webpack-plugin.git#vue --save-dev`?

4. Create a simplified example project (strip everything down, remove your biz logic or whatever) on your Github and give us a link to it here so we can diagnose the error first hand with what you are doing in the project.

piotr-oles (approved):

piotr-oles (approved):

piotr-oles:
Looks good, will be released as 0.3.0 :) Great work!

keenwon:
@prograhammer  this project can reproduce the problem.
https://github.com/keenwon/fork-ts-checker-webpack-plugin-test

prograhammer:
@keenwon 

It looks like you found a strange TypeScript bug which is not coming from `fork-ts-checker-webpack-plugin` (previous version or latest 0.3.0 with Vue functionality) but rather TypeScript itself (even the latest TypeScript version).  The reason you are seeing it when you turn on the `vue` flag is not due to `fork-ts-checker-webpack-plugin` but is because of that specific type of error inside the class (with the experimental decorator) in that Vue file...

For example, take this class with the error ` Block-scoped variable 'bar' used before its declaration`:

```
import Vue from 'vue'
import { Component, Prop, Watch } from 'vue-property-decorator'

@Component
export default class Test extends Vue {
  public foo: string = this.bar  // <-- This is an error that TypeScript can handle. Decorator may be the problem.
  private bar: string
}
```
and put it in a regular `.ts` file such as `test.ts` and import it into your main entry. Remove `fork-ts-checker-webpack-plugin` completely from your webpack config and then update `ts-loader` (remove the transpileOnly) to compile. You'll see TypeScript still breaks on that code. It even breaks the normal (non-Vue) TSLint extension in VSCode.

I'll report the issue to TypeScript (as well as look into it myself) and put the link here for anyone to follow. In the mean time, just fix that error in that class. Let us know if anything else breaks, thanks!

johnnyreilly:
Perhaps a silly question, but why not just swap the foo and bar lines? It seems like a legitimate TypeScript error to me...

prograhammer:
Yeah but it's an error TypeScript should be able to handle without breaking. It handles this error fine outside of experimental decorators (reports it as `Block-scoped variable 'bar' used before its declaration`). :cry: But all other errors seem to work fine in experimental decorators.

johnnyreilly:
Hmmm weird - definitely report it on the Typescript repo I reckon

Toilal:
I tried the new release with vue: true and tslint: true, and it works really well.

Only issue I found i that linter is running on empty script tag content, even when a `src` attribute is defined.

```vue
<template src="./App.html"></template>
<style scoped src="./App.css"></style>
<script lang="ts" src="./App.ts"></script>
```

This raises tslint errors for missing EOL

```
ERROR in C:/devel/projects/vue-ts-test/src/App.vue
(3,91): file should end with a newline
```

CKGrafico:
@Toilal can you upload your working example? in my nuxt project linting is still not working on vue files :/ 
About EOL I think that is a rule that has sense 4-5 years ago but now I usually disable it.

In my case
Working
`import SkipLink from './SkipLink.vue';`
Not working
`import SkipLink from '~/components/SkipLink.vue';`

But if I disable this plugin, everything is working

CKGrafico:
Wow ok my error is here... @prograhammer  https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/77/files#r159148429 haha '~' is super common to! (In Microsoft environments more) I'll made a PR for this :D

prograhammer:
@Toilal 

That's weird since vue-parser adds an EOL here: https://github.com/prograhammer/vue-parser/blob/master/src/index.ts#L27

Maybe it's complaining about the Vue file itself. Can you add an EOL to it and see what happens:

```
<template src="./App.html"></template>
<style scoped src="./App.css"></style>
<script lang="ts" src="./App.ts"></script>
<-- EOL here
```

prograhammer:
@CKGrafico Of course you know what the best solution is right? Switch to Ubuntu Gnome 3  for development (and dual boot with Windows for gaming/photoshop/etc.). :laughing: :laughing: :laughing:

CKGrafico:
Anybody needs something else?

Toilal:
I can't reproduce the EOL issue now ... Maybe after upgrading to tslint 5.9.1 ? (5.9.0 was broken at release, but shortly fixed)

Issue #80, opened 2017-12-15
Cannot read property 'getLineAndCharacterOfPosition' of undefined


nidu:
Hello. Just migrated to webpack 3 and trying to setup fast TS compilation, but stumbled on following error. Problem occurs for TS 2.6.x but not for e.g. 2.1.4. Any hints on what can be the cause? Maybe some broken dependency? If i proceed working and save files same error appears again and again.

Environment:

Windows 7x64
webpack 3.10.0
node 6.11.0
npm 3.10.7
typescript 2.6.2
fork-ts-checker-webpack-plugin 0.2.9
ts-loader 3.2.0

Plugin is created without options and here's ts-loader config:

```
{ 
  test: /\.tsx?$/, 
  loader: 'ts-loader', 
  options: {
    // disable type checker - we will use it in fork plugin
    transpileOnly: true,
    happyPackMode: true
  } 
}
```

Error:

```
Starting type checking service...
Using 1 worker with 2048MB memory limit
 67% building modules 982/1020 modules 38 active ...s\user-monitor3\UserStore.tsc:\nidu\dev\juno\juno\node_modules\fork-ts-checker-webpack-plugin\lib\service.js:22
        throw error;
        ^

TypeError: Cannot read property 'getLineAndCharacterOfPosition' of undefined
    at NormalizedMessage.createFromDiagnostic (c:\nidu\dev\juno\juno\node_modules\fork-ts-checker-webpack-plugin\lib\NormalizedMessage.js:15:39)
    at Array.map (native)
    at IncrementalChecker.getDiagnostics (c:\nidu\dev\juno\juno\node_modules\fork-ts-checker-webpack-plugin\lib\IncrementalChecker.js:125:58)
    at run (c:\nidu\dev\juno\juno\node_modules\fork-ts-checker-webpack-plugin\lib\service.js:13:31)
    at process.<anonymous> (c:\nidu\dev\juno\juno\node_modules\fork-ts-checker-webpack-plugin\lib\service.js:38:5)
    at emitTwo (events.js:106:13)
    at process.emit (events.js:191:7)
    at process.nextTick (internal/child_process.js:758:12)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
 94% asset optimization
```

johnnyreilly:
I'd suggest clearing your node modules and reinstalling - btw you didn't share your plugin setup..

[...]

PR #81, opened 2017-12-19
Check if diagnostic.file is defined in NormalizedMessage.createFromDi…

nidu:
…agnostic

It seems some compiler errors can be global and don't belong to specific file, thus don't contain file data and requesting `diagnostic.file` may return `undefined`. This request adds check for such case.

Fixes issue #80


johnnyreilly:
Thanks @nidu! @piotr-oles what do you think?

piotr-oles:
@nidu Thank you for the PR :)

PR #92, opened 2018-01-26
Support jsx in vue

wonderful-panda:
Currently, this plugin can't compile vue SFC with lang other than "ts" or "js"

- If lang is "tsx" or "jsx", compilation errors are raised.
- if lang is not specfied, content of <script> is ignored. 

This PR will fix this problem.


johnnyreilly:
What do you think @piotr-oles?

piotr-oles:
Thank you for your work and sorry for delay :) Merged :)

PR #94, opened 2018-01-31
Improve module resolution of *.vue

wonderful-panda:
Now, module resolution implementation of *.vue is in `VueProgram.resolveNonTsModuleName()`.
But this implementation is not same as TypeScript's one, so in some cases it can't resolve modules.

For example:
- TypeScript allows importing modules by path from compilerOptions.baseUrl, but this doesn't.
- TypeScript allows defining path mappings by compilerOptions.paths, but this doesn't
  (This will be partially supported by #86)

Concrete examples are in `test/integration/vue/imports/Test.vue`
(compilerOptions is in `test/integration/vue/tsconfig-imports.json`)

Document about TypeScript's module resolution is [here]( https://www.typescriptlang.org/docs/handbook/module-resolution.html)

This PR will be improve module resolution of *.vue.
This just delegates it to TypeScript's module resolution, so *.vue will be resolved as well as ts modules are resolved

I guess #88 may be fixed by this  PR.

PR #98, opened 2018-02-06
Fix small mistake in #94 (maybe potential problem)

wonderful-panda:
I've found a small mistake in my previous PR (#94)


johnnyreilly:
Hey @piotr-oles - I've merged in the changes for #94 into #93. The builds now passing again.

PR #100, opened 2018-02-23
Support overriding typescript version

### LINKED ISSUE #99: Support overriding typescript version
I need to be able to override the typescript version.
It would be great if compiler could be specified in options just as with ts-loader

Pajn:
fixes #99


src/CancellationToken.ts
```diff
@@ -2,25 +2,28 @@ import crypto = require('crypto');
...
   cancellationFileName: string;
 }
 
 class CancellationToken {
+  ts: typeof tsTypes;
   isCancelled: boolean;
   cancellationFileName: string;
   lastCancellationCheckTime: number;
-  constructor(cancellationFileName: string, isCancelled: boolean) {
+  constructor(ts: typeof tsTypes, cancellationFileName: string, isCancelled: boolean) {
```
piotr-oles: I would suggest to stay with `ts` as types namespace and change name from `ts` to `compiler` for a compiler instance.

piotr-oles (changes requested):

src/CancellationToken.ts
Pajn: I followed the pattern in NormalizedMessage, but I have no problem changing if that's preferred.
Should I change NormalizedMessage at the same time to keep things consistent?

piotr-oles: Yes, please :)

Pajn: Done

piotr-oles (approved):

johnnyreilly:
Closing due to inactivity - if this is still required please could you resubmit.

halfnibble:
We used to specify the version of the typescript compiler with the `typescript` option. What version did this option get removed in?

johnnyreilly:
5.0 I believe

PR #103, opened 2018-03-02
DOCS update: how to use overlay with webpack-dev-server

johnnyreilly:
If anyone is wondering how to use `overlay` with `webpack-dev-server` then this docs PR should set them straight; just set `async: false`.

Ironically I don't use this myself, but it's nice that I can!


johnnyreilly:
This is a docs PR so it definitely didn't break the build!  @piotr-oles  you might want to requeue travis...

piotr-oles:
Thank you @johnnyreilly - update merged :)

PR #109, opened 2018-03-29
Remove typescript and tslint from peerDependencies

### LINKED ISSUE #102: [suggestion] Remove tslint as a peer dependency
I like that the project supports tslint, but I don't think it should require it to be a peer dependency or even be a regular dependency. 

npm doesn't support having optionalPeerDependencies at this time, and I don't think it's necessary to nag the users of this library with unmet peer dependency warnings. My suggestion is to remove it from the peer dependency list.

This is just a suggestion. If you disagree, feel free to close the issue and I'll just deal with having the warning popup in my projects from time to time.

---
Related: [npm/npm#3066 - Optional Peer Dependencies](https://github.com/npm/npm/issues/3066)

marcofugaro:
Was also discussed in #102,

Our problem with those peerDeps, is that we optionally use typescript in our build process, so just by having this package in the dependencies, it throws the error `typescript is required`.

screen shot 2018-03-29 at 11 26 03

It should be like [ts-loader](https://github.com/TypeStrong/ts-loader/blob/master/package.json), typescript is required but it's up to the user to install it.

Thre README is fine since it says it explicitly to install also typescript.

Closes #102

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,8 +78,6 @@
     "webpack": "^4.0.0"
   },
   "peerDependencies": {
-    "tslint": "^4.0.0 || ^5.0.0",
-    "typescript": "^2.1.0",
     "webpack": "^2.3.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {


piotr-oles:
If we want to remove it from peerDeps, we should add some checks (for typescript package and it's version) in the plugin code to inform user that typescript is not available and he has to install it to use this plugin. :)

SimonSchick:
Isn't this what https://docs.npmjs.com/files/package.json#optionaldependencies is for?

nickserv:
Unfortunately optionalDependencies don't work with peerDependencies, they have to be direct dependencies of this package and not its dependent, which isn't helpful.

sebinsua:
This isn't correct.

`peerDependencies` will not cause an error about a module not existing. All it does is cause a warning to be printed during installation telling you what version range is expected. It's purely informational to help developers understand what version of the TypeScript compiler is supported. It's your choice whether you then install this.

If you want to not get runtime errors, you will need to write something like this.

```ts
let typescript;
try {
  typescript = require('typescript');
} catch (err) {
  console.warn('TypeScript needs to be installed for fork-ts-checker-webpack-plugin to do anything.');
}

// ...somewhere deep within the plugin...

if (typescript) {
  // Conditionally decide what to do if it is installed here.
}
```

glen-84:
> All it does is cause a warning to be printed during installation telling you what version range is expected.

... and this when using `npm list --depth=0`:

```
npm ERR! peer dep missing: tslint@^4.0.0 || ^5.0.0, required by fork-ts-checker-webpack-plugin@0.5.1
```

I'm not using the tslint integration, and it even requires setting `tslint: true`, so I'm not sure why it's listed as a peer dependency.

sebinsua:
If you remove the `peerDependency` you no longer have a way to inform the user of what version range of a dependency is supported.

I think what you're implicitly asking for is [optional `peerDependencies`. This feature is in discussion](https://github.com/yarnpkg/rfcs/pull/105).

marcofugaro:
Closing because it was done in #201

PR #113, opened 2018-04-21
Make vue-parser peerDep

### LINKED ISSUE #107: TypeScript types got broken by this project installation
We have pure-FrontEnd project, but after this project install, we have strange errors about Timer type, like this.

```
src/foo.ts(85,13): error TS2322: Type 'Timer' is not assignable to type 'number'.

85             this.timeout = setTimeout(loop, 0);
               ~~~~~~~~~~~~

```

this is because of node typings installed.

```
~/W/foo $ stat node_modules/@types/node/
16777220 8595635200 drwxr-xr-x 7 colch Users 0 224 "Mar 23 14:24:40 2018" "Mar 23 14:24:47 2018" "Mar 23 14:24:47 2018" "Mar 23 14:24:40 2018" 4194304 0 0 node_modules/@types/node/
```

these typings got installed because of this chain:

```
~/W/foo $ yarn why @types/node/
yarn why v1.5.1
warning package.json: No license field
[1/4] 🤔  Why do we have the module "@types/node/"...?
[2/4] 🚚  Initialising dependency graph...
warning yandex-pcode@0.0.1: No license field
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@types/node@9.4.6"
info Reasons this module exists
   - "fork-ts-checker-webpack-plugin#vue-parser#parse5" depends on it
   - Hoisted from "fork-ts-checker-webpack-plugin#vue-parser#parse5#@types#node"
info Disk size without dependencies: "20MB"
info Disk size with unique dependencies: "20MB"
info Disk size with transitive dependencies: "20MB"
info Number of shared dependencies: 0
✨  Done in 1.24s.
```

So, the question is, why `vue-parser` exists in `dependencies` anyway? We have no `vue` in project installed anyway, as well as react.

It seems that this introduced by https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/77

ColCh:
should close #107

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
   "peerDependencies": {
     "tslint": "^4.0.0 || ^5.0.0",
     "typescript": "^2.1.0",
+    "vue-parser": "^1.1.5",
     "webpack": "^2.3.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {
@@ -92,7 +93,6 @@
     "lodash.startswith": "^4.2.1",
     "minimatch": "^3.0.4",
     "resolve": "^1.5.0",
-    "tapable": "^1.0.0",
-    "vue-parser": "^1.1.5"
+    "tapable": "^1.0.0"
   }
 }


piotr-oles:
But not everyone use vue for development, so it will show warnings to users that don't use it. Instead we can add some checks in the plugin code to inform user that he has to install it to use vue mode. :)

ColCh:
@piotr-oles seems good!

PR #127, opened 2018-06-29
File system lag can cause Invalid source file errors to slip through

joshpcausey:
When using jetbrains idea on windows the save -> filesystem -> tslint process can be a bit lagging resulting in a partially written file being read by tslint when in watch mode. This causes a Invalid source file: ${FilePath}. Ensure that the files supplied to lint have a .ts, .tsx, .d.ts, .js or .jsx extension. fatal exception. When the file doesn't exist you are already ignoring errors (presumablly this error would be one of them). This PR expands that to also ignore the error when the filesystem isn't keeping up.

### DIFF
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -240,8 +240,7 @@ class IncrementalChecker {
       } catch (e) {
         if (
           fs.existsSync(fileName) &&
-          !e.message.trim().startsWith("Invalid source file") &&
-          fileName.match(/(.tsx?|.jsx?|.vue)$/)
+          !e.message.trim().startsWith("Invalid source file")
         ) {
           // it's not because file doesn't exist - throw error
           throw e;
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -238,7 +238,11 @@ class IncrementalChecker {
       try {
         this.linter.lint(fileName, undefined, this.linterConfig);
       } catch (e) {
-        if (fs.existsSync(fileName)) {
+        if (
+          fs.existsSync(fileName) &&
+          !e.message.trim().startsWith("Invalid source file") &&
+          /(.tsx?|.jsx?|.vue)$/.match(fileName)
+        ) {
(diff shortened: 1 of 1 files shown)


src/IncrementalChecker.ts
```diff
@@ -238,7 +238,10 @@ class IncrementalChecker {
       try {
         this.linter.lint(fileName, undefined, this.linterConfig);
       } catch (e) {
-        if (fs.existsSync(fileName)) {
+        if (
+          fs.existsSync(fileName) &&
+          !e.message.trim().startsWith("Invalid source file")
```
piotr-oles: Are we sure that linter always throws an "Error" like object with a message? I think it would be safer to add the additional check:
```
fs.existsSync(fileName) &&
!(e.message && e.message.trim().startsWith("Invalid source file"))
```
Maybe there is a better way to check if given error is about invalid source file - maybe we could add `instanceof` check instead of message matching? (just asking - not sure how it's implemented on tslint side).
The last thing - please add a comment here - it's not obvious that there could be a filesystem lag :)

piotr-oles (changes requested):

src/IncrementalChecker.ts
joshpcausey: I'll have to boot up my machine at work Monday to look again at it again. If I remember correctly it is just throwing a string error, but I'll give it another look. Thanks for quick response boss man. I'll get it all figured out!

joshpcausey:
It looks like there may be an issue with yarn and TravisCI

joshpcausey:
I interrogated the error and added a few more checks as well as a comment explaining the reasoning for the checks.

johnnyreilly:
@piotr-oles what do you think of the changes @joshpcausey has made?

piotr-oles (approved):

piotr-oles:
Thank you for your work! :)

joshpcausey:
Thank you for helping me through my first pull request on github (:

piotr-oles:
Released as v0.4.3 :)

PR #130, opened 2018-07-25
fix(vue): resolve src attribute on the script block on Vue files

### LINKED ISSUE #85: (vue) TSLint is running on empty script tags
TSLint is running on empty script tag content, even when a src attribute is defined.

```vue
<template src="./App.html"></template>
<style scoped src="./App.css"></style>
<script lang="ts" src="./App.ts"></script>
```

This raises tslint errors for missing EOL

```
ERROR in C:/devel/projects/vue-ts-test/src/App.vue
(3,91): file should end with a newline
```

### LINKED ISSUE #111: [vue] `src` attribute on script block is not resolved
Looks like `src` attribute on `<script>` block in Vue SFC is not handled by fork-ts-checker-webpack-plugin. For example, if we process the following `.vue` file:

```vue
<template>
  <h1>hi</h1>
</template>

<script lang="ts" src="./App.ts"></script>
```

It produces like the following error:
```
ERROR in **/src/main.ts
(2,17): File '**/src/App.vue' is not a module.
```

The complete reproduction is [here](https://github.com/ktsn/vue-fork-ts-checker-webpack-plugin-repro).

FYI, if I only use `ts-loader` with type checking, it successfully compiled without errors.

original issue: https://github.com/vuejs/vue-cli/issues/1104

ktsn:
This PR fixes #111 and https://github.com/vuejs/vue-cli/issues/1104. Also probably fixes #85.

I've replaced `vue-parser` with [`vue-template-compiler`](https://www.npmjs.com/package/vue-template-compiler) which is Vue official package for parsing `.vue` file. Because `vue-parser` looks not maintained and it would be better to use officially provided lib for consistency.

I didn't include `vue-template-compiler` in dependencies because there is a known issue that mismatching with user-installed `vue` version.
The users should not feel inconvenience by this because [`vue-loader`](https://github.com/vuejs/vue-loader) (official webpack loader for `.vue` file) enforces the users to install `vue-template-compiler` due to this issue. That means if the users want to use `vue` option of this plugin, they already install `vue-template-compiler` by themselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,10 +71,10 @@
     "ts-loader": "4.3.0",
     "tslint": "^5.0.0",
     "typescript": "^2.6.2",
-    "vue": "^2.5.9",
+    "vue": "^2.5.16",
     "vue-class-component": "^6.1.1",
     "vue-loader": "^15.2.4",
-    "vue-template-compiler": "^2.5.9",
+    "vue-template-compiler": "^2.5.16",
     "webpack": "^4.0.0"
   },
   "peerDependencies": {
@@ -92,7 +92,6 @@
     "lodash.startswith": "^4.2.1",
     "minimatch": "^3.0.4",
     "resolve": "^1.5.0",
-    "tapable": "^1.0.0",
-    "vue-parser": "^1.1.5"
+    "tapable": "^1.0.0"
   }
 }
--- a/src/VueProgram.ts
+++ b/src/VueProgram.ts
@@ -228,7 +228,9 @@ class VueProgram {
       const src = script.attrs.src.replace(/\.tsx?$/i, '');
       return {
         scriptKind,
-        content: `// tslint:disable\nexport { default } from '${src}';`
+        content: '// tslint:disable\n'
+          + `export { default } from '${src}';\n`
+          + `export * from '${src}';\n`
       };
     }
 
--- a/src/VueProgram.ts
+++ b/src/VueProgram.ts
@@ -216,7 +216,7 @@ class VueProgram {
     if (!script) {
       return {
         scriptKind: ts.ScriptKind.JS,
-        content: '// tslint:disable\nexport default {};\n'
+        content: '/* tslint:disable */\nexport default {};\n'
       };
     }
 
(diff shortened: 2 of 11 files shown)


johnnyreilly:
Thanks for the PR @ktsn! Looks initially good to me.

@prograhammer @CKGrafico @wonderful-panda I believe you're Vue users and have worked on the Vue support in the plugin. Would you be able to look at this PR and let us know what you think?

cc @yyx990803

CKGrafico:
Looks really nice!

wonderful-panda:
LGTM

yyx990803:
👍

johnnyreilly:
Okay - I think we're good to merge! @ktsn can you comment on whether the `vue-template-compiler` types are directly consumable or not?

ktsn:
Thanks for taking a look at this PR 🙏 

@johnnyreilly The current vue-template-compiler does not include the types but it would be included and we will be able to consume it directly in the future since [there is a PR already](https://github.com/vuejs/vue/pull/7918) (maybe vue-template-compiler v2.6.0?).

johnnyreilly:
Awesome! Then let's go with this for now and if that PR gets merged we can switch to consuming from there later. Thanks for your work!

johnnyreilly:
Merged - this will be published once a new token has been generated for publishing.  See https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/132

nobelhuang:
I'm not pretty sure, but seems this is the change that will somewhat cause tslint to complain if `no-consecutive-blank-lines` rule is used when we try to compile those vue files written in TS/TSX. I guess it is just because `vue-template-compiler` didn't remove those blank lines after extracting ts part from SFC, then we got this error:
> no-consecutive-blank-lines: Consecutive blank lines are forbidden

ktsn:
We should not trim blank line because reported diagnostics will point to wrong location if we do that.

I remember vue-template-compiler pads the blank lines with `//` to avoid such lint errors (almost the same approach with vue-parser) but it looks applicable only when `<script>` block lang is JavaScript.

I will make a PR to fix the issue later.

PR #134, opened 2018-08-01
Use minimum peer dependencies

### LINKED ISSUE #133: TypeScript 3 peerDependencies warning
If this project officially supports TypeScript 3, its peer dependencies should be updated to allow it.

nickserv:
Fixes #133.

I have used TypeScript 3 with this loader in a CRA-TS app and ran the build and test steps test no issues so far. Let me know if there's anything else we need to change or test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,9 +78,9 @@
     "webpack": "^4.0.0"
   },
   "peerDependencies": {
-    "tslint": "^4.0.0 || ^5.0.0",
-    "typescript": "^2.1.0",
-    "webpack": "^2.3.0 || ^3.0.0 || ^4.0.0"
+    "tslint": ">= 4.0.0",
+    "typescript": ">= 2.1.0",
+    "webpack": ">= 2.3.0"
   },
   "dependencies": {
     "babel-code-frame": "^6.22.0",
--- a/package.json
+++ b/package.json
@@ -78,9 +78,9 @@
     "webpack": "^4.0.0"
   },
   "peerDependencies": {
-    "tslint": ">= 4.0.0",
-    "typescript": ">= 2.1.0",
-    "webpack": ">= 2.3.0"
+    "tslint": "^4.0.0 || ^5.0.0",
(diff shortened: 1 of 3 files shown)


nickserv:
@johnnyreilly

johnnyreilly:
Looks good to me. What do you think @piotr-oles?

package.json
```diff
@@ -70,17 +70,17 @@
...
     "vue-class-component": "^6.1.1",
     "vue-loader": "^15.2.4",
     "vue-template-compiler": "^2.5.16",
     "webpack": "^4.0.0"
   },
   "peerDependencies": {
-    "tslint": "^4.0.0 || ^5.0.0",
-    "typescript": "^2.1.0",
-    "webpack": "^2.3.0 || ^3.0.0 || ^4.0.0"
+    "tslint": ">= 4.0.0",
```
piotr-oles: It's a bold assumption that next major releases of tslint, typescript and webpack will be compatible with this plugin :) Please add typescript ^3.0.0 instead

piotr-oles (changes requested):
Please add also cases in .travis.yml for TypeScript 2.x and TypeScript 3.x (using ENV variable - like it's done for ts-loader, ts-lint and vue-loader)

package.json
nickserv: I agree. Done.

nickserv:
I'm not sure if using ENV variables like that for TypeScript would be safe in the future. I updated TypeScript in the package's own devDependencies so you can take advantage of new syntax features, along with the improved project support and error messages. If you change source code to have syntax or config options that are only supported in TS 3, the TS 2 env will fail.

Are you sure you still want to try that? That would mean this package cannot use TS 3 features in its own TS files without changing the build setup further to separate TS transpilation from integration testing other versions of TS in other packages.

piotr-oles:
Good point - let's keep test as they are currently :) Thank you for your work!

piotr-oles (approved):

PR #141, opened 2018-08-08
Make options Partial<Options>

Larry1123:
When working with webpack.config.ts ts complains when you try to pass a partial object literal.
This fixes that.


johnnyreilly:
This looks great! Would you be able to increment the version to `0.4.6` in the `package.json` and put a corresponding entry in the changelog?

That will make it super easy for me to cut a release.

johnnyreilly:
Thanks - should now be released!

PR #146, opened 2018-08-10
Set "compilationDone" before "forkTsCheckerServiceBeforeStart"

walkerburgin:
I'm trying to use the `forkTsCheckerServiceBeforeStart` hook to defer type checking until after compilation has finished in order to pick up typings generated on the fly by a loader. In this case, compilation will be finished before `forkTsCheckerServiceBeforeStart` is resolved. Right now `this.compilationDone` will be incorrectly set to `false`, causing the build to hang indefinitely.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -313,6 +313,7 @@ class ForkTsCheckerWebpackPlugin {
     if ('hooks' in this.compiler) {
       // webpack 4
       this.compiler.hooks.compile.tap(checkerPluginName, () => {
+        this.compilationDone = false;
         this.compiler.hooks.forkTsCheckerServiceBeforeStart.callAsync(() => {
           if (this.cancellationToken) {
             // request cancellation if there is not finished job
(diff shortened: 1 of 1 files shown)


johnnyreilly:
@piotr-oles what do you think about this?

walkerburgin:
Anything I can do to help? I'm blocked at the moment on finding some way of accomplishing this (definitely open to better solutions)

johnnyreilly:
It's fine with me; but since it represents a change to behaviour I'd like @piotr-oles to review before we merge. He may be on holiday so please bear with us!

johnnyreilly:
Ps could you merge in upstream changes to your branch?

piotr-oles:
Looks good :) Thanks

walkerburgin:
Any chance we can get a release with this change?

johnnyreilly:
Do you want to submit a PR that updates the changelog and the package.json? That will make it easy for me to arrange...

See https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/DEPLOYING.md for context

walkerburgin:
Sure, opened https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/152

johnnyreilly:
Thanks - release should be out by the time you read this. 

Thanks again for your work!

PR #157, opened 2018-09-20
Fix for android when doing yarn start

freefaint:
Added hack to prevent crashing apps using this one when os.cpus returns undefined

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,7 +66,7 @@ interface Options {
 class ForkTsCheckerWebpackPlugin {
   static DEFAULT_MEMORY_LIMIT = 2048;
   static ONE_CPU = 1;
-  static ALL_CPUS = (os.cpus() || [null]).length;
+  static ALL_CPUS = os.cpus() ? os.cpus().length : 1;
   static ONE_CPU_FREE = Math.max(1, ForkTsCheckerWebpackPlugin.ALL_CPUS - 1);
   static TWO_CPUS_FREE = Math.max(1, ForkTsCheckerWebpackPlugin.ALL_CPUS - 2);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,7 +66,7 @@ interface Options {
 class ForkTsCheckerWebpackPlugin {
   static DEFAULT_MEMORY_LIMIT = 2048;
   static ONE_CPU = 1;
-  static ALL_CPUS = os.cpus().length;
+  static ALL_CPUS = (os.cpus() || [null]).length;
   static ONE_CPU_FREE = Math.max(1, ForkTsCheckerWebpackPlugin.ALL_CPUS - 1);
   static TWO_CPUS_FREE = Math.max(1, ForkTsCheckerWebpackPlugin.ALL_CPUS - 2);
 


johnnyreilly (commented):
What does `os.cpus()` return on Android? `0`? A falsey value?

freefaint:
It returns undefined and you're waiting for array trying to get its length

freefaint:
Android 8.1

johnnyreilly:
I wonder if `os.cpus() || 1` might achieve the same goal but be a little clearer?

freefaint:
You are waiting that os.cpus() returns an array, if you call os.cpus().length. So in case of undefined we need an array that has one element. It's safe, because in that case we don't really know real number of cores in device cpu.

I checked my code, it works) Thank you)

freefaint:
The problem Is that you're trying to get the length of undefined. It crashes if you get undefined from cpus(). so we need to prevent undefined before .length. the safest way is to polyfill array that has 1 element. This fix will not change your logic , just will prevent crash

johnnyreilly:
Sorry - I wrote the wrong thing; first time.  What I meant to write was: 

```
os.cpus() ? os.cpus().length : 1;
```

This should have the same effect as yours but be more obvious in its intent.  What I mean is, when I read `(os.cpus() || [null]).length` my immediate thought is "what's the significance of `null`"?  Why `null`?

You have to think some more before you draw the conclusion that `null` is not significant at all; it could be any value at all; all you care about is an array having a single member so you can `.length`.  

I'm very much a fan of code being very clear and easy to reason about; hence my suggestion.  What do you think?

freefaint:
Ok, null is shit really)

But think about your example.

os.cpus().length || 1

Do you understand, thit this will cause an error. You're trying to get length property from undefined still in this code. This moment crashes the app

freefaint:
Maybe this solving

(function() { try { return os.cpus().length} catch (e) { return 1 } })()

johnnyreilly:
> Do you understand, thit this will cause an error. You're trying to get length property from undefined still in this code. This moment crashes the app

You're right - that's why I suggested this: (bug in my initial idea)

`os.cpus() ? os.cpus().length : 1;`

freefaint:
In this way you decrease performance because call cpus() 2 times so thats why i solved it in my way...

But your code solves the problem) thank you)

johnnyreilly:
> In this way you decrease performance

A fair point. However, this is something that's run once only I think (upon instantiation); I'm pretty sure the difference to consumers will be so unnoticeable as to not matter. I'd lay good money that V8 optimises this away to nothing anyway :smile:

johnnyreilly:
Would you like to tweak the solution to my suggestion?  I don't see any reason why we couldn't merge that.

freefaint:
Im sorry, i'm bad at english, i'm afraid i cant fully understand your last comment :) If I understand you right, you wanted me to change the solution as we descided :) It's ready)

johnnyreilly:
That's great! Thanks. 

@piotr-oles you happy for us to merge?

piotr-oles:
Looks good - thank you for your work :)

freefaint:
thank you :)

now you we can launch react typescript applications with yarn in android environment (i used termux console emulator)

unbeliveable!!!

PR #163, opened 2018-10-04
Export main plugin class and update all imports ? = require(?) to named exports

### LINKED ISSUE #162: Unable to compile typescript after importing the latest version to my config files...
Hi, i have been using old version of this plugin for quite a long time, recently i tried to upgrade to the latest and it doesn't compile my typescript files anymore. probably due to the .d.ts that is being shipped along with it now... Here is how im using it:

webpack.config.ts

```ts
import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'

...
plugins: [
     new ForkTsCheckerWebpackPlugin(),
]
```

error

```
tsc --lib esnext webpack/webpack.*.ts

node_modules/fork-ts-checker-webpack-plugin/lib/types/NormalizedMessage.d.ts:2:30 - error TS2307: Cannot find module 'tslint'.

2 import tslintTypes = require('tslint');
                               ~~~~~~~~

webpack/webpack.common.ts:8:45 - error TS2497: Module '"G:/Code/demo/node_modules/fork-ts-checker-webpack-plugin/lib/types/index"' r
esolves to a non-module entity and cannot be imported using this construct.

8 import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! webpack4@1.0.0 prestart: `tsc --lib esnext webpack/webpack.*.ts`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the webpack4@1.0.0 prestart script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Rafael\AppData\Roaming\npm-cache\_logs\2018-10-03T11_37_04_173Z-debug.log
```

so once i forcibily install tslint (which im not using) and change 
```
import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
```
to
```
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

then so im able to "compile" my .ts config files...  

anyways is it possible to change the auto generated .d.ts files to allow it to be imported as:

```ts
 import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
 import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
 import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin'

anything but: 😞 

const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

and remove the forced dependency on tslint? or at least dont push the .d.ts to npm, perhaps publish to `@types/fork-ts-checker-webpack-plugin`?

milewski:
this PR allows the types exposed by this plugin to be effectively used inside a .ts file.

the imports can be as:

```ts
import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin'
import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
```

currently only this variants are allowed:

```ts
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

this PR also updates all commonjs style imports to named imports/export and closes #162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.4.10
+
+* [Allow fork-ts-checker-webpack-plugin to be imported in .ts files using ESM import syntax](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/163) (#163)
+
 ## v0.4.9
 
 * [Set "compilationDone" before resolving "forkTsCheckerServiceBeforeStart"](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/146) (#146)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fork-ts-checker-webpack-plugin",
-  "version": "0.4.9",
+  "version": "0.4.10",
   "description": "Runs typescript type checker and linter on separate process.",
   "main": "lib/index.js",
   "types": "lib/types/index.d.ts",
--- a/src/CancellationToken.ts
+++ b/src/CancellationToken.ts
@@ -1,15 +1,15 @@
-import crypto = require('crypto');
-import fs = require('fs');
-import os = require('os');
-import path = require('path');
-import ts = require('typescript');
+import * as crypto from 'crypto';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import * as ts from 'typescript';
 
 interface CancellationTokenData {
   isCancelled: boolean;
   cancellationFileName: string;
 }
 
-class CancellationToken {
+export class CancellationToken {
   isCancelled: boolean;
   cancellationFileName: string;
   lastCancellationCheckTime: number;
@@ -72,5 +72,3 @@ class CancellationToken {
     }
   }
 }
-
-export = CancellationToken;
--- a/src/FilesRegister.ts
+++ b/src/FilesRegister.ts
@@ -1,4 +1,4 @@
-import ts = require('typescript');
+import * as ts from 'typescript';
 
 interface DataShape {
   source: ts.SourceFile;
@@ -6,7 +6,7 @@ interface DataShape {
   lints: any[];
 }
 
-class FilesRegister {
+export class FilesRegister {
   files: { [filePath: string]: { mtime: number; data: DataShape; }};
   dataFactory: (_data?: any) => DataShape; // It doesn't seem that the _data parameter is ever used?
 
@@ -74,4 +74,4 @@ class FilesRegister {
     }
   }
 }
-export = FilesRegister;
+
--- a/src/FilesWatcher.ts
+++ b/src/FilesWatcher.ts
@@ -1,8 +1,8 @@
-import chokidar = require('chokidar');
-import path = require('path');
+import * as chokidar from 'chokidar';
+import * as path from 'path';
 import startsWith = require('lodash.startswith');
 
-class FilesWatcher {
+export class FilesWatcher {
   watchPaths: string[];
   watchExtensions: string[];
   watchers: chokidar.FSWatcher[];
@@ -71,5 +71,3 @@ class FilesWatcher {
     }
   }
 }
-
-export = FilesWatcher;
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -1,26 +1,26 @@
-import fs = require('fs');
+import * as fs from 'fs';
 import endsWith = require('lodash.endswith');
-import path = require('path');
-import ts = require('typescript');
-import tslintTypes = require('tslint'); // Imported for types alone; actual requires take place in methods below
-import FilesRegister = require('./FilesRegister');
-import FilesWatcher = require('./FilesWatcher');
-import WorkSet = require('./WorkSet');
-import NormalizedMessage = require('./NormalizedMessage');
-import CancellationToken = require('./CancellationToken');
-import minimatch = require('minimatch');
-import VueProgram = require('./VueProgram');
+import * as path from 'path';
+import * as ts from 'typescript';
+import { Configuration, Linter } from 'tslint'; // Imported for types alone; actual requires take place in methods below
+import { FilesRegister } from './FilesRegister';
+import { FilesWatcher } from './FilesWatcher';
+import { WorkSet } from './WorkSet';
+import { NormalizedMessage } from './NormalizedMessage';
+import { CancellationToken } from './CancellationToken';
+import * as minimatch from 'minimatch';
+import { VueProgram } from './VueProgram';
 
 // Need some augmentation here - linterOptions.exclude is not (yet) part of the official
 // types for tslint.
-interface ConfigurationFile extends tslintTypes.Configuration.IConfigurationFile {
+interface ConfigurationFile extends Configuration.IConfigurationFile {
   linterOptions?: {
     typeCheck?: boolean;
     exclude?: string[];
   };
 }
 
-class IncrementalChecker {
+export class IncrementalChecker {
   programConfigFile: string;
   linterConfigFile: string | false;
   watchPaths: string[];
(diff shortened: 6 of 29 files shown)


johnnyreilly:
Thanks for the work! Looks good. With this change do the existing imports still work? i.e.

```
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

If not then this would need to be a breaking changes release. 

Could you also update the CHANGELOG / version number in the `package.json` please?

milewski:
it is a breaking change... the output of that object would be:

```ts
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')

console.log(ForkTsCheckerWebpackPlugin)

{
  default: [Function ForkTsCheckerWebpackPlugin]
  ForkTsCheckerWebpackPlugin: [Function ForkTsCheckerWebpackPlugin]
}
```

so the import should be like this in this case:

```ts
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin').ForkTsCheckerWebpackPlugin
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin').ForkTsCheckerWebpackPlugin

or

const { ForkTsCheckerWebpackPlugin } = require('fork-ts-checker-webpack-plugin')
import { ForkTsCheckerWebpackPlugin } = require('fork-ts-checker-webpack-plugin')
```

similiar how it was done here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/163/commits/d48440f7e9af0e04f39bacd12d47261f5ba11f6d#diff-f41e9d04a45c83f3b6f6e630f10117feL16

milewski:
but hold on, i think there is a way to keep the exported object like this:

```
{
  [Function ForkTsCheckerWebpackPlugin]
  default: [Function ForkTsCheckerWebpackPlugin]
  ForkTsCheckerWebpackPlugin: [Function ForkTsCheckerWebpackPlugin]
}
```

so it wouldn't be a breaking change let me do some experiments here first..

johnnyreilly:
Sure

milewski:
OK i got it working, but this has to assume that user tsconfig.json module is set to `commonjs` if it set to `esnext` for example this syntax wont work 

```ts
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

but this will

```ts
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin').ForkTsCheckerWebpackPlugin
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin').ForkTsCheckerWebpackPlugin

or

const { ForkTsCheckerWebpackPlugin } = require('fork-ts-checker-webpack-plugin')
import { ForkTsCheckerWebpackPlugin } = require('fork-ts-checker-webpack-plugin')

or

import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin'
import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
```

johnnyreilly:
So this commit is very noisy; could you undo the whitespace changes please? I can't make out what's actually changed here... https://github.com/Realytics/fork-ts-checker-webpack-plugin/commit/2106654b03b52da9dd63a23df996712bc0461190

Probably we should switch to using prettier for consistent formatting in future.

It's unclear to me whether consumption is still unchanged; I'd like users to be able to use this consumption ideally:

```
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```

It feels that ought to be possible. ts-loader manages it. Maybe this helps? 


https://github.com/TypeStrong/ts-loader/blob/d28ba6f732346067535d25889b8d56c95961c4a6/src/index.ts#L505

https://github.com/TypeStrong/ts-loader/blob/master/src/tsconfig.json

milewski:
@johnnyreilly sorry my IDE had that format before commit option on..... forgot to unckeck.. it is clean now...

and if i only export it like this:

```ts
export = ForkTsCheckerWebpackPlugin
```
it would be exactly the same as it was originally.... so i would still face the same problem importing it..

and what i meant about this being a breaking change or not... let says that the user has `module: commonjs` on his tsconfig.json, this **wouldn't** be a breaking change for them, however those with the option set to `module: esnext or es2015` would break, because it would no longer be a commonjs import but ES one.. which has to be imported as any of these forms:

```ts
import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin'
import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
```

and of course importing by `require('fork-ts-checker-webpack-plugin')` would still work the only difference is that doesnt return a function but an object like this:

```ts
{
  default: [Function ForkTsCheckerWebpackPlugin]
  ForkTsCheckerWebpackPlugin: [Function ForkTsCheckerWebpackPlugin]
}
```

so i think this change is a breaking change, but it would only affect a very low percentage of this plugin users... because their webpack.config would have to be written in .ts for them to feel any effect, which i believe everyone defaults to a .js file..

johnnyreilly:
Hey @milewski 

Thanks for that. I want to clarify the situation for users writing vanilla js `webpack.config.js` files (most users)

They can still write `
const  ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')` and they *don't* need to then `.ForkTsCheckerWebpackPlugin` to use the plugin? i.e. it's a default export still?

milewski:
@johnnyreilly sorry for the confusion... i followed the example u sent (https://github.com/TypeStrong/ts-loader/blob/d28ba6f732346067535d25889b8d56c95961c4a6/src/index.ts#L505) and this seems to be a much better solution, by doing so i dont need to change the export method and **it will work for .js and .ts** with the exacly same signature 

the last commit reverts the export back how it was originally and 

for .js

```js
console.log(require('fork-ts-checker-webpack-plugin'))
// { [Function: ForkTsCheckerWebpackPlugin] }
```

for .ts

```ts
console.log(require('fork-ts-checker-webpack-plugin'))
// { [Function: ForkTsCheckerWebpackPlugin] }

import * as ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
console.log(ForkTsCheckerWebpackPlugin)
// { [Function: ForkTsCheckerWebpackPlugin] }

// Invalid imports for .ts, which is ok as there is at least 1 that works for .ts files
import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin'

```

so this PR **is not a breaking change** for neither .ts users nor .js users

johnnyreilly:
Great work!  Do you want to update the `package.json` to a new version please? And also update the `CHANGELOG.md` too.  That makes it easy for us to ship a release

milewski:
done 👍

johnnyreilly:
Thanks! Shipping with https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v0.4.10

PR #166, opened 2018-10-14
[WIP] webpack@5 `hooks` support

liangchunn:
In `webpack@5`, custom hooks are no longer allowed to be added directly onto `this.compiler.hooks`. `WeakMap + static getHooks()` should be used instead (More details here: https://github.com/webpack/webpack/projects/5#card-10390019)

This is a WIP, please give feedback on what should be updated.

### Changed
- [x] (webpack5): Use `static getHooks()` 
- [x] (webpack5): Use `require('webpack').version` when available in `test/integration/webpackVersion.js`
- [x] (ci): Add `webpack@5.0.0-alpha.0` into CI env
- [x] (docs): Add migration section for webpack4+ hooks
- [x] (package): remove support for node6
- [x] (webpack5): rename hooks

### Resolved
- [x] (package): Decide if webpack5 support should be added into existing version/have breaking major version
- [x] ~(webpack5): Decide if `_pluginCompat` should be removed (*)~ Removed `_pluginCompat`
- [x] ~(ci): Decide if we should reorder `yarn add $WEBPACK $TSLOADER $VUELOADER -D` in `travis.yml`~
- [x] ~(ci): Skip tests for `node6 + webpack@5`~

### Resolved in #169 instead
~- [x] (ci): Lint in CI~
~- [x] (lint): Upgrade TSLint and ESLint~
~- [x] (lint): Lint `*.ts` and `*.js` files~
~- [x] (tests): don't use object spread (node6 does not support this: https://node.green/#ES2018-features-object-rest-spread-properties)~

---

~(*) I don't really know what `_pluginCompat` does, removing it doesn't cause build failures on CI~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,4 @@ deploy:
   on:
     tags: true
     branch: master
+  tag: next
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
  - yarn add $WEBPACK $TSLOADER $VUELOADER -D
  - yarn lint
 env:
+ - WEBPACK=webpack@^5.0.0-alpha.0 TSLOADER=ts-loader@^5.0.0 VUELOADER=vue-loader@^15.2.4
  - WEBPACK=webpack@^4.0.0 TSLOADER=ts-loader@^5.0.0 VUELOADER=vue-loader@^15.2.4
  - WEBPACK=webpack@^3.10.0 TSLOADER=ts-loader@^3.4.0 VUELOADER=vue-loader@^13.5.0
  - WEBPACK=webpack@^2.7.0 TSLOADER=ts-loader@^3.4.0 VUELOADER=vue-loader@^13.5.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: node_js
 node_js:
  - '10'
  - '8'
- - '6'
 install:
  - yarn install
  - yarn build
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,0 +1,51 @@
+import * as webpack from 'webpack';
+import { AsyncSeriesHook, SyncHook } from 'tapable';
+
+const compilerHookMap = new WeakMap();
+
+export const customHooks = {
+  forkTsCheckerServiceBeforeStart: 'fork-ts-checker-service-before-start',
+  forkTsCheckerCancel: 'fork-ts-checker-cancel',
+  forkTsCheckerServiceStartError: 'fork-ts-checker-service-start-error',
+  forkTsCheckerWaiting: 'fork-ts-checker-waiting',
+  forkTsCheckerServiceStart: 'fork-ts-checker-service-start',
+  forkTsCheckerReceive: 'fork-ts-checker-receive',
+  forkTsCheckerServiceOutOfMemory: 'fork-ts-checker-service-out-of-memory',
+  forkTsCheckerEmit: 'fork-ts-checker-emit',
+  forkTsCheckerDone: 'fork-ts-checker-done'
+};
+
+function createForkTsCheckerWebpackPluginHooks(): Record<
+  keyof typeof customHooks,
+  SyncHook | AsyncSeriesHook
+> {
+  return {
+    forkTsCheckerServiceBeforeStart: new AsyncSeriesHook([]),
+    forkTsCheckerCancel: new SyncHook(['cancellationToken']),
+    forkTsCheckerServiceStartError: new SyncHook(['error']),
+    forkTsCheckerWaiting: new SyncHook(['hasTsLint']),
+    forkTsCheckerServiceStart: new SyncHook([
+      'tsconfigPath',
+      'tslintPath',
+      'watchPaths',
+      'workersNumber',
+      'memoryLimit'
+    ]),
+    forkTsCheckerReceive: new SyncHook(['diagnostics', 'lints']),
+    forkTsCheckerServiceOutOfMemory: new SyncHook([]),
+    forkTsCheckerEmit: new SyncHook(['diagnostics', 'lints', 'elapsed']),
+    forkTsCheckerDone: new SyncHook(['diagnostics', 'lints', 'elapsed'])
+  };
+}
+
+export function getForkTsCheckerWebpackPluginHooks(compiler: webpack.Compiler) {
+  let hooks: Record<
+    keyof typeof customHooks,
+    SyncHook | AsyncSeriesHook
+  > = compilerHookMap.get(compiler);
+  if (hooks === undefined) {
+    hooks = createForkTsCheckerWebpackPluginHooks();
+    compilerHookMap.set(compiler, hooks);
+  }
+  return hooks;
+}
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,48 +1,58 @@
 import * as webpack from 'webpack';
 import { AsyncSeriesHook, SyncHook } from 'tapable';
 
-const compilerHookMap = new WeakMap();
+type ForkTsCheckerHooks =
+  | 'serviceBeforeStart'
+  | 'cancel'
+  | 'serviceStartError'
+  | 'waiting'
+  | 'serviceStart'
+  | 'receive'
+  | 'serviceOutOfMemory'
+  | 'emit'
+  | 'done';
+type ForkTsCheckerHookMap = Record<
+  ForkTsCheckerHooks,
+  SyncHook | AsyncSeriesHook
+>;
+type ForkTsCheckerLegacyHookMap = Record<ForkTsCheckerHooks, string>;
 
-export const customHooks = {
-  forkTsCheckerServiceBeforeStart: 'fork-ts-checker-service-before-start',
-  forkTsCheckerCancel: 'fork-ts-checker-cancel',
-  forkTsCheckerServiceStartError: 'fork-ts-checker-service-start-error',
-  forkTsCheckerWaiting: 'fork-ts-checker-waiting',
-  forkTsCheckerServiceStart: 'fork-ts-checker-service-start',
-  forkTsCheckerReceive: 'fork-ts-checker-receive',
-  forkTsCheckerServiceOutOfMemory: 'fork-ts-checker-service-out-of-memory',
-  forkTsCheckerEmit: 'fork-ts-checker-emit',
-  forkTsCheckerDone: 'fork-ts-checker-done'
+const compilerHookMap = new WeakMap<webpack.Compiler, ForkTsCheckerHookMap>();
+
+export const legacyHookMap: ForkTsCheckerLegacyHookMap = {
+  serviceBeforeStart: 'fork-ts-checker-service-before-start',
+  cancel: 'fork-ts-checker-cancel',
+  serviceStartError: 'fork-ts-checker-service-start-error',
+  waiting: 'fork-ts-checker-waiting',
+  serviceStart: 'fork-ts-checker-service-start',
+  receive: 'fork-ts-checker-receive',
+  serviceOutOfMemory: 'fork-ts-checker-service-out-of-memory',
+  emit: 'fork-ts-checker-emit',
+  done: 'fork-ts-checker-done'
 };
 
-function createForkTsCheckerWebpackPluginHooks(): Record<
-  keyof typeof customHooks,
-  SyncHook | AsyncSeriesHook
-> {
+function createForkTsCheckerWebpackPluginHooks(): ForkTsCheckerHookMap {
   return {
-    forkTsCheckerServiceBeforeStart: new AsyncSeriesHook([]),
-    forkTsCheckerCancel: new SyncHook(['cancellationToken']),
-    forkTsCheckerServiceStartError: new SyncHook(['error']),
-    forkTsCheckerWaiting: new SyncHook(['hasTsLint']),
-    forkTsCheckerServiceStart: new SyncHook([
+    serviceBeforeStart: new AsyncSeriesHook([]),
+    cancel: new SyncHook(['cancellationToken']),
+    serviceStartError: new SyncHook(['error']),
+    waiting: new SyncHook(['hasTsLint']),
+    serviceStart: new SyncHook([
       'tsconfigPath',
       'tslintPath',
       'watchPaths',
       'workersNumber',
       'memoryLimit'
     ]),
-    forkTsCheckerReceive: new SyncHook(['diagnostics', 'lints']),
-    forkTsCheckerServiceOutOfMemory: new SyncHook([]),
-    forkTsCheckerEmit: new SyncHook(['diagnostics', 'lints', 'elapsed']),
-    forkTsCheckerDone: new SyncHook(['diagnostics', 'lints', 'elapsed'])
+    receive: new SyncHook(['diagnostics', 'lints']),
+    serviceOutOfMemory: new SyncHook([]),
+    emit: new SyncHook(['diagnostics', 'lints', 'elapsed']),
+    done: new SyncHook(['diagnostics', 'lints', 'elapsed'])
   };
 }
 
 export function getForkTsCheckerWebpackPluginHooks(compiler: webpack.Compiler) {
-  let hooks: Record<
-    keyof typeof customHooks,
-    SyncHook | AsyncSeriesHook
-  > = compilerHookMap.get(compiler);
+  let hooks = compilerHookMap.get(compiler);
   if (hooks === undefined) {
     hooks = createForkTsCheckerWebpackPluginHooks();
     compilerHookMap.set(compiler, hooks);
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,21 +12,9 @@ import { createCodeframeFormatter } from './formatter/codeframeFormatter';
 import { FsHelper } from './FsHelper';
 import { Message } from './Message';
 
-import { AsyncSeriesHook, SyncHook } from 'tapable';
+import { getForkTsCheckerWebpackPluginHooks } from './hooks';
 
 const checkerPluginName = 'fork-ts-checker-webpack-plugin';
-
-const customHooks = {
-  forkTsCheckerServiceBeforeStart: 'fork-ts-checker-service-before-start',
-  forkTsCheckerCancel: 'fork-ts-checker-cancel',
-  forkTsCheckerServiceStartError: 'fork-ts-checker-service-start-error',
-  forkTsCheckerWaiting: 'fork-ts-checker-waiting',
-  forkTsCheckerServiceStart: 'fork-ts-checker-service-start',
-  forkTsCheckerReceive: 'fork-ts-checker-receive',
-  forkTsCheckerServiceOutOfMemory: 'fork-ts-checker-service-out-of-memory',
-  forkTsCheckerEmit: 'fork-ts-checker-emit',
-  forkTsCheckerDone: 'fork-ts-checker-done'
-};
 
 type Formatter = (message: NormalizedMessage, useColors: boolean) => string;
 
@@ -183,6 +171,10 @@ class ForkTsCheckerWebpackPlugin {
     this.vue = options.vue === true; // default false
   }
 
+  private static getCompilerHooks(compiler: webpack.Compiler) {
+    return getForkTsCheckerWebpackPluginHooks(compiler);
+  }
+
   private static createFormatter(type: 'default' | 'codeframe', options: any) {
     switch (type) {
       case 'default':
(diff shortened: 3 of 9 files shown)


src/hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+            'memoryLimit',
+        ]),
+        forkTsCheckerReceive: new SyncHook(['diagnostics', 'lints']),
+        forkTsCheckerServiceOutOfMemory: new SyncHook([]),
+        forkTsCheckerEmit: new SyncHook(['diagnostics', 'lints', 'elapsed']),
+        forkTsCheckerDone: new SyncHook(['diagnostics', 'lints', 'elapsed']),
+    };
+}
+
+export function getForkTsCheckerWebpackPluginHooks(compiler: any) {
```
johnnyreilly: Ideally I'd like to avoid the use of `any`; could you take a look at this please?

src/index.ts
```diff
@@ -156,6 +142,10 @@ class ForkTsCheckerWebpackPlugin {
     this.vue = options.vue === true; // default false
   }
 
+  static getHooks(compiler: any) {
```
johnnyreilly: Again, can we avoid `any` please!

src/index.ts
```diff
@@ -272,59 +259,38 @@ class ForkTsCheckerWebpackPlugin {
...
-        case customHooks.forkTsCheckerReceive:
-        case customHooks.forkTsCheckerServiceOutOfMemory:
-        case customHooks.forkTsCheckerEmit:
-        case customHooks.forkTsCheckerDone:
-          return true;
-      }
-      return undefined;
-    });
-  }
+  // registerCustomHooks() {
```
johnnyreilly: Can we avoid commented out code please?

src/index.ts
```diff
@@ -272,59 +259,38 @@ class ForkTsCheckerWebpackPlugin {
...
+  //     return undefined;
+  //   });
+  // }
 
   pluginCompile() {
     if ('hooks' in this.compiler) {
+      const hooks = ForkTsCheckerWebpackPlugin.getHooks(this.compiler);
       // webpack 4
-      this.compiler.hooks.compile.tap(checkerPluginName, () => {
+      this.compiler.hooks.compilation.tap(checkerPluginName, () => {
```
johnnyreilly: Does this work for webpack 4 and 5?  If so, can you amend the comment above this to `// webpack 4+` please?

src/index.ts
```diff
@@ -485,8 +451,9 @@ class ForkTsCheckerWebpackPlugin {
     );
 
     if ('hooks' in this.compiler) {
+      const hooks = ForkTsCheckerWebpackPlugin.getHooks(this.compiler);
       // webpack 4
```
johnnyreilly: If this is webpack 4/5 compatible could this comment read `// webpack 4+` please?

src/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "es6",
```
johnnyreilly: Given the minimum supported node version is 4 I think bumping the target emit from es5 to es6 is probably okay.  Are there any implications to be aware of that you can think of?

test/integration/vue.spec.js
```diff
@@ -1,71 +1,83 @@
...
-      module: {
-        rules: [
-          {
-            test: /\.vue$/,
-            loader: 'vue-loader'
+    plugin = new ForkTsCheckerWebpackPlugin(
+      Object.assign({}, options, { silent: true })
+    );
+    compiler = webpack(
+      Object.assign(
```
johnnyreilly: why the switch to `Object.assign`?  For my money that makes the code a little "noisier" 😄

src/index.ts
liangchunn: Whoops, forgot to remove this.

src/index.ts
liangchunn: From webpack's docs: https://webpack.js.org/api/compiler-hooks/#compilation , it should be `compilation`

test/integration/vue.spec.js
liangchunn: Object spread is not supported in ES6 syntax, it's ES2018+. Linting with ESLint will fail, and tests might not run with Node 6.

https://node.green/#ES2018-features-object-rest-spread-properties

johnnyreilly:
First of all, thanks for your work here; this is awesome!  I've put some adhoc comments on the PR but I've realised a few things.

First of all there's quite a lot of "noise" in this PR due to formatting changes / linting related stuff.  I'd really like to separate out the webpack 5 support changes from that.

We use prettier in ts-loader; (see the `package.json` for details: https://github.com/TypeStrong/ts-loader/blob/master/package.json - including a pre-commit hook).  

I've long thought I'd like to apply prettier in fork-ts-checker-webpack-plugin.  I'd be happy to do that, get it merged in and then have you merge that into your PR and work against that.  

Or you could apply prettier yourself in a separate PR?  

Or you could just drop all the styling changes entirely from this PR.  I honestly don't mind which; but I'd like this PR to be webpack 5 focussed only if that's okay!

Sorry - this may sound ungrateful; but I'm just trying to make it easier to focus on doing a good review and catching the right issues. At present I'm quite distracted by all the formatting changes!

johnnyreilly:
Can you confirm if this supports webpack 4 as well as 5 please?  I think it might; but I'm not certain.

src/tsconfig.json
liangchunn: I think this was a mistake, should be emitting es5 instead of es6. I added this while implementing WeakMap in the hooks.ts file.

liangchunn:
@johnnyreilly 
I've added tests for webpack#next in TravisCI. It's working with both webpack 4 and 5. I'll drop the styling changes and add in prettier and probably with lint-staged and husky in another PR.

Are there any comments on the `_pluginCompat`? I still can't find any documentation on it though.

test/integration/vue.spec.js
johnnyreilly: Shouldn't that mean they'd be failing on this already though? :smile:

johnnyreilly:
> Are there any comments on the _pluginCompat? I still can't find any documentation on it though.

Do you mean the support for webpack 2/3?

test/integration/vue.spec.js
liangchunn: Nope! Travis CI is only configured to be running on Node 8 😛

liangchunn:
Yes and no. 

If you look into this line on master (https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/src/index.ts#L189-L191), the `_pluginCompat` will never be reached if it's webpack < 4

src/index.ts
johnnyreilly: According to the docs, `compilation` runs later than `compile`; what's the rationale for switching?

johnnyreilly:
> If you look into this line on master (https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/src/index.ts#L189-L191), the _pluginCompat will never be reached if it's webpack < 4

I see. Take off and nuke `_pluginCompat`  from orbit; it's the only way to be sure. :smile: (By which I mean: feel free to remove `_pluginCompat` - it's clearly unused)

johnnyreilly:
~Can you provide some more information about `static getHooks()` please? I tried your link but it didn't have anything that seems related.~

I'm curious as to how the webpack 4 back compatibility works.

liangchunn:
@johnnyreilly 

I've been trying to understand why this change was made in webpack@5 as well... Unfortunately webpack@5 is a **major** change, most plugins _will definitely_ break, and you might want to publish a major version of this plugin with no backwards compatibility.

This is because webpack@5 is removing hooks on `compiler.hooks`. Any plugin that depends on a particular plugin, which supports webpack@5, has to be updated as well!

This means that plugins like `fork-ts-checker-notifier-webpack-plugin` will **no longer be able to tap into the events** that `fork-ts-checker-webpack-plugin` exposes.

---

**Example (https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin/blob/master/index.ts#L101):**
Tapping into `compiler.hooks.forkTsCheckerReceive` will yield nothing, since hooks are not installed into the webpack compiler instance, but rather on the static property on the class, ie: `ForkTsCheckerWebpackPlugin.getHooks()`

---

I am going to close this PR.
Since webpack@5 is still not done, we do not have any certainty on how the new plugin system works. At least we have some knowledge about it though.

johnnyreilly:
Hey @liangchunn ,

Actually - I'd quite like to leave this PR open if that's okay with you?  We're not ready to ship but I suspect we're very close!  Also, doesn't the:

```
    let hooks = compilerHookMap.get(compiler);
    if (hooks === undefined) {
```

provide the webpack 4 compatibility? That seems to be implied in this PR: https://github.com/webpack/webpack/pull/7672

Actually - I'm not totally clear; but I feel this could well become the webpack 5 support we need in future!

liangchunn:
I don't think that those lines provide webpack 4 compat. That PR changes its own HMR plugins which is shipped with webpack@5, AFAIK.

johnnyreilly:
Ah right; either way, the fact we got passing tests with all versions of webpack was very encouraging!

liangchunn:
One way we could provide webpack@5 support is to use `Object.isFrozen()` on `compiler.hooks`, and subsequently injecting hooks into `compiler.hooks` when it's not frozen, and using `getHooks` when it is frozen. I fear that this might complicated the logic behind the exposed plugins.

I'll experiment again when I have some free time on hand.

liangchunn:
I've added tests against `node6`, `node8`, and `node10` in my [own branch](https://travis-ci.com/liangchunn/fork-ts-checker-webpack-plugin/builds/87923780) and made tests againts `webpack#next` non-failing. It seeems like `webpack@5` only supports `node>=8.9.0`.

```
error webpack@5.0.0-next: The engine "node" is incompatible with this module. Expected version ">=8.9.0".
error Found incompatible module
```

johnnyreilly:
> I've added tests against node6, node8, and node10 in my own branch and made tests againts webpack#next non-failing.

Awesome!

> It seeems like webpack@5 only supports node>=8.9.0.

Yup sounds about right.

johnnyreilly:
Hey @liangchunn,

Would you be up for continuing your work now 5 alpha is out?

https://github.com/webpack/webpack/issues/8537

src/hooks.ts
```diff
@@ -0,0 +1,45 @@
...
+    forkTsCheckerServiceStart: 'fork-ts-checker-service-start',
+    forkTsCheckerReceive: 'fork-ts-checker-receive',
+    forkTsCheckerServiceOutOfMemory: 'fork-ts-checker-service-out-of-memory',
+    forkTsCheckerEmit: 'fork-ts-checker-emit',
+    forkTsCheckerDone: 'fork-ts-checker-done',
+};
+
+function createForkTsCheckerWebpackPluginHooks(): Record<keyof typeof customHooks, SyncHook | AsyncSeriesHook> {
+    return {
+        forkTsCheckerServiceBeforeStart: new AsyncSeriesHook([]),
```
sokra: You could use shorter names now, as it's already scoped to this plugin:

```suggestion
        serviceBeforeStart: new AsyncSeriesHook([]),
```

src/index.ts
```diff
@@ -156,6 +142,10 @@ class ForkTsCheckerWebpackPlugin {
     this.vue = options.vue === true; // default false
   }
 
+  static getHooks(compiler: webpack.Compiler) {
```
sokra:
```suggestion
  static getCompilerHooks(compiler: webpack.Compiler) {
```

That would be our naming convention, but technically you are free to choose whatever you like.

We have chosen this naming to allow adding hooks for other objects later, i. e. `getCompilationHooks`.

liangchunn:
@johnnyreilly Will do when I have the time!

johnnyreilly:
Awesome!

johnnyreilly:
Hey @liangchunn ,

I just tried to make your life easier by catching the branch up with upstream changes.  For some reason the GitHub UI is reporting conflicting files.... I'm not sure why; it's fine when I build locally.

Maybe it's just easier to reapply your changes (with @sokra's feedback) on a new PR.  Not sure... Apologies if I've just made things harder... 😢

liangchunn:
@johnnyreilly 

I've re-applied all my changes with the current master branch.
**This commit will be a breaking change for all other plugins that consume this plugin along with webpack 4+**

---

## Hook naming
Now for the naming, as @sokra suggested, we could rename the all the hooks since it's now scoped by default: 
```diff
src/hooks.ts
-  forkTsCheckerServiceBeforeStart: 'fork-ts-checker-service-before-start',
+  serviceBeforeStart: 'fork-ts-checker-service-before-start',
-  forkTsCheckerCancel: 'fork-ts-checker-cancel',
+  cancel: 'fork-ts-checker-cancel',
-  forkTsCheckerDone: 'fork-ts-checker-done',
+  done: 'fork-ts-checker-done',
```

Which means that tapping into the plugin's events will change (for webpack 4+):
```diff
-  compiler.hooks.forkTsCheckerDone.tap(...args)
+  const forkTsCheckerHooks = ForkTsCheckerWebpackPlugin.getCompilerHooks(compiler)
+  forkTsCheckerHooks.done.tap(...args) 
```

I've introduced a migration section in `README.md`, but I have yet to make any changes regarding the hook's naming. If you green-light the proposed changes, I'll commit accordingly.


## Tests Failures on node 6
There are some issues with the build with node 6 only:
- In `travis.yml`, `yarn add $WEBPACK $TSLOADER $VUELOADER -D` is called after `yarn install` which causes yarn to install `webpack@5.0.0-alpha` from `yarn.lock` first, and fails all builds
  - Possible solution: reorder commands in `travis.yml`
- webpack5 builds should skipped
  - Possible solution: use `matrix.exclude` in `travis.yml` as described [here](https://docs.travis-ci.com/user/customizing-the-build/#excluding-jobs)

johnnyreilly:
First of all, amazing work @liangchunn! 

To your points:

> This commit will be a breaking change for all other plugins that consume this plugin along with webpack 4+

I think that's fine. This is a major version bump for webpack. It's reasonable that we have the same here.

> Hook naming

I think renaming the hooks as suggested by @sokra is a good idea. Particularly if we're going to go for a breaking changes version bump.

Let's do it. I maintain https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin which depends upon this. I'm happy to make the required changes around hook renaming there.

> Tests Failures on node 6

I think it's time to drop support for node 6. webpack 5 drops support for it and the minimum recommended download of nodejs is node 10.  Further to that, node 6 is being end of lifed in 3 months time. 

Can you remove node 6 from the Travis test matrix please?

Also could you align the engines section of the `package.json` with the equivalent webpack 5 `package.json` please? i.e. update this:

```
  "engines": {
    "node": ">=6.11.5"
  },
```

Regarding what version number we should bump the plugin to, I've long felt that it's time this plugin should have a major version number; `1.0.0` rather than `0.6.x`

What do you think? cc @piotr-oles

liangchunn:
> Regarding what version number we should bump the plugin to, I've long felt that it's time this plugin should have a major version number; 1.0.0 rather than 0.6.x

I'd fully agree to this. Recently upgraded from `0.4.x` to `0.5.x` and it broke the build system in [typescript-node-scripts](https://github.com/liangchunn/typescript-node-scripts) (due to automatic dep upgrades from Greenkeeper), and it took me a few days to find the cause of it.

liangchunn:
@johnnyreilly 
Ready for review

In 8773e6b and af98c77, I've:
- removed node6 from `travis.yml`
- updated engines.node to `>=8.9.0` (same with webpack5)
- renamed all the hooks from `forkTsCheckerXXX` to `XXX`
- moved webpack2/3 string hooks into `hooks.ts` as `legacyHookMap`
- improved typings on WeakMap and exported hooks

johnnyreilly:
Tremendous work @liangchunn! I've hacked on your branch and:

- set the version number to `1.0.0-alpha.0`
- added a `tag: next` to the `.travis.yml` so we can publish this package with a `next` dist tag.
- updated the changelog to add details of the new version alongside the breaking changes / migration guide info which I moved from the `README.md`

I think we're pretty good to push this out as an alpha release.  Is there anything else you think we should get in there first?

Also, I've been thinking about how to upgrade [`fork-ts-checker-notifier-webpack-plugin`](https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin/blob/master/index.ts#L101) and I'm not sure how to upgrade it.  Perhaps you can advise?  The `getCompilerHooks` method is marked as `private` so I'm guessing I'm not supposed to be able to access it... Is that right?

```
   private static getCompilerHooks(compiler: webpack.Compiler) {
     return getForkTsCheckerWebpackPluginHooks(compiler);
   }
```

liangchunn:
@johnnyreilly 
> The getCompilerHooks method is marked as private so I'm guessing I'm not supposed to be able to access it... Is that right?

It should be exposed, I messed up 😅. Moved it to a public method in c410927.
Didn't notice this since test files are written in JS and there's no type checking for private method access.

> Also, I've been thinking about how to upgrade fork-ts-checker-notifier-webpack-plugin and I'm not sure how to upgrade it. Perhaps you can advise?

You would want to do:
```ts
import * as forkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'

forkTsCheckerWebpackPlugin.getCompilerHooks(compiler).tap('...', () => {})
```
EDIT: I've tried it locally and the tests seem to pass

--- 
AFAIK there isn't anything left to do, I've tested this against a real project (without tapping into hooks) and everything is working fine for me!

johnnyreilly:
That's awesome - let's get this out there! Thanks for all your hard work 🌻♥️

johnnyreilly:
Should be available on npm now - can you confirm? It's published as `@next`

liangchunn:
Yep! It's there: https://www.npmjs.com/package/fork-ts-checker-webpack-plugin?activeTab=versions

.travis.yml
```diff
@@ -2,13 +2,13 @@ language: node_js
 node_js:
  - '10'
  - '8'
- - '6'
```
alfaproject, line 5: You shouldn't really remove this if you are still officially supporting WP4.

johnnyreilly: We still officially support webpack 2 and 3 as well; following your logic we should add node 4 back into the test matrix? 😄 

In all seriousness, support is always something we take seriously.  However, given [node 6 is already in "Maintenance" and is being end of lifed in April](https://nodejs.org/en/blog/release/v6.9.0/) there seems little merit in continuing to support it.  No-one should be actively using node 6 at this stage.

alfaproject: Well, why not transpile to ES2015 then? ^_^

johnnyreilly: We do: https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/51981e79106a771b5f05eddec42c06eb13529523/src/tsconfig.json#L3

alfaproject: This is why I'm confused. You know that's not compatible with node 4, right? How can you say you are compatible with WP2? Probably no one is using in a node 4 environment and that's ok, but that seems to be just a happy accident. O_o

liangchunn: No. 

This plugin never explicitly targeted node 4 AFAIK, looking at the history of the project. 

The misconception is that node `^4.0.0` is _absolutely_ required to run this plugin alongside webpack2, which is false. **Users can freely choose which ever node versions they want to run on, as long as all the dependencies' runtime requirements are fulfilled.**

We've only recently changed to target es6 instead of es5. [Node 4 already supports 97% of es5 features](http://kangax.github.io/compat-table/es5/#node4), which was why it worked most of the time.

With the introduction of webpack5 support (with backwards compatibility of older webpack versions), we needed to bump the minimum required node version to `>=8.9.0` because webpack5 requires it.

We're talking mostly about language features; compatibility with webpack2 _doesn't mean_ that it has to 100% run on node4, rather webpack2 itself requires _at least_ node 4. So you could go off running webpack 2 on Node 11 with the latest version of this plugin with no issues ([source](https://github.com/webpack/webpack/blob/webpack-2/package.json#L72)).

johnnyreilly:
@alfaproject you might be interested to learn that `fork-ts-checker-webpack-plugin` is node 6 compatible again as of https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v1.0.0-alpha.4

We can't add it back to the travis test matrix as the project will not build on node 6 due to webpack 5 dependency issues.  It runs just fine though as long as you don't use the new `measureCompilationTime` feature which depends upon node 8+. The `engines` section has been updated back to:

```json
  "engines": {
    "node": ">=6.11.5"
  },
```

PR #173, opened 2018-10-23
Add compilerOptions option

ianschmitz:
Adds `compilerOptions` option, using [ts-loader](https://github.com/TypeStrong/ts-loader#compileroptions-object-default) as inspiration. This allows overriding `compilerOptions` of your `tsconfig.json` explicitly when initializing the plugin.

Here it is in action using [create-react-app](https://github.com/facebook/create-react-app) with the newly added TypeScript support:

image

Here is the discussion over at CRA for the background on this feature: https://github.com/facebook/create-react-app/pull/5514

**TODO:**
- [x] Unit/Integration tests?
- [ ] Validate Vue works as expected

/cc @Timer, @brunolemos

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ It helps to distinguish lints from typescript's diagnostics.
 * **tsconfig** `string`:
 Path to *tsconfig.json* file. Default: `path.resolve(compiler.options.context, './tsconfig.json')`.
 
-* **compilerOptions** `string`:
+* **compilerOptions** `object`:
 Allows overriding TypeScript options. Should be specified in the same format as you would do for the `compilerOptions` property in tsconfig.json. Default: `{}`.
 
 * **tslint** `string | true`: 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ It helps to distinguish lints from typescript's diagnostics.
 * **tsconfig** `string`:
 Path to *tsconfig.json* file. Default: `path.resolve(compiler.options.context, './tsconfig.json')`.
 
+* **compilerOptions** `string`:
+Allows overriding TypeScript options. Should be specified in the same format as you would do for the `compilerOptions` property in tsconfig.json. Default: `{}`.
+
 * **tslint** `string | true`: 
 Path to *tslint.json* file or `true`. If `true`, uses `path.resolve(compiler.options.context, './tslint.json')`. Default: `undefined`.
 
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -22,6 +22,7 @@ interface ConfigurationFile extends Configuration.IConfigurationFile {
 
 export class IncrementalChecker {
   programConfigFile: string;
+  compilerOptions: object;
   linterConfigFile: string | false;
   watchPaths: string[];
   workNumber: number;
@@ -41,6 +42,7 @@ export class IncrementalChecker {
 
   constructor(
     programConfigFile: string,
+    compilerOptions: object,
     linterConfigFile: string | false,
     watchPaths: string[],
     workNumber: number,
@@ -49,6 +51,7 @@ export class IncrementalChecker {
     vue: boolean
   ) {
     this.programConfigFile = programConfigFile;
+    this.compilerOptions = compilerOptions;
     this.linterConfigFile = linterConfigFile;
     this.watchPaths = watchPaths;
     this.workNumber = workNumber || 0;
@@ -68,8 +71,8 @@ export class IncrementalChecker {
     }));
   }
 
-  static loadProgramConfig(configFile: string) {
-    return ts.parseJsonConfigFileContent(
+  static loadProgramConfig(configFile: string, compilerOptions: object) {
+    const parsed = ts.parseJsonConfigFileContent(
       // Regardless of the setting in the tsconfig.json we want isolatedModules to be false
       Object.assign(ts.readConfigFile(configFile, ts.sys.readFile).config, {
         isolatedModules: false
@@ -77,6 +80,10 @@ export class IncrementalChecker {
       ts.sys,
       path.dirname(configFile)
     );
+
+    parsed.options = { ...parsed.options, ...compilerOptions };
+
+    return parsed;
   }
 
   static loadLinterConfig(configFile: string): ConfigurationFile {
@@ -189,7 +196,10 @@ export class IncrementalChecker {
   loadVueProgram() {
     this.programConfig =
       this.programConfig ||
-      VueProgram.loadProgramConfig(this.programConfigFile);
+      VueProgram.loadProgramConfig(
+        this.programConfigFile,
+        this.compilerOptions
+      );
 
     return VueProgram.createProgram(
       this.programConfig,
@@ -203,7 +213,10 @@ export class IncrementalChecker {
   loadDefaultProgram() {
     this.programConfig =
       this.programConfig ||
-      IncrementalChecker.loadProgramConfig(this.programConfigFile);
+      IncrementalChecker.loadProgramConfig(
+        this.programConfigFile,
+        this.compilerOptions
+      );
 
     return IncrementalChecker.createProgram(
       this.programConfig,
--- a/src/VueProgram.ts
+++ b/src/VueProgram.ts
@@ -12,7 +12,7 @@ interface ResolvedScript {
 }
 
 export class VueProgram {
-  static loadProgramConfig(configFile: string) {
+  static loadProgramConfig(configFile: string, compilerOptions: object) {
     const extraExtensions = ['vue'];
 
     const parseConfigHost: ts.ParseConfigHost = {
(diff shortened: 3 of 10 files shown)


ianschmitz:
@johnnyreilly see https://github.com/facebook/create-react-app/pull/5514#issuecomment-432092629. Do you think this PR still has value?

README.md
```diff
@@ -60,6 +60,9 @@ It helps to distinguish lints from typescript's diagnostics.
 * **tsconfig** `string`:
 Path to *tsconfig.json* file. Default: `path.resolve(compiler.options.context, './tsconfig.json')`.
 
+* **compilerOptions** `string`:
+Allows overriding TypeScript options. Should be specified in the same format as you would do for the `compilerOptions` property in tsconfig.json. Default: `{}`.
+
 * **tslint** `string | true`: 
 Path to *tslint.json* file or `true`. If `true`, uses `path.resolve(compiler.options.context, './tslint.json')`. Default: `undefined`.
```
johnnyreilly, line 67: Should the documented type of `compilerOptions` be `object` instead?

johnnyreilly:
I think the PR still has value; it's another option that people can use - it's been useful with ts-loader. Always considered it an escape hatch :smile:

A test to cover this would be awesome!

ianschmitz:
Unit tests should be good to go now

johnnyreilly:
Great! Could you increment the version in the `package.json` and put an entry in the `changelog.md` please? It makes shipping a release much easier for me!

ianschmitz:
👍

README.md
```diff
@@ -60,6 +60,9 @@ It helps to distinguish lints from typescript's diagnostics.
 * **tsconfig** `string`:
 Path to *tsconfig.json* file. Default: `path.resolve(compiler.options.context, './tsconfig.json')`.
 
+* **compilerOptions** `string`:
```
johnnyreilly: Could 
```suggestion
* **compilerOptions** `object`:
```

johnnyreilly (commented):
Could you take a look at the comments below please?

README.md
ianschmitz: You're absolutely right. Good catch! Was a late night for me yesterday 😛

src/IncrementalChecker.ts
```diff
@@ -68,15 +71,19 @@ export class IncrementalChecker {
...
+    const parsed = ts.parseJsonConfigFileContent(
       // Regardless of the setting in the tsconfig.json we want isolatedModules to be false
       Object.assign(ts.readConfigFile(configFile, ts.sys.readFile).config, {
         isolatedModules: false
       }),
       ts.sys,
       path.dirname(configFile)
     );
+
+    parsed.options = { ...parsed.options, ...compilerOptions };
```
johnnyreilly, line 84: I'm not sure if we can use object spread as it won't work on node 6 (and fork-ts-checker supports node 6).  https://node.green/#ES2018-features-object-rest-spread-properties-object-spread-properties

Can we switch to `Object.assign` please?

src/VueProgram.ts
```diff
@@ -40,6 +40,7 @@ export class VueProgram {
     );
 
     parsed.options.allowNonTsExtensions = true;
+    parsed.options = { ...parsed.options, ...compilerOptions };
```
johnnyreilly, line 43: I'm not sure if we can use object spread as it won't work on node 6 (and fork-ts-checker supports node 6). https://node.green/#ES2018-features-object-rest-spread-properties-object-spread-properties

Can we switch to Object.assign please?

src/IncrementalChecker.ts
ianschmitz: Typescript will compile it down to `Object.assign` for you as long as your target is correct. It's a little tough for me to check as I'm on my phone right now but will double check tomorrow.

ianschmitz: If you use assign elsewhere in code base I don't mind changing either way!

johnnyreilly: > Typescript will compile it down to Object.assign for you as long as your target is correct. It's a little tough for me to check as I'm on my phone right now but will double check tomorrow.

If that's the case then that's fine with me!

src/VueProgram.ts
johnnyreilly: Again - fine with this staying as is if TypeScript transpiles us to glory! 😄

Timer:
Having this for the `exclude` property would be 💯.

ianschmitz:
> Having this for the `exclude` property would be 💯.

@Timer unfortunately this doesn't work for `exclude` 😞 . See https://github.com/facebook/create-react-app/pull/5514#issuecomment-432092629

Timer:
I meant as a separate change. Is there another way to override this value, i.e. a new PR? Or is it not possible?

ianschmitz:
> Is there another way to override this value, though?

This looks interesting: https://github.com/TypeStrong/ts-loader#reportfiles-string-default. The implementation can be found here: https://github.com/TypeStrong/ts-loader/blob/35b9a5bbf298b5ea4e5b32057dbf252337b9fd08/src/utils.ts#L71

Looks like they attach to the TS diagnostics and filter out errors if the files aren't part of the `reportFiles` globs.

Timer:
Ooh, that'd be *fantastic* to implement. I don't want to ship `excludes` with CRA (in the `tsconfig` -- I'd way rather silently filter out tests).

ianschmitz:
> Ooh, that'd be _fantastic_ to implement. I don't want to ship `excludes` with CRA.

Agreed! 

@johnnyreilly what do you think about this? I haven't dug too deep into this repo to know how easy it would be to hook the after compile diagnostics.

Timer:
https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/f57ece232510f490841682d71ddad9bae0822e18/src/index.ts#L713-L760

Here's some usage, but it's littered over quite a few places. Not sure where it'd be best to filter this.

johnnyreilly:
> @johnnyreilly what do you think about this? 

Worth investigating - if it's straightforward then do it!

Timer:
@johnnyreilly could you suggest a location to perform the filtering at?

In the workers themselves, at where the worker aggregates, somewhere else? Can you please point to a file or some lines? 😄

Timer:
Also, in relation to this PR:
For an option, e.g. `module`, do you need to pass `ts.ModuleKind.ESNext` or `"esnext"`?

johnnyreilly:
Merged - thought it was good to take as is. Figure that the `reportFiles` merits it's own PR :smile:

ianschmitz:
> Also, in relation to this PR:
> For an option, e.g. `module`, do you need to pass `ts.ModuleKind.ESNext` or `"esnext"`?

Very good point @Timer. I think that may be a deficiency. Just looking through https://github.com/TypeStrong/ts-loader/blob/d49dfbaa7e35c728090554d47a122ad2571cb09c/src/config.ts#L66, they apply the transformation before running it through `ts.parseJsonConfigFileContent()`. Should be an easy change in this repo as we just need to break out the `readConfigFile()` into a separate variable and merge in the options.

To clarify, we should not require the user to provide the value of `ts.ModuleKind.ESNext` imo. `"esnext"` should be sufficient, but that would require modifying the result of `readConfigFile()`.

Sorry about that!

ianschmitz:
@johnnyreilly do you want to hold off publishing until we merge in the fix?

johnnyreilly:
@ianschmitz - too late! https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v0.4.12

Still, it's an extra option which people won't be using yet so no harm done - if you could submit a PR for the fix that'd be appreciated!

johnnyreilly:
> @johnnyreilly could you suggest a location to perform the filtering at?
>
> In the workers themselves, at where the worker aggregates, somewhere else? Can you please point to a file or some lines? 😄

@Timer - I don't have a strong opinion. I have a feeling that experimenting during implementation should be a good guide. I'd expect this to be a fairly simple tweak, not requiring of masses of code; if you find yourself writing War and Peace then you might want to reconsider where you're doing the filtering :smile:

Essentially, go with your gut.

ianschmitz:
@johnnyreilly no problem. I'll have a PR up tomorrow with the fix!

johnnyreilly:
Awesome!

johnnyreilly:
@Timer - sorry about the delay in replying!

The simplest implementation would seem to be taking this approach here: 

https://github.com/TypeStrong/ts-loader/blob/d49dfbaa7e35c728090554d47a122ad2571cb09c/src/utils.ts#L54-L75

and adding a `filter` in before the `forEach` here:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/458cd748519c04501298c342de31acf1b2f84336/src/index.ts#L741

This would be handling it in the worker aggregate position.  It's possible that there might be a minor performance benefit to doing this in the worker itself but I'm not sure how big the gain would be. For now, I'd take simplicity.

PR #175, opened 2018-10-25
chore(deps): update lodash, mocha away from vulnerable versions

StephM-W:
### Why
* lodash modularized (all versions) are vulnerable to https://nvd.nist.gov/vuln/detail/CVE-2018-3721 and are unmaintained (~2 years)
* mocha ^3.4.1 is vulnerable to https://nodesecurity.io/advisories/534 and transitively to https://nodesecurity.io/advisories/146

### What
* Upgraded lodash to tree-shakeable 4.17.11
* Updated lodash requires from lodash.xxx to lodash/xxx
* Updated tslint to whitelist lodash submodules
* Upgraded mocha to ^5.2.0
* Updated npm test script to add --exit mocha option (to prevent hanging, as per docs, this was a breaking change from 3 -> 4+)
* Updated tsconfig to add es2016 lib for Map and Set


johnnyreilly:
Thanks for this! A couple of questions:

> Updated tsconfig to add es2016 lib for Map and Set

`Map` and `Set` are part of es2015 I believe. Would it make more sense to just change `es2015.core` to `es2015` and drop `es2016`?

StephM-W:
> Thanks for this! A couple of questions:
> 
> > Updated tsconfig to add es2016 lib for Map and Set
> 
> `Map` and `Set` are part of es2015 I believe. Would it make more sense to just change `es2015.core` to `es2015` and drop `es2016`?

Works for me - patch updated 😄

johnnyreilly:
Awesome!

Regarding the switch to using lodash full rather than submodules... I'm curious to know whether using submodules directly is no longer possible or if it's just that you prefer this approach?

Ultimately it doesn't matter too much I suspect as fork-ts-checker-webpack-plugin should only ever be a devDependency. Nevertheless I'm curious

StephM-W:
I think this is the direction the lodash maintainers have taken, in fact in [v5](https://github.com/lodash/lodash/issues/3838#issuecomment-398592530) I understand they'll remove submodules entirely (everything will [import from the root namespace](https://github.com/lodash/lodash/wiki/Roadmap#v500-2018)) and depend on babel and webpack wizardry.

If the CVE hadn't been raised, I think most folk would have happily left things as they were - as it stands, those packages are used in so MANY places, it's quite daunting... and I'm just helping update packages I depend on, there's thousands of projects out there.

johnnyreilly:
Great - works for me! Thanks for doing this!

PR #179, opened 2018-10-28
Add support for reportFiles option

### LINKED ISSUE #177: `reportFiles` as an `exclude` alternative
@Timer said here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/173#issuecomment-432803475

> Ooh, that'd be fantastic to implement. I don't want to ship excludes with CRA (in the `tsconfig` -- I'd way rather silently filter out tests).

> @johnnyreilly could you suggest a location to perform the filtering at?

> In the workers themselves, at where the worker aggregates, somewhere else? Can you please point to a file or some lines? 😄 

That's a merged PR and so it's less visible.  I didn't want it to get lost as I think it's a **great** suggestion.  😄  Do take a look at the conversation there.

cc @ianschmitz

ianschmitz:
Closes #177.

This PR adds a new option `reportFiles` that behaves the same way as as [ts-loader](https://github.com/TypeStrong/ts-loader#reportfiles-string-default). When an array of file path globs are provided, it will use these globs to filter the ts/tslint diagnostics to only include those matching.

Here is an example using create-react-app where there are no errors in the `reportFiles` (you can see both `__tests__/App.tsx` and `App.test.tsx` have type errors):
image

And an example when there is an error in one of the `reportFiles` (which is src/App.tsx in this case):
image

/cc @Timer, @johnnyreilly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.4.14
+
+* [Add support for `reportFiles` option](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/179) (#179)
+
 ## v0.4.13
 
 * [Merge in `compilerOptions` prior to calling `parseJsonConfigFileContent`](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/176) (#176)
--- a/README.md
+++ b/README.md
@@ -79,6 +79,14 @@ List of typescript diagnostic codes to ignore.
 * **ignoreLints** `string[]`: 
 List of tslint rule names to ignore.
 
+* **reportFiles** `string[]`: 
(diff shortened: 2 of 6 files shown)


johnnyreilly:
Brilliant! Thanks! ❤️

Timer:
You've been killing it with these features, @ianschmitz! Thank you!

mix0000:
Where this options is gone??? My setup dont works withour `reportFiles`

piotr-oles:
Please use `issues` hook to filter-out issues that you're not interested in: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#plugin-hooks :)

mix0000:
Yes, last night i did it with: 
```javascript
issue: {
  exclude: {
    file: "node_modules/**/*"
  }
}
```

dasler-fw:
Can't find any migration guides. Can someone make it please?

PR #182, opened 2018-11-13
chore: cleanup package and dependencies - change compile target to es6

SimonSchick:
Things done:

- Apply strict null checks to make sure I don't break stuff :)
  - Some null assertions were used, see comments
- Bump up build target to es6
  - This is fully supported by node v6
- Replaces `lodash` with native methods
  - should work according to https://node.green/
  - please note that node v6 is not in the build matrix
- Remove unused `resolve` dependency
- Got rid of (presumably) accidentally packaged package tar file
- Got rid of `.npmignore` as it is redundant since you use `files` in package.json
- Update dependencies and apply `lib`s to fix tests
  - they were failing from the start for me


src/VueProgram.ts
```diff
@@ -234,7 +234,7 @@ export class VueProgram {
     );
   }
 
-  private static getScriptKindByLang(lang: string) {
+  private static getScriptKindByLang(lang?: string) {
```
johnnyreilly, line 237: Is this optional?

johnnyreilly:
Thanks for this @SimonSchick! I'm a little hesitant about the targeting ES6 purely because it requires that all consumers need to be on > node 6.4.

Mind you, for webpack 4 that ship sailed back in February so it's probably fine. And node 10 is now the LTS install. Yeah it's probably fine.

@piotr-oles do you have any thoughts?

src/formatter/codeframeFormatter.ts
```diff
@@ -21,17 +21,18 @@ export function createCodeframeFormatter(options: any) {
...
+      fs.existsSync(file) &&
+      fs.readFileSync(file, 'utf-8');
     let frame = '';
 
     if (source) {
       frame = codeFrame(
         source,
-        message.line,
-        message.character,
+        message.line!, // Assetions: `codeFrame` allows passing undefined, typings are incorrect
```
johnnyreilly: Should this read "Assertions"?

src/tsconfig.json
```diff
@@ -1,14 +1,17 @@
...
+    "target": "es6",
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "suppressImplicitAnyIndexErrors": true,
-    "strictNullChecks": false,
-    "lib": ["es5", "es2015", "dom"],
+    "strictNullChecks": true,
```
johnnyreilly: If you want to go strict, why not save typing and use `strict` instead?

https://www.typescriptlang.org/docs/handbook/compiler-options.html

SimonSchick:
@johnnyreilly The package does specify a minimum node version.

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/97c86d66f1b870898414e32e8fc777e7253d9104/package.json#L40

or is that considered none binding?

In any case you can release it as `0.5` when it doubt.

johnnyreilly:
I think it's probably fine - I've already done this myself here: https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin/blob/master/tsconfig.json

I'd like to hear what @piotr-oles thinks about this but I think this looks generally fine. Could you address my comments please?

johnnyreilly:
Hi @SimonSchick,

Please could you address my review comments? Thanks

SimonSchick:
@johnnyreilly I don't see any review comments, are you sure you've submitted them?

johnnyreilly:
How about now?

src/tsconfig.json
SimonSchick: I do this in personal projects but I'd rather not do this in others as new TS releases may add new strict checks which might cause build failures, this happened in the past.
If you don't mind this I'll gladly change this.

src/VueProgram.ts
SimonSchick: See comment at 245 ```// when lang is "js" or no lang specified```

src/formatter/codeframeFormatter.ts
SimonSchick: Yup, typo :)

SimonSchick:
I'd consider all comments addressed.

src/tsconfig.json
johnnyreilly: New build failures welcome 😁

In all seriousness, we do welcome new build failures as it's likely a sign of a bug the compiler has spotted. And in any case, we use yarn to manage our dependencies so upgrading to the newest TypeScript is something you have to actively do.

SimonSchick: Interesting, the usual attitude I got was to avoid breaking builds whenever possible :P
I will update this :)

SimonSchick:
@johnnyreilly done

johnnyreilly:
Any reservations about the `target: es2015` stuff @piotr-oles?

SimonSchick:
I'd also like to point out again that node v6 is **not** in the build matrix and I haven't tested my changes on node v6, I validated my changes again node.green however.
If you want to support v6 in the future I can ammend it to this PR, otherwise we might as well skip to the  es2017 build target.

SimonSchick:
@johnnyreilly thoughts?

johnnyreilly:
I was hoping @piotr-oles would comment but I guess he's busy. I think I'm fine with this though. I'd like to release it as 0.5 just in case there's any old node users out there.

Do you want to update the `package.json` and the `CHANGELOG.md` please?

SimonSchick:
I bumped up dev deps and added ts-loader@5 to the test matrix as well.

Also do you have any plans to drop webpack 2-3 since v5 is around the corner?

.travis.yml
```diff
@@ -7,6 +7,7 @@ install:
  - yarn add $WEBPACK $TSLOADER $VUELOADER -D
  - yarn lint
 env:
+ - WEBPACK=webpack@^4.0.0 TSLOADER=ts-loader@^5.0.0 VUELOADER=vue-loader@^15.2.4
  - WEBPACK=webpack@^4.0.0 TSLOADER=ts-loader@^4.3.0 VUELOADER=vue-loader@^15.2.4
```
johnnyreilly: Do you want to drop ts-loader 4.x from the test matrix; whilst technically breaking changes, 4-5 was not a change that could ever affect fork-ts-checker-webpack-plugin; there's no need to test ts-loader 4.x anymore

johnnyreilly:
> Also do you have any plans to drop webpack 2-3 since v5 is around the corner?

Not as long as supporting both remains straightforward (and so far it does).

Also, I'm not that comfortable with the dropping of support for node 6 without @piotr-oles commenting first I'm afraid.

SimonSchick:
All good, was just wondering if there were any plans :)

package.json
```diff
@@ -37,7 +37,7 @@
     "webpack-plugin"
   ],
   "engines": {
-    "node": ">=6.11.5"
+    "node": ">=8.9.0"
```
piotr-oles: Why do you want to drop support for node 6? Our 2 major peer dependencies supports:
webpack 2.3.0 -> node 4.3.0
typescript 2.3.0 -> node 0.8.0

Even if we would drop support for webpack 2 and 3, webpack 4 still supports node >= 6.11.5 :) For me it makes more sense to drop webpack 2 support than node 6 :)

src/IncrementalChecker.ts
```diff
@@ -211,8 +214,8 @@ export class IncrementalChecker {
       this.programConfig,
       path.dirname(this.programConfigFile),
       this.files,
-      this.watcher,
-      this.program
+      this.watcher!,
```
piotr-oles, line 196: Why do you have to put here exclamation marks? I think we are missing if statement or typing of VueProgram is incorrect :)

src/IncrementalChecker.ts
```diff
@@ -227,19 +230,19 @@ export class IncrementalChecker {
...
+      this.program!
     );
   }
 
-  hasLinter() {
-    return this.linter !== undefined;
-  }
-
   getDiagnostics(cancellationToken: CancellationToken) {
+    const { program } = this;
```
piotr-oles, line 218: It looks like the only place in this class where you use destructuring. For me it's not consistent :)

piotr-oles (changes requested):

package.json
SimonSchick: As explained above, v6 isn't in the build matrix, when I asked @johnnyreilly he did suggest I bump it up, I don't have a problem reverting that change :)

src/IncrementalChecker.ts
SimonSchick: Does it matter? It's just as good as `const program = this.program`

src/IncrementalChecker.ts
SimonSchick: I did it here because `loadVueProgram` is called by `nextIteration` which ensures that `watcher` is set, it felt kinda silly to add an assertion, I can change that if you want.

SimonSchick: I used null assertion in lots of other places btw, mostly because you rely on lots of env variables, I.
I wanted to enable the lint rule that bans null assertions but I'd have to add a LOT of if statements 😛

src/IncrementalChecker.ts
piotr-oles: I'm not saying that it's worse or better, just inconsistent :)

package.json
piotr-oles: Please revert this change - and could you add v6 to the build matrix? :)

src/IncrementalChecker.ts
piotr-oles: Yes, I saw that you add these assertions - thank you for this - it will make it more error proof. Please add this assertion or make it private method if it's possible. Currently it's public method - someone can call this method and get weird error :)

piotr-oles: Yes, I saw that you add these assertions - thank you for this - it will make it more error proof. Please add this assertion or make it private method if it's possible. Currently it's public method - someone can call this method and get weird error :)

SimonSchick:
@piotr-oles I went ahead and added public/private *eveywhere* and enabled the corresponding lint rule.

While doing so I also noticed that the properties of `NormalizedMessage` were accessed sometimes directly and indirectly (via getter), I marked all properties as `public readonly` instead, might even give a speed boost :)

SimonSchick:
Also please note, adding node v6 to the build matrix doubled the job count, adding v10 will triple it, you might want to consider dropping wp 2/3 in master and maintain them in a separate branch.

piotr-oles:
Thank you @SimonSchick for your work. I'm aware that it will make builds a lot longer but it shouldn't be an issue in terms of development. You can still run your test locally on one node version and in 99% cases it will pass on all node versions :)

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## v0.5.0
+ * Removed unused dependency `resolve`.
+ * Replace `lodash` usage with native calls.
+ * Test against ts-loader v5
+ * **Breaking changes**:
+   * Dropped support for node v6, minimum is now v8.9
```
piotr-oles: Please remove this point - we didn't drop support :)

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## v0.5.0
```
piotr-oles: If we don't drop node v6, maybe it shouldn't be a major release?

src/IncrementalChecker.ts
```diff
@@ -303,7 +282,7 @@ export class IncrementalChecker {
       cancellationToken.throwIfCancellationRequested();
 
       try {
-        this.linter.lint(fileName, undefined, this.linterConfig);
+        this.linter!.lint(fileName, undefined!, this.linterConfig); // REVIEW: this shouldn't work?
```
piotr-oles: what do you mean by "this shouldn't work?"?

CHANGELOG.md
SimonSchick: This is true, although I don't know what the next version will be, will rename to `next`.

src/IncrementalChecker.ts
SimonSchick: `undefined!` 😛

SimonSchick: Improved the comment.

piotr-oles (approved):
Ok, everything is fine - thank you :)

SimonSchick:
Let me just fix the build :)

SimonSchick:
So turns out es2015 !== es6 build target 😅

SimonSchick:
@piotr-oles you should look into porting tests to TS too, turns out tests don't work in node v6.

SimonSchick:
Sorry for the mess, I'm in the process of moving and haven't had a whole lot of time to work on this more than 5min in a block.

johnnyreilly:
This is why "squash and merge" was invented :wink:

SimonSchick:
@johnnyreilly builds now~~ 🎉

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## next
```
johnnyreilly: Please could you change `## next` to `0.5.0` and put an entry detailing the changes made to the `NormalizedMessage` please? That will affect anyone who supplies their own formatter (eg Create React App)

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.4.15",
```
johnnyreilly: Please could you update the version to `0.5.0`? Thanks

johnnyreilly:
Great - I've put a few more comments around the `package.json` and `changelog.md`. if you could address these then I think we're good to go!

SimonSchick:
@johnnyreilly I don't see any new comments, did you forget to request the changes? Happens to me all the time 😛

johnnyreilly (commented):
How about now? :smile:

johnnyreilly:
Now?

package.json
SimonSchick: I suggest you use a tool such as https://github.com/sindresorhus/np to do that, manually bumping versions has burned me in the past,

johnnyreilly: That looks cool - I'd like to look into that in the future. Not right now though! (There's already more in this PR than I'm entirely comfortable with.)

SimonSchick:
I think I got everything now, on a intercontinental flight atm, might not respond immediately.

johnnyreilly:
Happy landings! https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v0.5.0

Thanks for your help ❤️

PR #183, opened 2018-11-14
Update dependecies and some fixes for compatibility with the latest npm packages/libraries.

NickMaev:
* Updated dependencies due to deprecation (for compatibility with the latest npm packages/libraries, for ex. [antd](https://github.com/ant-design/ant-design/)).
* Replaced "require" by "import" (for compatibility with EcmaScript module output).
* Added ESM support (for compatibility purposes).


.gitignore
```diff
@@ -24,4 +24,8 @@ package-lock.json
 
 # Editor directories and files
 .idea
-.vscode
\ No newline at end of file
+.vscode
+/.vs/slnx.sqlite
```
SimonSchick: Should use `.vs` only

SimonSchick:
So this would be a breaking change, correct?

.gitignore
NickMaev: Agreed

SimonSchick:
Also I'm curious: What is the use case for `module`?
Iirc node esm modules are still pretty niche.

NickMaev:
> Also I'm curious: What is the use case for `module`?
> Iirc node esm modules are still pretty niche.

Didn't understand you, rephrase please. Where did you see `module` in my code?

johnnyreilly:
Closing due to inactivity

PR #190, opened 2018-12-13
Working as expected on first line diagnostic.start

radiosterne:
If Typescript diagnostic starts on line 0, fork-ts-checker-webpack-plugin as of 0.5.1 fails and throws an error, which it shouldn't. This humble and small PR fixes it :) I'll update package.json and changelog.md if needed.

### DIFF
--- a/src/CancellationToken.ts
+++ b/src/CancellationToken.ts
@@ -4,6 +4,8 @@ import * as os from 'os';
 import * as path from 'path';
 import * as ts from 'typescript';
 
+import { FsHelper } from './FsHelper';
+
 interface CancellationTokenData {
   isCancelled: boolean;
   cancellationFileName: string;
@@ -46,7 +48,7 @@ export class CancellationToken {
     if (duration > 10) {
       // check no more than once every 10ms
       this.lastCancellationCheckTime = time;
-      this.isCancelled = fs.existsSync(this.getCancellationFilePath());
+      this.isCancelled = FsHelper.existsSync(this.getCancellationFilePath());
     }
 
     return this.isCancelled;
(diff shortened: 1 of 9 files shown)


src/NormalizedMessage.ts
```diff
@@ -51,7 +51,7 @@ export class NormalizedMessage {
     let character: number | undefined;
     if (diagnostic.file) {
       file = diagnostic.file.fileName;
-      if (!diagnostic.start) {
+      if (typeof diagnostic.start === 'undefined') {
```
johnnyreilly: Could this be `diagnostic.start === undefined` instead please? Marginally more performant and my personal preference.

johnnyreilly (commented):
Thanks for contributing! Could you get the tests to pass please?

johnnyreilly:
Oh and yes updating `package.json` and `CHANGELOG.md` appreciated!

radiosterne:
@johnnyreilly looks like the tests were not passing in master as well due to `fs.existsSync` changes in node v10. Given that `existsSync` is deprecated, I took the liberty of creating a helper to emulate it :)

johnnyreilly:
Thanks!

johnnyreilly:
Should be released in 30 mins or so - thanks for the contribution! 🌻♥️

src/CancellationToken.ts
```diff
@@ -46,7 +48,7 @@ export class CancellationToken {
     if (duration > 10) {
       // check no more than once every 10ms
       this.lastCancellationCheckTime = time;
-      this.isCancelled = fs.existsSync(this.getCancellationFilePath());
+      this.isCancelled = FsHelper.existsSync(this.getCancellationFilePath());
```
SimonSchick, line 51: `existsSync` is NOT deprecated!
See https://nodejs.org/api/fs.html#fs_fs_existssync_path

johnnyreilly: Oh that's correct! Would you care to submit a PR reverting that change?

PR #169, opened 2018-10-16
Support linting + formatting with pre-commit hook

liangchunn:
## Description
This PR enforces linting with `prettier`, `eslint` and `tslint` for the project by leveraging pre-commit looks with `husky` and `lint-staged`

- Upgraded `tslint`, `eslint` packages
- Added `prettier`, `tslint-config-prettier` (to disable conflicting rules), `husky`, and `lint-staged` packages
- Ran `prettier` for files matching `src/**/*.{ts,json}` and `src/**/*.js` files once
- Hand off `eslint` and `tslint` styling rules, like spacing, and semicolons, to `prettier` with a pre-commit hook **(needs review)** 
- Run `yarn lint` in CI

I've tested the pre-commit hooks, and they seem to be working fine.

PR #194, opened 2018-12-23
refactor(object-spread): use spread instead of assign

liangchunn:
Just following up from #166 and #169:

Since we've already dropped support for Node 6 in #166, object spread properties can now be safely used. (As briefly discussed [here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/166#pullrequestreview-164500790))

In `.eslintrc.json`, `"experimentalObjectRestSpread": true` and `"ecmaVersion": 6` should be kept, since Node 8 supports 99% of ES2015 features, and object-spread is just a ES2018 feature, but is [fully supported in Node 8](https://node.green/#ES2018-features-object-rest-spread-properties) and above. (Also mentioned in #169, [here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/169#pullrequestreview-165328521))

---

Note: This PR does not introduce any functionality changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.0.0-alpha.1
+
+* [Use object-spread instead of `Object.assign`](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/194) (#194)
+
 ## v1.0.0-alpha.0
 
 * [Add support for webpack 5](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/166)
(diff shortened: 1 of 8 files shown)


johnnyreilly:
Nice - either way; we transpile typescript down to es2015 so the `object-spread` can happily be used since (I believe) it gets transpiled into `Object.assign` anyway.  But yeah - completely happy with these changes.  Do you want update the version number in `package.json` / add an entry to `CHANGELOG.md` and I'll merge this.  This can go out as `alpha.1` I guess.

liangchunn:
> we transpile typescript down to es2015 so the object-spread can happily be used since (I believe) it gets transpiled into `Object.assign` anyway.

Yep! This only impacts test files, and just making them look nicer. Either way we could push this as alpha, or not at all. (Up to you, I don't really mind 😄)

johnnyreilly:
Thanks!

Issue #212, opened 2019-02-02
silent option silences errors and warnings


lizzzp1:
I'm trying to silence only `WARNINGS` from tslint. Adding `silent` true will silence errors as well. Is there a way to specify only silencing warnings and not errors?

Thanks.

johnnyreilly:
Hey @lizzzp1

There's `ignoreLints` but that requires you specify the lints you want to ignore. It seems like a potentially useful feature though; do you fancy submitting a PR?

lizzzp1:
@johnnyreilly yeah, I can give it a shot! - Thanks for the quick reply.

lizzzp1:
@johnnyreilly what is the preferred api/option for this?

johnnyreilly:
Maybe a new option called `ignoreLintWarnings`? Default value of `false`. Try it out - see how it feels.

PR #213, opened 2019-02-06
ignoreLintWarnings option

lizzzp1:
Attempts to address: https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/212
Adds a new option `ignoreLintWarnings`.

Question:
Should this have a test?

Thanks!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -38,13 +38,13 @@ interface Options {
   async: boolean;
   ignoreDiagnostics: number[];
   ignoreLints: string[];
+  ignoreLintWarnings: boolean;
   reportFiles: string[];
   colors: boolean;
   logger: Logger;
   formatter: 'default' | 'codeframe' | Formatter;
   formatterOptions: any;
   silent: boolean;
-  ignoreLintWarnings: boolean;
   checkSyntacticErrors: boolean;
   memoryLimit: number;
   workers: number;
@@ -85,10 +85,10 @@ class ForkTsCheckerWebpackPlugin {
   private watch: string[];
   private ignoreDiagnostics: number[];
   private ignoreLints: string[];
+  private ignoreLintWarnings: boolean;
   private reportFiles: string[];
   private logger: Logger;
   private silent: boolean;
-  private ignoreLintWarnings: boolean;
   private async: boolean;
   private checkSyntacticErrors: boolean;
   private workersNumber: number;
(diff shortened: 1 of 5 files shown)


johnnyreilly:
> Should this have a test?

Yes definitely 😄 

Take a look at the current tests here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/tree/master/test

Hopefully there's a natural place for a new test to "fit" - unit or integration test I'm not sure.  Hopefully an obvious candidate should present itself.

lizzzp1:
@johnnyreilly added test. Let me know if it's not correct.

johnnyreilly:
Thanks! I'll take a look soon

johnnyreilly:
Hey @lizzzp1

I just wanted to check that the functionality in your PR not superseded by this PR:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/214

I kind of think they're different enough but I wanted to double check with you.

lizzzp1:
Hi @johnnyreilly, yeah that does look to have different functionality.

johnnyreilly:
Cool - I'd like to get yours in first. Could you update the `package.json` and `CHANGELOG.md` please?

src/index.ts
```diff
@@ -838,7 +841,7 @@ class ForkTsCheckerWebpackPlugin {
           (this.lints || []).concat(this.diagnostics).forEach(message => {
             const formattedMessage = this.formatter(message, this.useColors);
 
-            message.isWarningSeverity()
+            message.isWarningSeverity() && !this.ignoreLintWarnings
               ? this.logger.warn(formattedMessage)
```
johnnyreilly: I think this might be an issue in the case where you have a warning and you are ignoring warnings - it looks like an error would be logged in that case ..

lizzzp1: ah. So int that case I would want to do nothing here, right?

johnnyreilly: Yeah I think so

lizzzp1: Ok, updated. Let me know if that looks correct to you.

src/index.ts
```diff
@@ -808,6 +811,20 @@ class ForkTsCheckerWebpackPlugin {
     return function noopEmitCallback() {};
   }
 
+  private getLoggerMessage(
```
johnnyreilly: Thanks for this change - could I request one more tweak to avoid confusion please?

The name and the return type of the method suggests a logger message is passed back. In reality no value is ever returned I think? Could this be tweaked to make intention a bit clearer please?

lizzzp1: Oops yes, you're totally right. Will update.

lizzzp1:
~~@johnnyreilly what needs to be updated in package.json?~~ I see

johnnyreilly:
Awesome - thanks for your work!

Once the CI is happy I'll look to merge this and get a release out. 🤗

PR #215, opened 2019-02-18
Format code samples

stackdumper:
- indent with 2 spaces
- set highlight language to javascript

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,11 +174,11 @@ This plugin provides some custom webpack hooks (all are sync):
 ## Vue
 1. Turn on the vue option in the plugin in your webpack config:
 
-```
-    new ForkTsCheckerWebpackPlugin({
-      tslint: true,
-      vue: true
-    })
+```js
+new ForkTsCheckerWebpackPlugin({
+  tslint: true,
+  vue: true
+})
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -213,31 +212,31 @@ import Hello from '@/components/hello.vue'
 ```
 4. Add rules to your `tslint.json` and they will be applied to Vue files. For example, you could apply the Standard JS rules [tslint-config-standard](https://github.com/blakeembrey/tslint-config-standard) like this:  
 
-```json
```
johnnyreilly: Isn't json  the correct choice here?

stackdumper: I think it's better to keep it consistent and use `js` even here, to make code highlight identical in all code samples where JSON is involved.

johnnyreilly: Where it is JSON rather than JS isn't highlighting as JSON a more helpful choice?  The distinction seems like it could be useful..

stackdumper: Sure it could. However, in this case the only distinction between `js` and `json` is the color of colons, so..

Can you change it to `json` or should I add another commit? 🙂

stackdumper: 65b2503 🙂

PR #223, opened 2019-03-03
fix `useTypescriptIncrementalApi` flag to work according to docs

### LINKED ISSUE #222: bug: `useTypescriptIncrementalApi` flag is not true by default, even if TS Version > 3
The bug is caused by wrong `this.options.useTypescriptIncrementalApi` ([source code here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/03c612ed742e6e6002c32cce2ddfef5bdba7b220/src/index.ts#L572)):

```
USE_INCREMENTAL_API:
            this.options.useTypescriptIncrementalApi === true,
```

the fix should be:

```
USE_INCREMENTAL_API:
            this.useTypescriptIncrementalApi === true,
```

yanivefraim:
`useTypescriptIncrementalApi` doesn't work as supposed to - it is not true by default, for TS version > 3.

fixes #222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.0.0-alpha.10
+
+* [Fix incremental api to work with TS 3+ by default](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/223)
+
 ## v1.0.0-alpha.9
 
 * [Default to incremental api usage to true if TS 3+](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/217)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fork-ts-checker-webpack-plugin",
-  "version": "1.0.0-alpha.9",
+  "version": "1.0.0-alpha.10",
   "description": "Runs typescript type checker and linter on separate process.",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
--- a/src/index.ts
+++ b/src/index.ts
@@ -205,12 +205,13 @@ class ForkTsCheckerWebpackPlugin {
 
     this.validateVersions();
 
+    this.vue = options.vue === true; // default false
+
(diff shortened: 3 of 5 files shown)


johnnyreilly:
First of all, thanks for this - good catch!

There's now some failing tests which I believe are related to https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/219

i.e. the incremental compiler has some issues with Vue which we hope will be solved by this PR:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/220

~In the meantime, you could fix it by forcing the Vue tests that are failing to *not* use the incremental API. That would work for now. Sound good?~

Actually - better idea: could you apply the changes from this commit to your PR: https://github.com/Realytics/fork-ts-checker-webpack-plugin/commit/90836f2d330024e2e1c7796ffffb2d0ceaf788a5 (apart from the reasons.md that is). That should stop the incremental API being used and the tests should pass.

yanivefraim:
> Actually - better idea: could you apply the changes from this commit to your PR: 90836f2 (apart from the reasons.md that is). That should stop the incremental API being used and the tests should pass.

@johnnyreilly - I cherry-picked 90836f2d330024e2e1c7796ffffb2d0ceaf788a5, without the readme part, wasn't sure about tests, so I just kept them...

johnnyreilly:
Awesome. Thanks!

Could you increment the version in the `package.json` and put an entry in the `CHANGELOG.md` please? I'll then merge and push out a release

yanivefraim:
> Could you increment the version in the package.json and put an entry in the CHANGELOG.md please? I'll then merge and push out a release

@johnnyreilly  - done

johnnyreilly:
Thanks!

johnnyreilly:
Will go out with https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v1.0.0-alpha.10

Thanks again!

Issue #224, opened 2019-03-04
Allow to trigger a typecheck for "virtual files" from outside?


phryneas:
Hi there!

I'm trying to find a way to trigger a typecheck for `.mdx` files. Those files have to be transpiled to js/ts first using `mdx-loader`. After that, I would like to typecheck the generated source code.

My idea was to put a loader behind `mdx-loader` and from there somehow signal `fork-ts-checker-webpack-plugin` to check the transpiled source code.

Is this somehow possible?

Currently I'm doing that typecheck in a self-written loader, but as I don't really know a lot about webpack and tsc, this is very inefficient.

(I also opened an issue in the docz project, because that would be the other side of what I want to do. In the end, docz's .mdx-files could be typesafe: https://github.com/pedronauck/docz/issues/664 )

johnnyreilly:
Wow that's an unexpected request!

I expect that you could do worse than look at how we handle Vue files. It's not exactly the same use case but there's some similar work at play

phryneas:
@johnnyreilly Thanks for the pointer - I didn't think about looking at the vue-specific-code. We have slackday on friday, so I'll try and spend that day trying to figure out how everything works.

One thing though: could you give me a few hints on how fork-ts-checker works in general? That could save me a few hours - I don't really know how the plugin knows which files to validate. Does it start working from the webpack entry points, or does is monitor which files are opened by webpack somehow? 
Stuff like that - every little bit helps, no need to be too extensive :)

johnnyreilly:
Sure. For speed reasons the plugin doesn't get triggered by webpack. A file change should trigger a change. There's 2 ways the plugin can work:

1. File watching mode - watches files and triggers compilation (`useTypescriptIncrementalApi: false`).

2.  Using TypeScripts watch API (`useTypescriptIncrementalApi: false`).

If experimenting I'd start with approach 1. As it's probably easier to get going with

phryneas:
Thanks :)

PR #225, opened 2019-03-08
add support for type-checking .mdx files

phryneas:
This allows for typechecking .mdx files for use with systems like `docz`. (see #224 and https://github.com/pedronauck/docz/issues/664 )

To use it with docz, add the following to the `doczrc.js`:

```javascript
import ForkTsCheckerPlugin from 'fork-ts-checker-webpack-plugin';

export default {
    modifyBundlerConfig: config => {
        config.plugins.push(
            new ForkTsCheckerPlugin({
                mdx: true,
                compilerOptions: {
                    noImplicitAny: false
                }
            })
        );

        return config;
    }
};
```

Note: as far as I am aware, this will not work for `.mdx` files if they are imported from another file as I did not transfer the `resolveNonTsModuleName` method from `VueProgram.ts`.
Apparently, it is also possible to use `.mdx` files to create components that are imported from normal TypeScript files. 

This is something that would not be done in the context of `docz` which I am using - in docz, an mdx file is always a top level file. I don't know if anyone out there really uses it that way, so I did not implement it.

### DIFF
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -18,6 +18,8 @@ import * as minimatch from 'minimatch';
 import { VueProgram } from './VueProgram';
 import { FsHelper } from './FsHelper';
 import { IncrementalCheckerInterface } from './IncrementalCheckerInterface';
+import { MdxProgram } from './MdxProgram';
+import { ProgramType } from './ProgramType';
 
 export class IncrementalChecker implements IncrementalCheckerInterface {
   // it's shared between compilations
@@ -59,7 +61,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
     private workNumber: number = 0,
     private workDivision: number = 1,
     private checkSyntacticErrors: boolean = false,
-    private vue: boolean = false
+    private type: ProgramType = ProgramType.typescript
   ) {
     this.hasFixedConfig = typeof this.linterConfigFile === 'string';
   }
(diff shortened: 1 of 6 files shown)


johnnyreilly:
Well done for putting something together that seems to meet your needs! I really appreciate the PR but I'm somewhat hesitant about adding direct support for `.mdx` files specifically.

I don't really want to keep expanding a switch statement and adding more and more program types to the plugin if possible as it creates more and more code to maintain.

I'd still like to help if it's feasible, but I wonder if we could think about creating a more "pluggable" approach, whereby people could supply their own Program to the plugin. 

That way a thousand flowers could bloom without bloating the fork-ts-checker-webpack-plugin codebase.

What do you think?

phryneas:
I have to admit I kind of had the same thought, but if I had followed it to the end, that would also somehow question the existing vue support, which was nothing I wanted to suggest from the get-go, so I submitted it as-is for now and wanted to wait for your response.

I think something pluggable would be quite easy to realize.
`fork-ts-checker-webpack-plugin` could export an interface like this:
```typescript
interface PluggableProgramFactory {
  createProgram(config: {
    typescript: typeof ts,
    configFile: string,
    compilerOptions: object
    files: FilesRegister,
    watcher: FilesWatcher,
    oldProgram: ts.Program
  }): { programConfig: ts.ParsedCommandLine; program: ts.Program};
  watchExtensions: string[];
}
```
that would need to be the top-level-export of any npm module to serve as a pluggable program.

The config could then just get the name of the npm module, pass it via environment variable to the service, be required there & the createProgram method could be invoked there (in place of `loadDefaultProgram`).

Is that in line with what you had in mind?

johnnyreilly:
> Is that in line with what you had in mind?

Pretty much - apart from this:

> The config could then just get the name of the npm module, pass it via environment variable to the service, be required there 

I'd have the pluggable program as an option that you pass to the plugin on initialisation. i.e. you'd do it in the `webpack.config.js` file.

Do you see what I mean? We already have the ability to supply custom formatters to the plugin - imagine this being the logical equivalent

phryneas:
I think I just wrote this bad - I _wanted_ to say that it could be passed as a plugin constructor option.
All the rest was just because I wanted to highlight that it had to be a string, otherwise it wouldn't be able to be passed over the  `fork` border (at least I don't see a way, maybe I'm missing something).
The formatters can be a function, but they never leave the main process so they don't have to be passed around, do they?

johnnyreilly:
Yup - you're right. 

The new process / thread mechanism means that you're going to be working with unfortunate constraints.

Your mission (should you choose to accept it :smile: ) is to try and solve this as best you can. Experimentation encouraged!

phryneas:
Sounds like a fun quest. I'll give it a go in a separate PR ;)

phryneas:
well, see #227 ;)

PR #231, opened 2019-03-16
WIP: typescript sys proxy

phryneas:
Well, I had teasered this in #227 and I guess it's better reviewable if I have an open PR for this branch ;)

So far, this seems to work transparently with `useTypescriptIncrementalApi: true` and the module resolution is working just fine as well with the latest commit.

I'm currently testing this with `.mdx` files, but there should not be anything stopping this to work with `.vue` as well.
I think I'm gonna do some testing with  `useTypescriptIncrementalApi: false` and get that to work.

From there on, I'm gonna wait for some feedback from you:

* How would you structure this? (everything in one wrapTypeScript.ts is not really feasible - would splitting off a wrapperUtils.ts and a wrapCompilerHost.ts be enough?)
* Should this replace the .vue code?
* Do you want to keep the .vue and .mdx specific code in this repo (now that they're going to be only about 20 lines per extension) or do you still want to have a pluggable approach (which would most likely be more code than the extension-specific handling code itself).
* Overall: what do you think of this? Is this even feasible?


johnnyreilly:
You have been busy 😄

> How would you structure this? (everything in one wrapTypeScript.ts is not really feasible - would splitting off a wrapperUtils.ts and a wrapCompilerHost.ts be enough?)

Splitting it seems like a good idea.

> Should this replace the .vue code?

Possibly - I'd be interested to see how that looks.

> Do you want to keep the .vue and .mdx specific code in this repo (now that they're going to be only about 20 lines per extension) or do you still want to have a pluggable approach (which would most likely be more code than the extension-specific handling code itself).

I'm not decided on this yet. My current feeling is that Vue should remain in the box at least for now since it's already there. But I'm unlikely to add anything else - including mdx I'm afraid. I'm still keen on a pluggable solution. 

> Overall: what do you think of this? Is this even feasible?

I think this is potentially very interesting. I'm afraid I'm occupied with a number of other things at present and so haven't devoted too much thought to it. But I think it has merit. 

In terms of sequencing I'd like to see the Vue incremental fixes land first: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/220

Essentially I want to get existing users to a good place before looking at changing the underlying architecture and potentially open up the possibility of new users.

I think the work you're doing is very cool though - and I'm grateful for the effort you're putting in 🌻❤️😄

phryneas:
Well, now you can take a look at how it looks with .vue ^^
Essentially, all you need is a config like specified here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/12e85371d95c858cf82054948d8c312476a1bd1c/src/wrapTypeScript.ts#L19
That just references this loader: https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/12e85371d95c858cf82054948d8c312476a1bd1c/src/handleVueContents.ts#L4

But right now, there are some things now working yet and tests are failing.

From what I have gathered so far what goes wrong:
* as I just attach .tsx on an early file system level to all vue files without looking at the file contents, every vue file is regarded as .tsx, no special handling for .ts is possible :cry:  Guess I need to figure out a smart way around that *sigh*
* module resolution still has it's kinks (the test with the @-namespace is failing for example)
* generally, a lot of .vue-tests are failing. I haven't looked at all of them yet.

phryneas:
Okay, apparently nothing is too bad with my module resolution - it's just that there seems to be an implicit mapping for a virtual "@"-module in vue.

Adding something like
```
baseUrl: './src',
        paths: {
          '@/*': ['./*']
        }
```
to the compilerOptions resolves that problem.

Now, I don't know vue myself. Is that something that should implicitly be added?

johnnyreilly:
> Now, I don't know vue myself. Is that something that should implicitly be added?

Also not a Vue user I'm afraid; don't know 😥

phryneas:
It seems like that "@"-wildcard was added by @CKGrafico - can you tell us if that should be there implicitly or if that is something a user would have added anyways to their .tsconfig to be able to use it?

phryneas:
Also, `ts-loader` (as well as apparently babel-loader?)  is going with two configuration options (appendTsSuffixTo, appendTsxSuffixTo), to determine which vue files are ts and which ones are tsx - I think I will go for a similar approach for the time being.

Or is there something actually requiring this auto-detection that is currently in there?

johnnyreilly:
> Or is there something actually requiring this auto-detection that is currently in there?

Not to my knowledge. If memory serves @herringtondarkholme and I came up with this when Vue support was added to ts-loader. See https://github.com/TypeStrong/ts-loader/issues/270 and https://github.com/TypeStrong/ts-loader/pull/354

phryneas:
Okay, then I'll go with that for now. Makes it a little less intimidating ;)

phryneas:
For now, I've added the '@' namespace to the tsconfig of the associated test - I guess that is sensitive.
Also, I've associated .vuex files with .tsx and .vue files with .ts.

This makes all but three tests green: the tests that now complain that `example-js.vue`, `example-jsx.vue` and `example-nolang.vue` now also report errors.

An alternative solution to using file extensions would be to preprend files from `handleVueContents.ts` with their filetype like `// @fork-ts-checker-handle-file-as (ts|tsx|js|jsx|json|none)`.

Checking & using that in IncrementalChecker would be no problem as the CompilerHost is wrapped there anyways and `host.getSourceFile` could easily be overridden, in ApiIncrementalChecker this would require to start wrapping the CompilerHost. Not the end of the world, but it still looks so prisine - so I'll leave that approach open to discussion :grin: 

Aside from that, everything should be working now (?) - it would be cool if an actual vue user could test this, as I only have the test cases ;)

johnnyreilly:
> This makes all but three tests green: the tests that now complain that example-js.vue, example-jsx.vue and example-nolang.vue now also report errors.

Might be worth tweaking the tests to get them to get more information....

I don't get why some Vue tests would pass and some would fail. Is there a pattern? Looks like ts and tsx are fine whilst js / jsx are not. That's interesting

phryneas:
No, that's pretty much expected behaviour at this point. As all .vue files are now handled  as either .tsx or .ts, the three files containing non-ts-javascript now throw type errors that would have been ignored before. So the tests actually do what they are supposed to do ;)

I guess this could only be mitigated by actually handling the files by their contents, not their extension - which is why I suggested that `@fork-ts-checker-handle-file-as` file header comment to be injected.

But I believe won't get around to implement that this weekend - maybe next weekend :)

johnnyreilly:
> I guess this could only be mitigated by actually handling the files by their contents, not their extension - which is why I suggested that @fork-ts-checker-handle-file-as file header comment to be injected.

Maybe that's a good shout. If memory serves @herringtondarkholme had some reservations about the appendTsxSuffixTo approach.

Thanks for your continued experimentation 😄

johnnyreilly:
I've just tweaked your branch so it runs the Vue tests when `useTypescriptIncrementalApi` is both `true` and `false` - remember to `git pull` 😄

phryneas:
I just did some reorganization - this way there should be less changes to the original files, everything is a little more transparent.

On the part of filetype-autodetection, I'm having a little bit of a blocker. For the IncrementalChecker, that won't be a problem - I can simply hook into host.getSourceFile and set the correct scriptKind. But I don't see a similar way for the ApiIncrementalChecker - do you have any idea how I could change the scriptKind there? (except changing the file extension dynamically by file content, I'd like to avoid that)

PS: #219 does that by hooking into host.fileExists and checking the contents, deciding depending on the contents if the file "should exists" - of course I could do something like that, but then every file would get parsed at least twice. So I'd love a more elegant solution :/

phryneas:
Okay... I could hook into `typescript.createEmitAndSemanticDiagnosticsBuilderProgram` and wrap the `host` parameter there once more, from there I can wrap `getSourceFile` even for ApiIncrementalChecker. Not very elegant but seems to be working. I've gotta give this some thought. And if you have a better idea - I'd be happy to hear it :)

johnnyreilly:
> Not very elegant but seems to be working. I've gotta give this some thought. And if you have a better idea - I'd be happy to hear it :)

If it's a straight choice between not working and inelegant but working I'll always go for the second option 😄

johnnyreilly:
BTW - is this is what MDX is? https://gatsby-mdx.netlify.com Pretty cool!

Up until now I'd thought it was this: https://en.m.wikipedia.org/wiki/MultiDimensional_eXpressions - without having the first idea how that could be so 😁

phryneas:
Oh, lol, I didn't even know MultiDimensional eXpressions ^^

Yup, that is the mdx I was talking about. Markdown with JSX. Also used in https://github.com/jxnblk/mdx-deck and https://www.docz.site/ - and even currently being implemented to be a part of storybook.

phryneas:
FYI: I just noticed that your commit to run tests with `useTypescriptIncrementalApi` true|false is pretty pointless: Later in `helpers.createVueCompiler`, a `IncrementalChecker` is created manually, so most tests run as if useTypescriptIncrementalApi were false. Would be cool if you could look into this :)

I'm out for now - long day ;)

johnnyreilly:
>  I just noticed that your commit to run tests with useTypescriptIncrementalApi true|false is pretty pointless: Later in helpers.createVueCompiler, a IncrementalChecker is created manually, so most tests run as if useTypescriptIncrementalApi were false. 

I guess you mean this:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/2c03b7e45a9fc023efa3053403cc05ecff9146c0/test/integration/helpers.js#L59

This was added when initial Vue support for the incremental API was being implemented:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/commit/53d690b85c7f0956b5c2909ee323178d2da90adb

@0xorial can you recall what the purpose of this  was please? Presumably we shouldn't need this? This line seems particularly interesting:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/2c03b7e45a9fc023efa3053403cc05ecff9146c0/test/integration/helpers.js#L77

0xorial:
I've looked through the history and creation of `IncrementalChecker` seems to be there from [very beginning](https://github.com/Realytics/fork-ts-checker-webpack-plugin/commit/cda43412#diff-027e4e7ef48b3f7f654e6a4280ee736aR62). I merely moved it into another file (probably because I had an intention to make 2 separate test sets for incremental=true/false), but I did not pay close attention back then I presume... So, unfortunately, I have no idea why is it there. 
It have a feeling that author understood 'integration test' as 'any test that covers more than 1 file/class' while currently they are more like 'test that tests whole plugin as a black box'

johnnyreilly:
Cool - so probably we can drop it.

phryneas:
@johnnyreilly the tests are now running with the real checker from the webpack instance - but, in the same process, not in a forked one. 
I guess this is why that IncrementalChecker was added there in the first place.

Problem is: the tests still rely heavily on implementation details of IncrementalChecker that are not met by ApiIncrementalChecker like `checker.program` and in some cases can't even be met 1:1, like `checker.programConfig.fileNames`. 
I'll try and see what I can do (might need to rewrite the Checkers to match at least partially a common interface), but no promises here.

phryneas:
And most of the tests are actually green now.

But I seem to have some regression regarding #94 - and I'm currently clueless as to why `checkSyntacticErrors` does not seem to have an effect.

phryneas:
Okay, this took me a LOT of time: with `useTypescriptIncrementalApi: true`, as soon as a syntactic error is found, it completely stops type-checking the rest. So, if `checkSyntacticErrors = false`, it just returns zero errors and if it is true, it returns at most one (the syntactic error itself).
I guess this is a limitation of `TypescriptIncrementalApi` and these two tests can just never test anything meaningful in that mode?

johnnyreilly:
Aha! I think you've just been on the same learning journey @0xorial and went on when support was originally added: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/198#issuecomment-453812729

To quote myself:

Actually - I've dug into this myself.  I follow what you meant.  The project context includes a semantic error (`index.ts`) and a syntactic error (`syntacticError.ts`).  The test simply checks that there is any error surfaced.  There is different behaviour for `useTypescriptIncrementalApi: true` and `useTypescriptIncrementalApi: false` but that is *OKAY*. 😄 

In fact even good.

#### `useTypescriptIncrementalApi: true`

The compiler detects errors in both `index.ts` and `syntacticError.ts`.  The compiler has found a syntactic error and so doesn't bother to go on to surface the semantic error (for whatever reason - as you say likely performance).  But because we're not reporting syntactic errors we never get to hear about it; and our test fails.

#### `useTypescriptIncrementalApi: false`

Same story as with `true` but there's no short circuiting (that we're aware of) happening in the compiler in this case so we hear about our errors and our test passes.

-------

I think what we should probably do is amend our test pack so we either expect different behaviour depending upon the mode or (perhaps this is better) write tests which should pass consistently regardless of mode.

Essentially, if a single error is found then that's great. We're much more interested in the difference between 0 and 1 errors being found in a test in comparison with say 7 and 8 errors being found.

Do you see what I mean?

johnnyreilly:
> Essentially, if a single error is found then that's great. We're much more interested in the difference between 0 and 1 errors being found in a test in comparison with say 7 and 8 errors being found.

When I get a moment I'll try and improve the test pack to reflect this.

phryneas:
Just as a heads-up: I've not given up on this or anything. I just have a few turbulent weeks right now and don't have much time to work on this - it'll get better in 2-3 weeks.

But besides fixing up the last tests and maybe moving "fs" to a wrapped dependency injection for `IncrementalChecker` (so that [this line](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231/files#diff-e27b56838bc2968a5affe095bbd08efeR118) won't be necessary, keeping all the wrapping magic apart from the implementation itself), I think this PR is pretty much approaching a 'ready' state.
In the end, I'd remove the mdx-specific stuff and follow this up with a second PR for plugability.

So if you could @ a few vue-users to maybe give this a real-life test in the meantime, I guess it is already pretty much ready for that.

johnnyreilly:
> Just as a heads-up: I've not given up on this or anything. I just have a few turbulent weeks right now and don't have much time to work on this - it'll get better in 2-3 weeks.

Don't worry - life happens and there's no rush.

> So if you could @ a few vue-users to maybe give this a real-life test in the meantime, I guess it is already pretty much ready for that.

Happy to do that - recommend you first put a branch together which includes the `lib` folder on GitHub. This will allow people to test easily. We've done this in the past and provided instructions to make it easier for users to test (and hence more likely that it will happen.)

There's an example of what I mean here: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

@WolfspiritM I wonder if you might be able to give this a test? This might give us the incremental API fixes you were attempting in https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/220. Our difficulty right now is that neither I or @phryneas are Vue users so we need help!

johnnyreilly:
@phryneas I've just made change to the vue tests.  Essentially this change starts testing generated errors for the specific types of errors they represent; rather than an arbitrary count.  This is a more targeted testing approach and I think I want to roll this out wider than just the vue tests; but your work gave me a good chance to test the technique out.  

In other news, when I run tests locally I'm getting out of memory errors:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF66F72F04A v8::internal::GCIdleTimeHandler::GCIdleTimeHandler+5114
```

Any thoughts on this? It's the first time I've encountered it in `fork-ts-checker-webpack-plugin`

phryneas:
Yeah, I encountered that one, too.
The thing is - the tests aren't forking any more, everything is executed on the same process to get access to the _actual_ IncrementalChecker used by the plugin. (this wouldn't be possible if it were on another thread, that's why before the IncrementalChecker was created manually, not making it a real integration test).

As a result, the forked processes aren't exiting, and no memory is being freed that way. So they hog up on memory. 
But as far as I see, there is no other way of freeing memory, at least fork-ts-checker has no real mechanism for this.
And apparently, mocha is running all tests in the same process, too, so everything hogs up even more memory.
My only ideas to solve this would either be a cleanup method in the service, or to switch to a testing framework like jest that executes every test in it's own process.

johnnyreilly:
> The thing is - the tests aren't forking any more, everything is executed on the same process to get access to the actual IncrementalChecker 

I didn't quite understand what you meant by this... Could you try and explain some more?

> My only ideas to solve this would either be a cleanup method in the service, or to switch to a testing framework like jest that executes every test in it's own process.

I had an idea that we could try and force garbage collection between tests; see here: https://nodesource.com/blog/request-garbage-collection-for-node-js-processes-with-n-solid

Do you reckon that'd work?

phryneas:
> > The thing is - the tests aren't forking any more, everything is executed on the same process to get access to the actual IncrementalChecker
> 
> I didn't quite understand what you meant by this... Could you try and explain some more?
> 

I'll try :)

Take a look at `test/integration/helpers.js`:

Before my changes, the `createVueCompiler` function did two things:
* create a webpack instance
* create an instance of `IncrementalChecker`

In reality, half of the tests were not using the webpack instance, but the checker instance. Also, that checker instance was created slightly different than the actual checker instance created in service.ts, so this was more of a checker unit test than a real integration test. (and never tested the ApiIncrementalChecker)

Now, what I wanted to do was
* create a webpack instance
* return the checker instance that is used by that webpack instance

Essentially, that was impossible: as webpack forks a new process for the service, which then generates the `IncrementalChecker`, the checker is on another process and cannot be accessed from the test.  
So, I prevented the plugin from actually forking. In https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231/files#diff-b9e343b630cc669a829ec4b8417a5e10 , I fake `child_process.fork` to execute the code in the current process instead of spawning a new process. Also, I temporarily override `process.exit` to prevent the `service.ts` from killing the test itself (which is now the same process as the service).
That way, the test can access the checker that is actually being used by webpack - which is now running on the same process.
Suddenly, the tests also check ApiIncrementalChecker.

The downside: where before, calling `process.exit` just cleaned up all memory used by the service, it now does nothing. That's why, from test to test, more and more memory is being consumed. There's just no proper cleanup.

> > My only ideas to solve this would either be a cleanup method in the service, or to switch to a testing framework like jest that executes every test in it's own process.
> 
> I had an idea that we could try and force garbage collection between tests; see here: https://nodesource.com/blog/request-garbage-collection-for-node-js-processes-with-n-solid
> 
> Do you reckon that'd work?

I guess setting `global.checker` to `undefined` and then running garbage collection could work. Would you give it a try?


PS: I really appreciate you fiddling around with this, so yeah, of course you can push to this branch all you want :)

johnnyreilly:
Interesting! So the existing API incremental tests aren't really doing what we hope?  Great that you've found a solution; it's a shame it's not quite a test of what the users will experience as it's going to be in the same process.... But given the existing tests are really false positives this is demonstrably better!

> Essentially, that was impossible: as webpack forks a new process for the service, which then generates the IncrementalChecker, the checker is on another process and cannot be accessed from the test.

I wonder if there's a way to tweak the checker to expose the checker purely for the purposes of testing... That could be potentially useful - it wouldn't need to be something we exposed always; we could have a "test" mode or something... Ideas!

> I guess setting global.checker to undefined and then running garbage collection could work. Would you give it a try?

Yeah - not sure when but I'll try a few things out and let you know how I get on :smile:

johnnyreilly:
Tried it; no joy alas 😢 

Also tried upping `--max_old_space_size`:

```
    "test:integration": "node --max_old_space_size=8192 --expose-gc node_modules/mocha/bin/mocha -gc -R spec ./test/integration --exit && rimraf tmp",
```

It didn't work either 😢

phryneas:
Thanks to #246, the tests are working correctly now.

For testers, I published a build of this as https://www.npmjs.com/package/@phryneas/experimental-fork-ts-checker-webpack-plugin

johnnyreilly:
> For testers, I published a build of this as https://www.npmjs.com/package/@phryneas/experimental-fork-ts-checker-webpack-plugin

That's awesome. I'd like to publicise this to see if we can get some beta testers on board. I'll tweet about this later and mention a bunch of people in the comments here as well. 

Before I do that I just wanted to check: 
- do Vue users have to use the `useTypescriptIncrementalApi` flag to activate the incremental API for Vue? 

- aside from Vue users I guess we're interested to know generally if this has any impact to users generally. Functionality should be unchanged given our tests be passing. We've got no performance tests though. (If anyone ever gets a moment we'd love some to be added. I believe @jantimon has some that could be used for inspiration here: https://github.com/namics/webpack-config-plugins 😁) So a general check from people anecdotally that perf doesn't suffer on their own projects is probably useful?

phryneas:
>     * do Vue users have to use the `useTypescriptIncrementalApi` flag to activate the incremental API for Vue?

Yes, along with `vue: true` - it would also be useful to check with useTypescriptIncrementalApi: false if everything works correctly, because a lot changed there, too. 

>     * aside from Vue users I guess we're interested to know generally if this has any impact to users generally. Functionality should be unchanged given our tests be passing. We've got no performance tests though. (If anyone ever gets a moment we'd love some to be added. I believe @jantimon has some that could be used for inspiration here: https://github.com/namics/webpack-config-plugins grin) So a general check from people anecdotally that perf doesn't suffer on their own projects is probably useful?

I guess that couldn't hurt.

johnnyreilly:
Hello Vue users!

We're experimenting with a pluggable architecture for the `fork-ts-checker-webpack-plugin`.  However, the reason that's super interesting to you in particular is that along with these architectural changes we think we've been able unlock proper use of the TypeScript incremental API.

**However, we're not Vue users and we need your help!  Would Vue users out there be able to try out the following experimental version of the plugin:**

https://www.npmjs.com/package/@phryneas/experimental-fork-ts-checker-webpack-plugin

We're mostly interested in your experience with `useTypescriptIncrementalApi` set to `true`.  If you could try that out and report back we'd be *so* grateful.  It would be ❤️ 

In addition we're somewhat interested in your experience with `useTypescriptIncrementalApi` set to `false`. cc @CKGrafico @prograhammer  @wonderful-panda @ktsn @yyx990803 @LinusBorg

LinusBorg:
I'll try and give this a spin in the coming days, thanks for your work!

johnnyreilly:
@phryneas I seem to be having some problems with the `fork-ts-checker-notifier-webpack-plugin` which plugs into this https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin

I think this may mean that the integration hooks are in some way broken.... Alas I think we might be short of tests for that.  You should hopefully be able to reproduce this using this example with your tweaked plugin:

https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin

Also, with the experimental checker I'm not getting the TSLint issues being reported anymore.

Can you take a look please?  

It looks like other functionality is present and correct though 😄  I'll continue testing...

phryneas:
@johnnyreilly the problems seems to be that fork-ts-checker-notifier-webpack-plugin tries to include the original plugin by name, which is different. As a solution, just

```
yarn add fork-ts-checker-webpack-plugin@https://registry.yarnpkg.com/@phryneas/experimental-fork-ts-checker-webpack-plugin/-/experimental-fork-ts-checker-webpack-plugin-0.0.2.tgz
```

oh, and that one is a new build, with all new changes from worker-rpc and master.

phryneas:
Oh, @johnnyreilly could you please give the travis a kick? It seems the tests for this branch aren't running any more, and now that they're fixed I'd like to see the results for all the webpack/node versions :)

johnnyreilly:
I can see Travis running - is that not what you're seeing? 

Also, when using the experimental version today I experienced missing tslint messages consistently and missing error messages occasionally.

I can try the new build tomorrow. Have you experienced this?

phryneas:
Ah, now Travis is running. And it seems I got some errors due to the merge. Gotta investigate that.

As for your experiences - I currently don't have a big project with fork-ts-checker (we are running babel and `tsc --watch` in the current project as the vscode integration is more fluent that way - remember, I started all this off with the intent of checking mdx in docz :wink:), just some small test project with about three files (and thus, no tslint).

But it seems I need to set up a big test project. Do you have any open source projects in your mind that I could clone for testing?

phryneas:
Okay, apparently updating typescript from `3.3.3333` to `3.4.2` makes all test fail :(

I've gotta give this some more time, but not today. I've taken free days for the rest of the week though, so I think I'll get some hours in there in the next few days.

johnnyreilly:
> But it seems I need to set up a big test project. Do you have any open source projects in your mind that I could clone for testing?

Great question! And no... We should ask the internet?

phryneas:
> Great question! And no... We should ask the internet?

If you could ask your twitter, that would be great. I don't have that big of a following -.-

johnnyreilly:
> I've gotta give this some more time, but not today. I've taken free days for the rest of the week though, so I think I'll get some hours in there in the next few days.

Don't worry - take your time ❤️

Also, perhaps one of the dependant packages on npm represents a big project you could test with: https://www.npmjs.com/browse/depended/fork-ts-checker-webpack-plugin

It can't all be `react-scripts` and dev tools. (Well, maybe!)

zackschuster:
can i request that the mdx support be moved to a plugin? [according to packagephobia](https://packagephobia.now.sh/result?p=@mdx-js/mdx@1.0.10) this would add ~8mb to the dep tree. for reference, [fork-ts-checker itself is less than half that size](https://packagephobia.now.sh/result?p=fork-ts-checker-webpack-plugin@1.0.3)

phryneas:
@zackschuster before this gets merged, the mdx part will definitely be moved out again and a second PR with some kind of plugin functionality will be opened. this is currently just in there because I have no vue codebase to test ;)

zackschuster:
@phryneas thank you!

phryneas:
Okay, so the recent errors with TS 3.4 are fixed.

@johnnyreilly going forward with this - I would wait for #249 to be merged and as a next step open a new PR to get the corrected vue tests into master, just to reduce the sheer size of this PR as I think this is getting way too big. Is that okay with you?

Then I'd clone myself https://github.com/desktop/desktop and try to reproduce your missing error messages.

phryneas:
@johnnyreilly I did some experimentation and I could not reproduce your missing lints/diagnostics. Could you give me some more information on your setup? Like webpack/tslint/typescript version and plugin config?

I noticed something else, though: when working with typescript 2.5, my modifications did not work. Apparently, some internal APIs I was using have not been in existance for as long as I was hoping. Also, older typescript code seems to do much more internal reference to `typescript.sys` instead of getting it passed in via DI.

As a consequence, I replaced the functions on the actual imported object instead of wrapping that one - that approach seems to work quite well so far. 

I tweaked the travis config so it should now (for this branch) execute the tests for all minor typescript versions from 2.1 to 3.4 - I'm curious what more that will uncover. We'll see.

johnnyreilly:
> Could you give me some more information on your setup? Like webpack/tslint/typescript version and plugin config?

My setup is quite like this: https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin

> I noticed something else, though: when working with typescript 2.5, my modifications did not work. Apparently, some internal APIs I was using have not been in existance for as long as I was hoping. 

Good luck with your experimentations. I'm open to dropping support for older versions of TypeScript  should it prove useful. `ts-loader` dropped support for < 2.4 relatively recently which allowed us to simplify the codebase. This may not prove necessary but it's on the table if we need it

phryneas:
@johnnyreilly I got some failures, and as a comparison I am currently running the tests (depending on TS version without `useIncrementalApi`) against the current `master` branch. From the looks of it, at least some features are broken on old TS versions :disappointed: 

You can check it out here: https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/builds/522770455

Edit: I forgot to disable the useIncrementalApi tests for the vue code in that run - new run is incoming here:
https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/builds/522776393

phryneas:
I'll open an issue for this. ( -> #257 )

phryneas:
So after thinking about this some time I've put more work into this.

I've removed the code that was faking the `typescript.sys` object and taken a different approach for this:
Now I'm replacing the `fs` module with a fake one using `mock-require` which we had as a devDependency until now anyways. (We could also use something like monkey-fs, but those don't really do something different...)
That fake file system now modifies all paths passed to `fs` functions if they have specific extensions, just like I was doing it in the wrapped `typescript.sys` before. (X.vue <=> X.vue.__fake__.ts etc.)

That way, there is no special handling for this in our code any more, and almost no additional wrapping of TypeScript interna is required.

So far, this looks very promising. I will, however, have to take a look at the watching code, as currently that will emit events with un-faked filenames. 
I believe it might be pretty simple to get this to work with TypeScript itself, as according to #256 it internally uses `fs.watch` and `fs.watchFile` - but our own code is using `chokidar`, which handles this differently. On the other hand, if we were to drop the old watch mode, supporting this might not even be necessary.

Just to give you a short report on what's happening here :)

johnnyreilly:
Awesome work @phryneas!

zackschuster:
~~github notified me of this because i was apparently mentioned somewhere, but i can't find the reference.~~ nvm found it

thanks for the hard work, @phryneas!

phryneas:
So, I think I've got the file watching situation under control, at least for the `incrementalApi: true` mode. Also, I've added a pretty extensive unit test for the filesystem wrapper.

I've created a new build that can be used with
```bash
yarn add fork-ts-checker-webpack-plugin@https://registry.yarnpkg.com/@phryneas/experimental-fork-ts-checker-webpack-plugin/-/experimental-fork-ts-checker-webpack-plugin-0.0.3.tgz
```

As for me, I'll start dogfooding this version, but as I'm not a vue user myself I won't really see much of a change (and I still don't know any vue open source project already using fork-ts-checker-webpack-plugin to test this on). It could be great if a few vue users  (@zackschuster ? :smile:  ) could give this a go :)

PS: looks like the tests are breaking in the CI sometimes. file event watching is a flimsy thing, especially in a CI :(

PPS: the tests are failing for mac exclusively - but I think the code itself is right, only some mac-specifics I don't know yet in the tests. Gonna get a colleague with a Mac to run them for me next week.

johnnyreilly:
Awesome work! @zackschuster would you be able to give this a try please?

Likewise I'll start dogfooding when I'm back working on front end stuff again (hopefully in a week 🤞)

zackschuster:
@johnnyreilly @phryneas i don't currently have any vuejs projects that use typescript but i'll give it a shot when i'm back on my dev machine.

are there any known concerns i can give special attention to?

johnnyreilly:
Thanks @zackschuster! It would be great to know if `experimentalWatchApi: true` and `experimentalWatchApi: false` still work as expected. Also that a production build still works with Vue.

Anything else @phryneas ?

phryneas:
@johnnyreilly I think you mixed the `experimentalWatchApi` up with `ts-loader`, in `fork-ts-checker-webpack-plugin` it's `incrementalApi`  😋

@zackschuster Thanks a lot!
It would be great if you could test it with `incrementalApi: true` and `incrementalApi: false`.
Typechecking in general, but especially if .vue files are being watched. 
Vue with `incrementalApi: true` wasn't working before and should be working now, but on the downside this patch might be breaking file watching with `incrementalApi: false`. I'm especially interested if that's the case (also, if it really works with `true` of course :smile: )

There's one minor breaking change I can think of: Before, vue files were seen by the typescript compiler as `filename.vue.ts`, now they're seen as `filename.vue.__fake__.ts` - this should not change anything if your `tsconfig.json` is just including all `.ts` files, but if it only includes specific files you might need to change that there.

WolfspiritM:
Hi, Sorry that I wasn't able to continue on the vue fix but I gave this a try on our (not open source, sorry) project.
## TL;DR
1. Fastest was this experimental plugin without incrementalApi.
2. Linter won't be able to show codeframe anymore due to the "fake" name I think.
3. With incrementalApi enabled there is a HUGE (as in VERY HUGE...6 Minutes compared to 20 Seconds without incrementalApi) performance drop on initial build and also on incremental builds.
4. With incrementalApi enabled Linter errors are not reported immediatly but instead the compilation is done within milliseconds (I assume it fails silently?) after changing something else the compilation takes longer and finally reports the error.

For incremental building I changed a string in a *.vue file. Nothing big and always the same one.

### Config
```
                tslint: true,
                vue: true,
                useTypescriptIncrementalApi: ....,
                async: false,
                formatter: "codeframe",
                measureCompilationTime: true,
            }),
```
## Tests
### "fork-ts-checker-webpack-plugin" (watch)
##### Configuration: useTypescriptIncrementalApi: false
- initial build:
  ```Starting type checking and linting service...
  Using 1 worker with 2048MB memory limit
  webpack is watching the files…
  Compilation took: 22557.66 ms.```
- incremental build 1: `Compilation took: 9706.11 ms.`
- incremental build 2: `Compilation took: 5526.23 ms.`
- incremental build 3: `Compilation took: 5517.79 ms.`
- syntax error build:
	 ```
	  Compilation took: 5711.98 ms
	ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts
	./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& (C:/src/Project/node_modules/ts-loader??ref--2-0!C:/src/Project/node_modules/vue-loader/lib??vue-loader-options!./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts&)
	[tsl] ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts(87,29)
	      TS1005: ',' expected.
	 @ ./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& 1:0-193 1:209-212 1:214-404 1:214-404
	 @ ./scripts/components/Admin/AdminManager.vue
	 @ ./scripts/app.ts
	 @ multi ./scripts/app.ts
	 ```
 - lint error build:
   ```
	  Compilation took: 5689.21 ms.
	  ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	  ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	  95:1 Consecutive blank lines are forbidden
	      93 |     // tslint:enable:object-literal-sort-keys
	      94 | 
	    > 95 | 
	         | ^
	      96 |     private created() {
	      97 |         this.loadAll({from: 0, count: 20});
	      98 |     }
	 ```
- type error build:
	```
	Compilation took: 6052.88 ms.
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	95:13 Type '0' is not assignable to type 'string'.
	    93 |     // tslint:enable:object-literal-sort-keys
	    94 | 
	  > 95 |     private shouldBeString: string = 0;
	       |             ^
	    96 | 
	    97 |     private created() {
	    98 |         this.loadAll({from: 0, count: 20});
	```


### "@phryneas/experimental-fork-ts-checker-webpack-plugin" (watch)
##### Configuration: useTypescriptIncrementalApi: false
- initial build:
  ```Starting type checking and linting service...
  Using 1 worker with 2048MB memory limit
  webpack is watching the files…
  Compilation took: 19435.08 ms.
  ```

- incremental build 1: `Compilation took: 4081.24 ms`
- incremental build 2: `Compilation took: 3112.74 ms`
- incremental build 3: `Compilation took: 2911.64 ms`
- syntax error build: 
	 ```
	Compilation took: 3098.1 ms.
	ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts
	./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& (C:/src/Project/node_modules/ts-loader??ref--2-0!C:/src/Project/node_modules/vue-loader/lib??vue-loader-options!./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts&)
	[tsl] ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts(87,26)
	      TS1005: ',' expected.
	 @ ./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& 1:0-193 1:209-212 1:214-404 1:214-404
	 @ ./scripts/components/Admin/AdminManager.vue
	 @ ./scripts/app.ts
	 @ multi ./scripts/app.ts
	 ```
- lint error build:
	``` 
	Compilation took: 2393.22 ms.
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue.__fake__.ts
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue.__fake__.ts
	95:1 Consecutive blank lines are forbidden
	```
- type error build:
	```
	Compilation took: 2815.3 ms
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	95:13 Type '0' is not assignable to type 'string'.
	    93 |     // tslint:enable:object-literal-sort-keys
	    94 | 
	  > 95 |     private shouldBestring: string = 0;
	       |             ^
	    96 | 
	    97 |     private created() {
	    98 |         this.loadAll({from: 0, count: 20});
	```

### "@phryneas/experimental-fork-ts-checker-webpack-plugin" (watch)
##### Configuration: useTypescriptIncrementalApi: true

- initial build:
	```
	 Using 1 worker with 2048MB memory limit
	 webpack is watching the files…
	 Compilation took: 400250.35 ms.
	```
- incremental build 1: `Compilation took: 11017.46 ms.`
- incremental build 2: `Compilation took: 16267.02 ms.`
- incremental build 3: `Compilation took: 10982.65 ms.`
- incremental build 4: `Compilation took: 10485.2 ms.`
- incremental build 5: `Compilation took: 10634.52 ms.`
- syntax error build: 
	 ```
	Compilation took: 3098.1 ms.
	ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts
	./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& (C:/src/Project/node_modules/ts-loader??ref--2-0!C:/src/Project/node_modules/vue-loader/lib??vue-loader-options!./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts&)
	[tsl] ERROR in C:\src\Project\scripts\components\Admin\AdminManager.vue.ts(87,29)
	      TS1005: ',' expected.
	 @ ./scripts/components/Admin/AdminManager.vue?vue&type=script&lang=ts& 1:0-193 1:209-212 1:214-404 1:214-404
	 @ ./scripts/components/Admin/AdminManager.vue
	 @ ./scripts/app.ts
	 @ multi ./scripts/app.ts
	 ```
- lint error build: `Compilation took: 88.46 ms.` (**No error reported!**)
- lint error build (changed something else unrelated): 
	```
	Compilation took: 15772.66 ms.
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue.__fake__.ts
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue.__fake__.ts
	95:1 Consecutive blank lines are forbidden
	```
- type error build:
	```
	Compilation took: 11127.61 ms.
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	ERROR in C:/src/Project/scripts/components/Admin/AdminManager.vue
	95:13 Type '0' is not assignable to type 'string'.
	    93 |     // tslint:enable:object-literal-sort-keys
	    94 | 
	  > 95 |     private shouldBeString: string = 0;
	       |             ^
	    96 | 
	    97 |     private created() {
	    98 |         this.loadAll({from: 0, count: 20});
	```

phryneas:
@WolfspiritM phew, these results are a bummer :disappointed: 
Thank you for testing though, it's a great help! :+1: 
 One thing I would be interested in: could you give the tests a second round with `async: true`? There seems to be weird behaviour with different values of `async` sometimes, so I'm interested if that makes a difference.

WolfspiritM:
Sorry for the late response. While trying to figure out https://github.com/TypeStrong/ts-loader/issues/988 I gave this a try again as it's more complicated to fix ts-loader and maybe ts-loader can use the mocked FS aswell as I was playing around with a similar way to get ts-loader working correctly by replacing ts.sys. But why reimplementing the wheel as you've already done that here.

I think I figured out why it is so slow.

It seems like the initial compile is fast (10 seconds compared to more then 20-30 before) but it's the linter that is very slow. The reason seems to be the call to getProgram() as the program seems to be always invalidated and recreated for every single file. The reason why that happens is that typescript seems to invalidate the program as soon as the host does custom module resolution.

https://github.com/microsoft/TypeScript/blob/01e1b1bb276f48eaa0612a34dd8f1eec5d86bd49/src/compiler/watch.ts#L805

I then removed the "resolveModuleNames" function from the CompilerHost as it didn't seem to do anything special here and everything was done within 10 seconds!

On watch it did first recompile with about 1 second and afterwards it even compiled within 300ms! Awesome!

The project I was testing with usually took around 30 Seconds to compile and 5-10 seconds to recompile.

Here are the changes I made: https://github.com/WolfspiritM/fork-ts-checker-webpack-plugin/commit/bd913cdd1bb6ea02b7da5832c246edf16ce121f0

EDIT: Tested type checks and linter checks and both seem to work fine for normal mode aswell as watch mode

phryneas:
This is awesome, thank you for taking another look! And for those awesome findings! 
I'm currently on vacation but I'll spend some time with this once I'm back home. 
-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

zackschuster:
@johnnyreilly @phryneas fyi i haven't done any testing for vue and i don't plan on using it for future projects, so i won't be following through on testing it. my apologies!

WolfspiritM:
@phryneas Anything new?

phryneas:
@WolfspiritM I'd love to get to this, but seeing that I'd have to essentially re-do the whole PR for a third or fourth time on the current master (this is too far away for a simple rebase), I just haven't found the time for it yet.
I hope to maybe get to it in the holidays after new year's, but that's the earliest I'm seeing for this.
If you want to take a go with it before that, feel free to do so though :)

WolfspiritM:
I've redone the changes on master and pushed everything to a new branch:
https://github.com/WolfspiritM/fork-ts-checker-webpack-plugin/tree/vue-fake-filesystem

I have some problems to get the tests working. Even on master the vue tests take ages to complete for me. And then fail with:
```Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.```
Not sure yet why that is the case. Same happens on my branch.

Also I had to remove the precommit hooks (locally).
Husky doesn't let me commit anything otherwise.
Is it possible that master is having eslint issues?
On a freshly pulled master:
```
yarn run v1.19.2
$ tslint --project src/tsconfig.json && eslint ./test

ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/FilesRegister.ts:5:27 - Module 'eslint' is not listed as dependency in package.json
ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/NormalizedMessageFactories.ts:6:25 - Module 'eslint' is not listed as dependency in package.json
ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:244:5 - Multiple variable declarations in the same statement are forbidden
ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:276:5 - Multiple variable declarations in the same statement are forbidden
ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:305:31 - Module 'eslint' is not listed as dependency in package.json

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I've merged the settings with the patchTypescript config.
With that branch everything rebuilds within 500ms with one worker...with the current master it takes about >=10 Seconds with multiple workers.

@phryneas Can you maybe help me with the tests?

**Comment by phryneas** [comment]
> I've redone the changes on master and pushed everything to a new branch:
> https://github.com/WolfspiritM/fork-ts-checker-webpack-plugin/tree/vue-fake-filesystem
> 
> I have some problems to get the tests working. Even on master the vue tests take ages to complete for me. And then fail with:
> `Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 60000ms timeout specified by jest.setTimeout.`
> Not sure yet why that is the case. Same happens on my branch.

Hm. Back then, the vue integration tests were running at 150-200 seconds. Since then, they almost doubled with #355, and they take about 350 seconds on my system with your branch. So while that *is* a lot of time, it reflects what I would expect, considering those integration tests spawn a webpack instance for almost every test.

I have no timeouts though. Maybe that's a Windows thing?

I would go through those tests and run them test by test using `it.only` instead of just `it`.

> 
> Also I had to remove the precommit hooks (locally).
> Husky doesn't let me commit anything otherwise.

You can always commit by force with `git commit --no-verify`.

> Is it possible that master is having eslint issues?
> On a freshly pulled master:
> 
> ```
> yarn run v1.19.2
> $ tslint --project src/tsconfig.json && eslint ./test
> 
> ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/FilesRegister.ts:5:27 - Module 'eslint' is not listed as dependency in package.json
> ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/NormalizedMessageFactories.ts:6:25 - Module 'eslint' is not listed as dependency in package.json
> ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:244:5 - Multiple variable declarations in the same statement are forbidden
> ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:276:5 - Multiple variable declarations in the same statement are forbidden
> ERROR: C:/Work/fork-ts-checker-webpack-plugin2/src/index.ts:305:31 - Module 'eslint' is not listed as dependency in package.json
> 
> error Command failed with exit code 2.
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> ```
> 
I can confirm that - the affected lines have last been touched in #305. @johnnyreilly, do you remember anything related from that PR?

> I've merged the settings with the patchTypescript config.
> With that branch everything rebuilds within 500ms with one worker...with the current master it takes about >=10 Seconds with multiple workers.
> 
> @phryneas Can you maybe help me with the tests?

**Comment by WolfspiritM** [comment]
> I have no timeouts though. Maybe that's a Windows thing?

Good guess! It was a windows thing. Interestingly the tests don't seem to run on windows as webpack is having issues. adding some console.logs allowed me to catch many errors like that:
```
Module parse failed: Unexpected token (2:0)
    File was processed with these loaders:
     * ../../../../../node_modules/vue-loader/lib/index.js
    You may need an additional loader to handle the result of these loaders.
    |
    > <div class="hello">
    |   <h1>Example</h1>
    | </div>
```

However...that's another issue. I am now using WSL to test and it works fine. That way I was able to figure out that all that was left was adding corresponding `__fake__.ts` to the tsconfig includes. I pushed that to my branch and on my end all tests seem to pass now.

I'm wondering if it's okay to have the user specify the suffix in tsconfig (and add it to the readme) or if we need to figure out a way to rewrite the tsconfig aswell (including all "extend"s)

johnnyreilly:
> @johnnyreilly, do you remember anything related from that PR?

Could you be more specific? I'm not sure what the question is 😄

Side note: I mean to do this sometime soon: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/356

phryneas:
> > @johnnyreilly, do you remember anything related from that PR?
> 
> Could you be more specific? I'm not sure what the question is smile
> 
> Side note: I mean to do this sometime soon: #356

These eslint warnings all seemed to be "last touched" by that PR. But, also,  taking a second look, currently eslint *should* only execute on the `test' folder, but those errors are inside `src` o_O

Anyways, I guess they'll be gone after #356, so lets ignore them for now.

phryneas:
> I'm wondering if it's okay to have the user specify the suffix in tsconfig (and add it to the readme) or if we need to figure out a way to rewrite the tsconfig aswell (including all "extend"s)

If you can get that to work reliably, I think that'd be cool!

piotr-oles:
I will close it as I was able to release [v5.0.0-alpha.1](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/tree/v5.0.0-alpha.1) version :)

Issue #230, opened 2019-03-14
Emitting .d.ts files while type checking 


rogerlucena:
In our project we are working with TypeScript, Webpack and Babel 7. 

To deal with the TypeScript files we were using the awesome-typescript-loader but there were happening a lot of problems when the generated .js files were taken by Babel. Then, we decided not to use the awesome-typescript-loader anymore, but instead stick to Babel and, since Babel does not do type checking for TypeScript, we decided to do it by using your plug-in fork-ts-checker-webpack-plugin, and we are enjoying a lot the experience. 

The only problem now is that, for our application, we need to generate the .d.ts files so that the imports from one Project to another will be done properly.

We observed that your project is missing only the "emit" step for fixing the problem above.

We want to make a contribution to your project by following the steps bellow and your opinion would be very welcome.

In service.ts, inside the "run" function we will do the "getDiagnostics" for the type checking as it is done now but we will also "emit" to generate the d.ts files.

In order to not affect the other users the "emit" phase should be optional. Thus we are creating a new field in the plug-ins Options interface called "ignoreEmit" with default value of "true".

johnnyreilly:
I'm really glad you're enjoying using the plugin!

I'm afraid I have got very mixed feelings about what you're suggesting though.  The fork-ts-checker-webpack-plugin has always been intended to be a checker - not something that generates TS code in any way. 

Classically code generation has been the responsibility of ts-loader which is typically what people use alongside the fork-ts-checker-webpack-plugin. (There are examples of how you can use them together: https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin - which reminds me; I must update them now fork-ts-checker-webpack-plugin has 1 point oh-ed)

I see you're using Babel - that's cool. (It's only relatively recently that this has become possible - super cool!) 

If you wanted to switch to use ts-loader and try and add the functionality there then I'd be happy to help make that happen. The switch should be relatively simple. I think there's a need for what you're trying to do.

Alternatively if you wanted to roll your own plugin which plays well alongside the fork-ts-checker-webpack-plugin that's a great idea too. There's already examples of this such as the https://github.com/dividab/tsconfig-paths-webpack-plugin

What do you think?

piotr-oles:
I'm closing this issue as there no further comments :)

PR #237, opened 2019-03-29
Making the emission of files (like .d.ts files) possible and optional while type checking

rogerlucena:
Now that it is possible, this optional emission of files can be controlled by a flag called **"enableEmitFiles"**. This flag can be used or not by the user, and has default value set to false as a way of not affecting the ones who want to keep using the plugin for type checking only. This way, these users will not even feel any difference regarding the use of the plugin.

The idea is to add this new feature to your code as a way of solving our problem and also the problem of others in a similar situation, enriching the plugin and also remembering to do not affect the experience of any other users who want to keep using it just like before. 

Correspondent issue on GitHub: #230  

### Technical description (flow of changes step by step):

Inside the file "index.ts" we added a boolean called "enableEmitFiles" to the interface "Options". with default value set to false. This flag control the emission of files, respecting of course the configurations set to the plugin by the user (like "emitDeclarationOnly: true" inside the "tsconfig.ts" file of our app for our use case here, for example). The value of this flag is received from the user through the constructor of "ForkTsCheckerWebpackPlugin", inside the object "options" of type "Partial<Options>".

Then, the value of this flag is set to an environmental boolean variable called "ENABLE_EMIT_FILES" there and so transferred to "service.ts". There, while creating the checker, it is used in its constructor. In the case of a checker of type "IncrementalChecker" this flag is used directly by the private method "emitFiles()" there to make or not a call to "program.emit()". But in the case of the checker being of type "ApiIncrementalChecker" it had to be transferred to the "CompilerHost" through its constructor and used inside its method "writeFile()" there to indeed control the emission of files, fixing possible unwanted emissions that were happening before when using this flag only inside the "emitFiles()" method of "ApiIncrementalChecker".


.gitignore
```diff
@@ -3,7 +3,7 @@ logs
 *.log
 npm-debug.log*
 
-lib
+#lib
```
piotr-oles: why did you comment out the lib directory? :)

src/ApiIncrementalChecker.ts
```diff
@@ -41,7 +41,8 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
     private context: string,
     private linterConfigFile: string | boolean,
     private linterAutoFix: boolean,
-    checkSyntacticErrors: boolean
+    checkSyntacticErrors: boolean,
+    canEmit: boolean = false
```
piotr-oles: The name `canEmit` is not an intuitive name for me :) What do you think about `emitFiles`?

src/IncrementalChecker.ts
```diff
@@ -359,4 +363,15 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       lints.map(this.createNormalizedMessageFromRuleFailure)
     );
   }
+
+  private emitFiles() {
+    const program = this.program;
```
piotr-oles: please use `this.program` instead of assigning it to `const program` - one variable less to track :)

src/ApiIncrementalChecker.ts
```diff
@@ -160,4 +164,21 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
...
+    try {
+      const program = this.tsIncrementalCompiler.getProgram();
+
+      if (!program) {
+        throw new Error(
+          'Program (a ts.Program object) in emitFiles() for ApiIncrementalChecker.ts not found'
+        );
+      }
+
+      program.emit(); // called here but the emission of files is in fact controlled by the boolean canEmit inside writeFile of CompilerHost
```
piotr-oles: I don't see any new tests in this PR - please add tests for emitting files :)

piotr-oles (changes requested):

johnnyreilly:
As per my comments here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/230#issuecomment-473172867

I have got very mixed feelings about this PR. The fork-ts-checker-webpack-plugin has always been intended to be a checker - not something that generates TS code in any way.

For me it makes more sense for this functionality to be in a separate plugin which is used alongside the fork-ts-checker-webpack-plugin. There's already examples of this such as the https://github.com/dividab/tsconfig-paths-webpack-plugin

All that said, well done for having a crack at putting something together. Let's see where it goes. 

I don't fully understand the use case of this functionality. Could you give me an example of where you need to generate `.d.ts` files? Who is the consumer of these `.d.ts` files? If they're generated from your own code then presumably it's not you as you could instead `import` the actual `.ts` files themselves?

It's clear I'm missing something here, so if you could expand on the use case in a simple fashion I'd greatly appreciate it!

johnnyreilly:
@rogerlucena could you respond to my questions when you get a moment please?

rogerlucena:
Hi @johnnyreilly, of course!  

I am sincerely sorry for the delay, today was an incredibly busy day and I wanted to do things in order before having the time to write well the answers for you both. 😅

Let me detail the flow of challenges we have been facing here for you to understand the context better.

We are using Babel 7 and Webpack for our project. Given this we started using the awesome-typescript-loader to deal with the .ts files, but problems followed after the files were passed through the loader and then directed to Babel. To illustrate, using the babel-plugin-react-intl after awesome-typescript-loader makes the translation map not to be emitted, for example.

In this context, we decided to rethink our solution. 

For the type checking step we discovered your plugin and we are having a pretty great experience so far. 

But, here we also deal with a number of modules that have some external Webpack dependencies between each other. An optimized and organized way to deal with it is using the .d.ts files. That is why it is important for us to generate them. 

Since your plugin **has been working very well and already runs in a separate process**, it would be perfect if it could be configurable to generate these files too. We studied a little your codebase and realized that with some simple additions we could add this functionality to it. That is the reason for this pull request.

I understand very well when you say that the fork-ts-checker-webpack-plugin has always been intended to be only a checker. That is why I worried to make the emission of files completely optional and also with default value to false, as a solution that does not affect in any way the experience of the ones who are already using and want to keep using it only as a checker. The emission of files being then an additional feature that could be turned on (or off, as default) whenever necessary.

I sincerely believe that this addition is only going to enrich your plugin, solving the problem of other developers in a similar situation as ours and adding a new useful feature to the plugin at the same time. I hope you consider it, and please feel free to reach out in the case of any other questions. 😃

.gitignore
rogerlucena: Hi Piotr! Thank you for your feedback!
It is because in my application here I am using "yarn add" directly from the github to another repository. But you are right, for the pull request I must uncomment it, I have just done so. :)

src/ApiIncrementalChecker.ts
rogerlucena: I have thought about that when naming the flag too. The idea was that the boolean will _make it possible_ to emit the files, that is why I thought about "can". 

Also, I am already using "emitFiles" for the name of a method inside "IncrementalChecker" and "ApiIncrementalChecker" too. But you are right, "canEmit" may not be the best name indeed. I changed it to "enableEmitFiles" instead, I believe it is more clear now. :)

src/IncrementalChecker.ts
rogerlucena: Done, thank you again Piotr. :)

src/ApiIncrementalChecker.ts
rogerlucena: I am working on this, would you have any more suggestions on tips on how you would like it to be done? :)

rogerlucena: I am working on this, would you have any more suggestions or tips on how you would like it to be done @piotr-oles ? :)

johnnyreilly:
Thanks for responding @rogerlucena!

> But, here we also deal with a number of modules that have some external Webpack dependencies between each other. An optimized and organized way to deal with it is using the .d.ts files. That is why it is important for us to generate them.

Could you provide some more details on what this means / looks like please? As I understand it this is precisely the use case that you're hoping to satisfy. I'd really like to understand what that use case is; at the moment I'm afraid I don't quite follow at present.

Assuming it's a common use case as you suggest then we'd certainly consider merging this PR. But we need to understand the problem it seeks to solve first!

rogerlucena:
Hi @johnnyreilly !

Of course, with pleasure, no problems. 😃 

The general problem solved here is just to make possible (and optional) the emission of files from your plugin while type checking, for all the possible applications it can have. Of course, this emission will comply with the details inside the "compilerOptions" (of type "ts.CompilerOptions") in the "tsconfig.ts" file. 

**For our use case here in particular**, for example, we have the flag "emitDeclarationOnly" set to true inside this "compilerOptions" since we are interested only in the .d.ts files, as I told you before. 

These .d.ts files are important for us because we have modules with some External Webpack dependencies between each other (https://webpack.js.org/configuration/externals/), and the optimized and more organized way to do these "imports" is using the .d.ts files, as I told you before too. (From now on I will be referring to these modules as "Apps"). 

Also, on our context here, the idea is of creating an "App store" where any developer can publish the App she/he has developed. Since this App can have dependencies on other Apps, and sometimes the source code of that other App is a private intellectual property, we face a problem that could be solved if we transferred from one App to another only the "interface" of the code, namely, the .d.ts files.

Now, specifically about our product here, we are building a serverless development environment where web programmers can develop their e-commerce store websites pretty easily, fast, without dealing with infrastructure and scalability concerns - just working with Apps and an "App store". To illustrate, if they want the "Visa checkout" feature on their online store, for example, they have just to add the "Visa checkout" App to it. And so it goes.

If you want to know more about it have a look at https://bit.ly/2VkvDpH (you can change the language for English on the top right corner of the website) or https://youtu.be/iCHTRS9Aa6g .

Coming back to your plugin, I emphasize again that this emission of files will be completely optional, having default value to false (as you can see in my code and as I have already tested a number of times here). It will be just an additional feature that, in my view, will only enrich you work, not affecting in any way the experience of the ones who want to keep using it strictly as a checker.

johnnyreilly:
Thanks for the context @rogerlucena - that's helpful. 

I have another question: doesn't ts-loader support this workflow? ts-loader can already be used alongside fork-ts-checker-webpack-plugin in `transpileOnly` mode so you get the desired performance that type checking on another process affords you. See: https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin - I suspect this setup already gives you what you need if you set the relevant `tsconfig.json` properties. You may also need: https://github.com/TypeStrong/ts-loader/blob/master/README.md#declarations-dts 

In which case are you maybe rebuilding something that already exists? Or have I misunderstood?

It's possible that what you want to achieve is already available. Or if not then that's useful information to know; i.e this PR will meet a need that is not already met and is therefore a good thing™️ 😄

src/ApiIncrementalChecker.ts
```diff
@@ -160,4 +164,21 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
...
+
+      if (!program) {
+        throw new Error(
+          'Program (a ts.Program object) in emitFiles() for ApiIncrementalChecker.ts not found'
+        );
+      }
+
+      program.emit(); // called here but the emission of files is in fact controlled by the boolean enableEmitFiles inside writeFile of CompilerHost
+    } catch (error) {
+      console.log('Error inside emitFiles() of ApiIncrementalChecker');
```
johnnyreilly, line 185: these should possibly use the `this.logger` rather than `console`?

src/IncrementalChecker.ts
```diff
@@ -359,4 +363,14 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       lints.map(this.createNormalizedMessageFromRuleFailure)
     );
   }
+
+  private emitFiles() {
+    if (!this.program) {
+      throw new Error('Invoked called before program initialized');
+    }
+
+    if (this.enableEmitFiles) {
```
johnnyreilly: generally I'd like to use `if (enableEmitFiles)` inside the main body of the checker; it's slightly confusing at present in that `this.emitFiles();` suggests that files are *always* emitted

src/ApiIncrementalChecker.ts
```diff
@@ -108,6 +110,8 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
     this.lastUpdatedFiles = diagnostics.updatedFiles;
     this.lastRemovedFiles = diagnostics.removedFiles;
 
+    this.emitFiles();
```
johnnyreilly, line 118: We don't seem to be testing with `if (this.enableEmitFiles) {` here?

src/ApiIncrementalChecker.ts
johnnyreilly: Given the nature of the change I think that integration tests are what we'd be interested in.  Ideally add them to the existing integration test pack.  

For an idea of a test that actually makes changes to the filesystem (as your change would seek to do) you might want to look at the tslint autofix test: https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/06f565d2b803c6fa2070a59b80fc2ce2660ce30d/test/integration/index.spec.js#L107

src/ApiIncrementalChecker.ts
rogerlucena: Hi @johnnyreilly . You are probably right, but there is no "this.logger" inside "ApiIncrementalChecker.ts". How do you suggest to transfer this object from "index.ts" to "service.ts" (that, in his turn, will be transferring it to "ApiIncrementalChecker.ts" through its constructor)?

src/ApiIncrementalChecker.ts
rogerlucena: Thank you John, I will have a look on this.

src/IncrementalChecker.ts
rogerlucena: You are right. I started doing this check inside the main body of the checker but a moment later I transferred it inside `this.emitFiles()` to test something and let it there. I have just taken it back to the main body, thank you for your comment John.

src/ApiIncrementalChecker.ts
rogerlucena: For `ApiIncrementalChecker` I tried to use this flag here but it does not seem to work - it keeps emitting the files even when the flag is `false` as I explained on the Technical Description of the Pull Request above. I do not know exactly why because I do not have the global understanding of the way your plugin was written, do you have any clue on this?

I verified that, for `ApiIncrementalChecker`, if I do this check with the flag `this.enableEmitFiles` inside the method `writeFile()` of `CompilerHost` it does work, that is why it is done this way at the moment.

rogerlucena:
> Thanks for the context @rogerlucena - that's helpful.
> 
> I have another question: doesn't ts-loader support this workflow? ts-loader can already be used alongside fork-ts-checker-webpack-plugin in `transpileOnly` mode so you get the desired performance that type checking on another process affords you. See: https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin - I suspect this setup already gives you what you need if you set the relevant `tsconfig.json` properties. You may also need: https://github.com/TypeStrong/ts-loader/blob/master/README.md#declarations-dts
> 
> In which case are you maybe rebuilding something that already exists? Or have I misunderstood?
> 
> It's possible that what you want to achieve is already available. Or if not then that's useful information to know; i.e this PR will meet a need that is not already met and is therefore a good thing™️ 😄

For our application here, answering the reply above, we had to use `babel-loader` for the `.ts files` as a way of fixing some issues, such as to deal with the `babel-plugin-react-intl` as I told you about before, for example. That is why, for type checking, we were looking for a **_plugin_**, not a loader, and then we found and started using yours.

So, you are suggesting @johnnyreilly for us to use two loaders in sequence for the `.ts files`, is it? After the `.ts files` were passed to the `ts-loader` for type checking and generation of the `.d.ts files` we would direct them to be treated by the `babel-loader` for transpilation, is it the workflow? 

How could we chain two loaders like this - we just add a new `use: {  ` with another loader in the `index.ts` of `webpack` for the test `test: /\.(tsx?)$/`? 

Thank you in advance once again for your attention.

johnnyreilly:
> So, you are suggesting @johnnyreilly for us to use two loaders in sequence for the .ts files, is it? After the .ts files were passed to the ts-loader for type checking and generation of the .d.ts files we would direct them to be treated by the babel-loader, is it the workflow?

Yes - that's a workflow I've used in the past. It's just a matter of chaining loaders (which is a standard use case for loaders)

Here's an example of that sort of setup: https://github.com/TypeStrong/ts-loader/blob/master/examples/react-babel-karma-gulp/webpack.config.base.js

The deps are out of date but you should get the general idea

piotr-oles:
Hi @rogerlucena !
Did you manage to setup webpack using @johnnyreilly's solution?

rogerlucena:
Hi @piotr-oles !

Unfortunately not, we tried @johnnyreilly's suggestion here two weeks ago but it did not work out. As soon as I have a moment I will come back here to share precisely with you what happened, and then we can discuss again how to proceed. Thank you for your inquiry Piotr. 🙂

17biubiu:
@rogerlucena  Do you solve it? i met the same question that generate .d.ts file

piotr-oles:
I'm closing it as there is no activity :) Feel free to create new PR in the future :)

PR #246, opened 2019-04-14
use worker-rpc library for inter-process communication

phryneas:
Over in #231, you suggested

> I wonder if there's a way to tweak the checker to expose the checker purely for the purposes of testing... That could be potentially useful - it wouldn't need to be something we exposed always; we could have a "test" mode or something... Ideas!

To do that I need a way to do some inter-process communication that does not interfere with the given inter-process communication.

This PR adds worker-rpc so we can send different types of messages between the processes. Also, as it has a nice interface around promises, it acutally makes some existing code more readable.

As it did not want this to be swept under the rug by the myriad of other commits in #231, I created a separate PR for this so you can review it (I will continue using this in #231 as well, though)


src/cluster.ts
```diff
@@ -1,10 +1,16 @@
...
 import * as process from 'process';
+import { RpcProvider } from 'worker-rpc';
 
-import { WorkResult } from './WorkResult';
 import { NormalizedMessage } from './NormalizedMessage';
 import { Message } from './Message';
+import { RunPayload, RunResult, RUN } from './RpcTypes';
+
+// helper function
+function noneUndefined<T>(results: (T | undefined)[]): results is T[] {
```
piotr-oles: I think this function is more like `noneFalsy` - it will return false also for array that contains null, 0, false or ''.
I'm not sure if it makes code more readable - it's used only in one place and it's 50 lines below.

src/cluster.ts
```diff
@@ -20,54 +26,58 @@ for (let num = 0; num < division; num++) {
...
-    // if we have result from all workers, send merged
-    if (result.hasAll()) {
-      const merged: Message = result.reduce(
-        (innerMerged: Message, innerResult: Message) => ({
-          diagnostics: innerMerged.diagnostics.concat(innerResult.diagnostics),
-          lints: innerMerged.lints.concat(innerResult.lints)
-        }),
-        { diagnostics: [], lints: [] }
-      );
+  if (!noneUndefined(workerResults)) {
```
piotr-oles: Is it the case when some of child process has been killed/cancelled and there is no result? If yes, I would suggest to name this condition: `const workerFinished = workerResults.every(workerResult => !!workerResult);` and then `if (!workerFinished) { return undefined; }`

src/service.ts
```diff
@@ -61,28 +73,25 @@ async function run(cancellationToken: CancellationToken) {
...
+    diagnostics,
+    lints
+  };
 }
 
-process.on('message', message => {
-  run(CancellationToken.createFromJSON(typescript, message));
-});
+rpc.registerRpcHandler<RunPayload, RunResult>(RUN, message =>
+  run(CancellationToken.createFromJSON(typescript, message!))
```
piotr-oles: Shouldn't we check if message is not undefined?

piotr-oles (changes requested):
Thank you for your work! I like the RPC approach - it's a good direction to clean-up the codebase :)

src/cluster.ts
phryneas: I had that here, too, but without the function with a `(results: (T | undefined)[]): results is T[] ` signature, the `undefined[] | Message[]` will not be inferred to be `Message[]` in the following code - that's why I created that method in the first place.

src/cluster.ts
phryneas: I can rename it to something more fitting to what it really does, but essentially I use this just as a typecast here. I can also move it down.

src/service.ts
phryneas: I think that's just a problem with the typings of worker-rpc - the lib is written for typescript 2.0 - and back then you could not correctly type that a method argument was there, or not - depending on generic.
This should in reality never be undefined.

But honestly, the check could not hurt, so I'll add it ;)

src/cluster.ts
piotr-oles: Ok - so I would suggest to just make this function less generic to improve readability:
```
function workerFinished(workerResult: (Message | undefined)[]): workerResult is Message[]
```

phryneas:
Okay, everything should be addressed - could you take a second look? :)

johnnyreilly:
Looks really interesting! 

My one reservation is that `worker-rpc` doesn't seem to be an actively maintained project and no-one is actively using it. (40 downloads a week on npm) This makes me hesitant to create a dependency on it.

However, it turns out `worker-rpc` is a pretty simple codebase; it pretty much amounts to this file: https://github.com/DirtyHairy/worker-rpc?files=1

I wonder, given it's already written in TypeScript and released under MIT if we couldn't just use that module in the `fork-ts-checker-webpack-plugin` codebase directly? I'd be happy to have a comment in the codebase that reflects that. And I'd be happy (if @dirtyhairy would like it) to have them listed as a contributor.

phryneas:
Oh, I should have added that info: @DirtyHairy is actually a colleague of mine - as far as I know it's not maintained because it does what it needs to and there were no changes requested or required in the last few years.

We are actively using it in some customer projects and as far as I know he is using it heavily for web worker communication in his atari emulator ( https://github.com/6502ts/6502.ts ), so from my point of view I'd say it is quite mature.

As for the failing tests - I'll take a look at that tomorrow in the evening. Guess there were still tests using `WorkResult.ts`, even though the code did not reference it - and as I had a compiled version still sitting in the lib folder, my local tests did not fail.

piotr-oles (approved):

DirtyHairy:
What @phryneas says: both `worker-rpc` and `microevent.ts` (on which it depends) are maintained; they just haven't required much maintenance since I published them 😏 Both are heavily used in [6502.ts](https://github.com/6502ts/6502.ts) and are covered with tests. I trust them enough to use them in customer projects.

phryneas:
Okay, there was actually a test for WorkResult, which was made obsolete by this PR. Now tests are actually running.

Tests are now failing for webpack 2 and 3, I'm on that.

phryneas:
This test suite is great :) I had missed to migrate one RPC call - now everything should be good :)

johnnyreilly:
Great! Question: does https://www.npmjs.com/package/@phryneas/experimental-fork-ts-checker-webpack-plugin use worker-rpc?

phryneas:
Yes, but the version before the fix of yesterday, so the current build will not work on webpack 2/3.
I can merge it over and create a new build this evening.

johnnyreilly:
That'd be awesome - thanks so much! I'm going to switch over my build at work to use this and see how it goes.

I'm very much a believer in dogfooding 😉

johnnyreilly:
It would be great to get a release out with this in. @piotr-oles do you want to make that happen?

Given the [experimental version](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231) has issues it would be good to ensure that the `worker-rpc` tweaks themselves are stable. I'm pretty optimistic, so let's ship!

piotr-oles:
Yes, please release if you want to :)

johnnyreilly:
Is there a way to do it without raising another PR to update the `package.json`... I'm guessing not?

piotr-oles:
I will create PR with semantic-release integration to make releasing and versioning fully automatic :)

johnnyreilly:
Nice! 😊❤️ 

Should I push out v1.0.3 in the meantime or would you rather I leave that so you can test your new mechanism?

piotr-oles:
Please release v1.0.3 - I don't want to block you. I will release RC versions on beta channel to test it.

johnnyreilly:
Should be out shortly https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v1.0.3

cc @phryneas

PR #248, opened 2019-04-17
Prepare v1.0.3 release

johnnyreilly:
Release for https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.0.3
+
+* [use worker-rpc library for inter-process communication](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231)
+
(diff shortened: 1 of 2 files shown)


phryneas:
Argh, bad timing :/ 
I just wanted to ask you to hold off this for a moment. 

Context: I did some experimentation and I noticed that now, if an exception occurs in one of the worker processes, that exception now permeates towards the parent process in the form of an uncaught promise rejection.

Usually, that won't be a problem because usually a worker thread should not throw any exceptions, but I'd like to catch that & handle it gracefully.

phryneas:
I'll add a PR and some thorough information on this in a few minutes.

phryneas:
See #249

PR #250, opened 2019-04-18
Adds support for custom resolvers to CompilerHost

### LINKED ISSUE #181: Custom resolveModuleName
Now that support has landed in `ts-loader` for a [custom `resolveModuleName` option](https://github.com/TypeStrong/ts-loader/pull/862), would `fork-ts-checker-webpack-plugin` be interested in a similar option?

My own use case it to make this plugin work with PnP (which is natively supported by create-react-app, itself using fork-ts-checker-webpack-plugin to power its TypeScript part) 🙂

arcanis:
Work in progress to add support for custom resolvers to fork-ts-checker-webpack-plugin (and thus create-react-app!). Closes #181.

The plumbing within CompilerHost is good but I'm not sure what's the best way to send the custom resolvers to the worker, since functions cannot cross the boundaries of Node processes. From what I saw the configuration is usually forwarded through the environment, so I guess it would require two new environment variable that would contain the path to modules that would export the resolvers, which would then be dynamically required in the worker thread - does that look coherent?

Of note, I think I should theoretically duplicate the code for both ApiIncrementalChecker and IncrementalChecker, but given that CRA seems to use the first one I'd prefer to focus my effort on it unless there's technical reasons (like ApiIncrementalChecker being deprecated or something).

cc @johnnyreilly

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -576,20 +576,20 @@ class ForkTsCheckerWebpackPlugin {
   }
 
   private spawnService() {
-    const env: any = {
+    const env: { [key: string]: string | undefined } = {
       ...process.env,
       TYPESCRIPT_PATH: this.typescriptPath,
       TSCONFIG: this.tsconfigPath,
       COMPILER_OPTIONS: JSON.stringify(this.compilerOptions),
       TSLINT: this.tslintPath || (this.tslint ? 'true' : ''),
       CONTEXT: this.compiler.options.context,
-      TSLINTAUTOFIX: this.tslintAutoFix,
+      TSLINTAUTOFIX: String(this.tslintAutoFix),
       WATCH: this.isWatching ? this.watchPaths.join('|') : '',
-      WORK_DIVISION: Math.max(1, this.workersNumber),
-      MEMORY_LIMIT: this.memoryLimit,
-      CHECK_SYNTACTIC_ERRORS: this.checkSyntacticErrors,
-      USE_INCREMENTAL_API: this.useTypescriptIncrementalApi === true,
-      VUE: this.vue
+      WORK_DIVISION: String(Math.max(1, this.workersNumber)),
+      MEMORY_LIMIT: String(this.memoryLimit),
+      CHECK_SYNTACTIC_ERRORS: String(this.checkSyntacticErrors),
+      USE_INCREMENTAL_API: String(this.useTypescriptIncrementalApi === true),
+      VUE: String(this.vue)
     };
 
     if (typeof this.resolveModuleNameModule !== 'undefined') {
--- a/src/index.ts
+++ b/src/index.ts
@@ -53,6 +53,8 @@ interface Options {
   vue: boolean;
   useTypescriptIncrementalApi: boolean;
   measureCompilationTime: boolean;
+  resolveModuleNameModule: string;
+  resolveTypeReferenceDirectiveModule: string;
 }
 
 /**
@@ -99,6 +101,8 @@ class ForkTsCheckerWebpackPlugin {
   private colors: Chalk;
   private formatter: Formatter;
   private useTypescriptIncrementalApi: boolean;
+  private resolveModuleNameModule: string | undefined;
+  private resolveTypeReferenceDirectiveModule: string | undefined;
 
   private tsconfigPath?: string;
   private tslintPath?: string;
@@ -155,6 +159,9 @@ class ForkTsCheckerWebpackPlugin {
     this.silent = options.silent === true; // default false
     this.async = options.async !== false; // default true
     this.checkSyntacticErrors = options.checkSyntacticErrors === true; // default false
+    this.resolveModuleNameModule = options.resolveModuleNameModule;
+    this.resolveTypeReferenceDirectiveModule =
+      options.resolveTypeReferenceDirectiveModule;
     this.workersNumber = options.workers || ForkTsCheckerWebpackPlugin.ONE_CPU;
     this.memoryLimit =
       options.memoryLimit || ForkTsCheckerWebpackPlugin.DEFAULT_MEMORY_LIMIT;
@@ -590,6 +597,9 @@ class ForkTsCheckerWebpackPlugin {
           MEMORY_LIMIT: this.memoryLimit,
           CHECK_SYNTACTIC_ERRORS: this.checkSyntacticErrors,
           USE_INCREMENTAL_API: this.useTypescriptIncrementalApi === true,
+          RESOLVE_MODULE_NAME: this.resolveModuleNameModule,
+          RESOLVE_TYPE_REFERENCE_DIRECTIVE: this
+            .resolveTypeReferenceDirectiveModule,
           VUE: this.vue
         },
         stdio: ['inherit', 'inherit', 'inherit', 'ipc']
--- a/src/index.ts
+++ b/src/index.ts
@@ -576,6 +576,34 @@ class ForkTsCheckerWebpackPlugin {
   }
 
   private spawnService() {
+    const env: any = {
+      ...process.env,
+      TYPESCRIPT_PATH: this.typescriptPath,
+      TSCONFIG: this.tsconfigPath,
+      COMPILER_OPTIONS: JSON.stringify(this.compilerOptions),
+      TSLINT: this.tslintPath || (this.tslint ? 'true' : ''),
+      CONTEXT: this.compiler.options.context,
+      TSLINTAUTOFIX: this.tslintAutoFix,
+      WATCH: this.isWatching ? this.watchPaths.join('|') : '',
+      WORK_DIVISION: Math.max(1, this.workersNumber),
+      MEMORY_LIMIT: this.memoryLimit,
+      CHECK_SYNTACTIC_ERRORS: this.checkSyntacticErrors,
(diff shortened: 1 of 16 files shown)


johnnyreilly:
First of all thank you for the work - I really appreciate it!

I'd love to get this functionality into the plugin.

> The plumbing within CompilerHost is good but I'm not sure what's the best way to send the custom resolvers to the worker, since functions cannot cross the boundaries of Node processes. From what I saw the configuration is usually forwarded through the environment, so I guess it would require two new environment variable that would contain the path to modules that would export the resolvers, which would then be dynamically required in the worker thread - does that look coherent?

I think you're suggesting a similar approach to what we did in ts-loader here: https://github.com/TypeStrong/ts-loader/pull/745

Is that about the same thing? And yeah - I think it's pretty much the only game in town when passing functionality from process to process in node. It doesn't feel tremendous. Maybe you're aware of any other alternatives? Always open to learning new possibilities! If not then I guess we can do no other.

> Of note, I think I should theoretically duplicate the code for both ApiIncrementalChecker and IncrementalChecker, but given that CRA seems to use the first one I'd prefer to focus my effort on it unless there's technical reasons (like ApiIncrementalChecker being deprecated or something).

Get the ApiIncrementalChecker working first I think. We will need both to work before we merge as Vue doesn't work with it *yet*. (Although @phryneas may change that https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231 😊)

arcanis:
Here we go! This diff plugs two new options into the plugin:

- `resolveModuleNameModule`
- `resolveTypeReferenceDirectiveModule`

Those options are expected to be the path to a file that will be required and whose symbols will be extracted (since we use different symbol names it makes it possible to export both in the same file).

I tested it with create-react-app by manually modifying the Webpack configuration to inject those options, and it seemed to work out of the box with the following resolver:

```js
const {resolveModuleName} = require(`ts-pnp`);

exports.resolveModuleName = (typescript, moduleName, containingFile, compilerOptions, resolutionHost) => {
  return resolveModuleName(moduleName, containingFile, compilerOptions, resolutionHost, typescript.resolveModuleName);
};

exports.resolveTypeReferenceDirective = (typescript, moduleName, containingFile, compilerOptions, resolutionHost) => {
  return resolveModuleName(moduleName, containingFile, compilerOptions, resolutionHost, typescript.resolveTypeReferenceDirective);
};
```

phryneas:
Please take also a look at #231, where I am in the progress of moving the resolveModuleName logic out ouf IncrementalChecker and ApiIncrementalChecker to one shared method in https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/231/files#diff-c91a0cb3fbe6ce4c93ee9d3d2807e353R29 

Also, I've done some experimentation with "plugging some code that is available in the forked process" in #227 and like you, I came to the conclusion that passing the name of a module/file to be included is probably the way to go.

(In the end, #231 in combination with something like #227 is essentially a blueprint to add other file extensions and compilers to `fork-ts-checker-webpack-plugin` so I'd really like your feedback there, too :) )

arcanis:
Oh nice I didn't see your work, I'll give it a look next week! A cursory glance makes it look like a large in-depth refactoring; would you say #231 is a blocker to land #250? I'm trying to get a sense of what would be the best path to ship the feature before the cra team releases their 3.0 😃

phryneas:
Might depend a bit on how you fast you need this merged - I wouldn't be against adding this and re-integrating it in #231 in parallel if you need this very fast.

But we definitely should coordinate configuration options you add with this, so we don't get any weird conflicts there.

johnnyreilly:
>  I wouldn't be against adding this and re-integrating it in #231 in parallel if you need this very fast.

I agree given that the changes in this PR are much more restricted than #231 then it will likely make sense for this to go in first.

> But we definitely should coordinate configuration options you add with this, so we don't get any weird conflicts there.

When you get a moment @phryneas could you take a look at this PR and call out if there's anything that you see as a potential problem?

johnnyreilly:
I think this looks good @arcanis. Could you provide some documentation on how this should be used to the `README.md` please? I'm guessing your comment here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/250#issuecomment-484682992 might get you partway there

Could you also get this plugged (ha!) in to the `IncrementalChecker` too.

Tests. Some kind of integration tests would be 👍👍.

arcanis:
Here you go!

phryneas:
I took a look at it - I'll have to re-implement most of it for #231, but the options look okay and with the integration test I have something to test again, so no complaints from me :grin:

phryneas:
oh, one nitpick - before this makes it into a changelog - could you change the PR title from "Adds support for custom compilers to CompilerHost " to "Adds support for custom *resolvers* to CompilerHost"? :)

johnnyreilly:
Can you catch up your branch with master please @arcanis?

arcanis:
Seems like there's a problem on Node 8 🤔

arcanis:
Should be fixed - apparently Node 8 stringifies `undefined` values in the `env` object before spawning the process (whereas Node 10 removes them). I never noticed this before 🤨

src/index.ts
```diff
@@ -569,35 +576,50 @@ class ForkTsCheckerWebpackPlugin {
   }
 
   private spawnService() {
+    const env: any = {
```
johnnyreilly: Do we need an `any` here? I'll confess to a personal peccadillo: I'm totally prejudiced against `any`. It triggers me 😁

johnnyreilly (commented):
I think we're pretty much ready to go. I'd rather we lost the `any` but honestly that's of minor importance despite my nervous twitches 😏

Would you like to update the `package.json` with a new version and `CHANGELOG.md` please? This is a new feature and I think it's non breaking?

src/index.ts
arcanis: Using a stricter type required to cast values so I tried to avoid it, but I don't mind changing that. I pushed a fix 👍

johnnyreilly:
Thanks! I was just looking at the work you did on ts-loader: https://github.com/TypeStrong/ts-loader/pull/862

You only added `resolveModuleName` there. 

This adds 2 options: `resolveModuleNameModule` and `resolveTypeReferenceDirectiveModule`

Does ts-loader need  a `resolveTypeReferenceDirective` to work with pnp as well? Or is it not necessary?

arcanis:
I think that might be useful yep. When I initially made the PR for ts-loader I wasn't aware of what `resolveTypeReferenceDirective` was used for (`/// <reference types="...">`) - I just discovered it when implementing the feature for fork-ts-checker-webpack-plugin and only because `create-react-app` uses it to automatically define the `*.svg` typings.

johnnyreilly:
Cool - I might have a crack at that myself. It'll give me an excuse to get my hands dirty with pnp. I'd like to add that and put together an example repo which illustrates how you can use ts-loader with pnp. Is it okay if I ping you with questions when I'm doing that?

Is there anything else required for this PR? If not then shall we merge and 🛳️?

arcanis:
> Is it okay if I ping you with questions when I'm doing that?

Of course! Feel free to pop onto our Discord channel and mention me there 👍

> Is there anything else required for this PR? If not then shall we merge and 🛳️?

johnnyreilly (approved):

johnnyreilly:
Will ship with https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases/tag/v1.1.0 (out as soon as Travis gets done here: https://travis-ci.org/Realytics/fork-ts-checker-webpack-plugin/builds/522420897

johnnyreilly:
There's a possibility that this PR may have introduced a regression. See: https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/258

johnnyreilly:
Hey @arcanis 

Bad news - it looks like there is a regression 😢 

The details are in #258 including simple repro repo - it looks like we don't have a test that covers this case which is a shame and definitely something I'll remedy in future.

The issue is, interestingly around `paths` resolution in TypeScript in `tsconfig.json`. I'm pretty sure you already know plenty about this given your work on PnP 😄. I've written briefly on the topic here: https://blog.johnnyreilly.com/2018/08/killing-relative-paths-with-typescript-and.html just to give some use case examples.

Is there any chance you'd be able to look into this?  If you don't have time (perfectly understandable) I might have to look at reverting this PR for now so users aren't impacted.

src/index.ts
```diff
@@ -569,35 +576,50 @@ class ForkTsCheckerWebpackPlugin {
...
+    const env: { [key: string]: string | undefined } = {
+      ...process.env,
+      TYPESCRIPT_PATH: this.typescriptPath,
+      TSCONFIG: this.tsconfigPath,
+      COMPILER_OPTIONS: JSON.stringify(this.compilerOptions),
+      TSLINT: this.tslintPath || (this.tslint ? 'true' : ''),
+      CONTEXT: this.compiler.options.context,
+      TSLINTAUTOFIX: String(this.tslintAutoFix),
+      WATCH: this.isWatching ? this.watchPaths.join('|') : '',
+      WORK_DIVISION: String(Math.max(1, this.workersNumber)),
```
johnnyreilly, line 588: I don't know if this is an issue or not but I just noticed that the options now generally use `String(....)` for values. So what we're previously `number` or `boolean` will now be `string`....

Possible issue?

Issue #257, opened 2019-04-21
Re-Evaluate supported TypeScript versions, open bugs in older versions


phryneas:
I just executed the current test suite against all supported TypeScript versions. The results aren't too promising:
https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/builds/522776393

Here's a collection of the failing tests:

## [TS 2.1](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776394)
    General:
    1) should not mark warnings as errors when ignoreLintWarnings passed as option
    2) should support custom resolution
    3) should fix linting errors with tslintAutofix flag set to true
    4) should not fix linting by default
    5) should not find syntactic errors when checkSyntacticErrors is false
    6) should find syntactic errors when checkSyntacticErrors is true
    7) should find the same errors on multi-process mode
    8) should only show errors matching paths specified in reportFiles when provided
    9) should handle errors within the IncrementalChecker gracefully as diagnostic
    Vue-Specific:
    10) should get syntactic diagnostics from Vue program
    11) should resolve src attribute but not report not found error
    12) should be able to compile *.vue with each lang"before all" hook
    13) should be able to detect errors in *.vue: "before all" hook
    14) should resolve *.vue in the same way as TypeScript counter check - should report report one generic compilation error

## [TS 2.2](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776395)
    as TS 2.1

## [TS 2.3](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776396)
    General as above
    Vue-Specific:
    10) should resolve src attribute but not report not found error

## [TS 2.4](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776397)
    as TS 2.3

## [TS 2.5](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776398)
    General
    1) should not mark warnings as errors when ignoreLintWarnings passed as option
    Vue-Specific
    2) should resolve src attribute but not report not found error

## [TS 2.6](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776399)
    General
    1) should not mark warnings as errors when ignoreLintWarnings passed as option

## [TS 2.7](https://travis-ci.org/phryneas/fork-ts-checker-webpack-plugin/jobs/522776400)
[...]

PR #268, opened 2019-04-22
improve generation of diagnostics, fix linter tests

phryneas:
(feat) improve generation of diagnostics
    
>    when the options `checkSyntacticErrors: false` and `useTypescriptIncrementalApi: true` both were active, no semantic errors were emitted as soon as the first syntactic error was encountered
>    this patches the `typescript` import to override that behavior
>    see discussion in #257

(fix) linter tests
    
>    actually use the correct `useTypescriptIncrementalApi` value for some of the linter tests that were using the default value before

(fix) linter tests that were doing nothing
    
>    fix a bug where, as `tslint` was set to true and the files being checked had no parent folder with a tslint.json, tslint never found any warnings
>    also fixes a failing test because a syntactic error was counted as linter warning

see discussion in #257

this is based off #253, so please merge that before you merge this :) (should also make the diff more readable)

### DIFF
--- a/src/patchTypescript.ts
+++ b/src/patchTypescript.ts
@@ -1,0 +1,64 @@
+// tslint:disable-next-line:no-implicit-dependencies
+import * as ts from 'typescript'; // Imported for types alone
+
+export interface TypeScriptPatchConfig {
+  /**
+   * Ususally, the compilerHost created with typescript.createWatchCompilerHost will bail out of diagnostics collection if there has been any syntactic error.
+   * (see [`emitFilesAndReportErrors`](https://github.com/Microsoft/TypeScript/blob/89386ddda7dafc63cb35560e05412487f47cc267/src/compiler/watch.ts#L141) )
+   * If this plugin is running with `checkSyntacticErrors: false`, this might lead to situations where no syntactic errors are reported within webpack
+   * (because the file causing a syntactic error might not get processed by ts-loader), but there are semantic errors that would be missed due to this behavior.
+   * This ensures that the compilerHost always assumes that there were no syntactic errors to be found and continues to check for semantic errors.
+   */
+  skipGetSyntacticDiagnostics: boolean;
+}
+
+/**
+ * While it is often possible to pass a wrapped or modified copy of `typescript` or `typescript.sys` as a function argument to override/extend some typescript-internal behavior,
+ * sometimes the typescript-internal code ignores these passed objects and directly references the internal `typescript` object reference.
+ * In these situations, the only way of consistently overriding some behavior is to directly replace methods on the `typescript` object.
+ *
+ * So beware, this method directly modifies the passed `typescript` object!
+ * @param typescript TypeScript instance to patch
+ * @param config
+ */
+export function patchTypescript(
+  typescript: typeof ts,
+  config: TypeScriptPatchConfig
+) {
+  if (config.skipGetSyntacticDiagnostics) {
+    patchSkipGetSyntacticDiagnostics(typescript);
+  }
+}
+
+/**
+ * Overrides the [`typescript.createEmitAndSemanticDiagnosticsBuilderProgram`](https://github.com/Microsoft/TypeScript/blob/89386ddda7dafc63cb35560e05412487f47cc267/src/compiler/builder.ts#L1176)
+ * method to return a `ts.Program` instance that does not emit syntactic errors,
+ * to prevent the [`typescript.createWatchCompilerHost`](https://github.com/Microsoft/TypeScript/blob/89386ddda7dafc63cb35560e05412487f47cc267/src/compiler/watch.ts#L333)
+ * method from bailing during diagnostic collection in the [`emitFilesAndReportErrors`](https://github.com/Microsoft/TypeScript/blob/89386ddda7dafc63cb35560e05412487f47cc267/src/compiler/watch.ts#L141) callback.
+ *
+ * See the description of TypeScriptPatchConfig.skipGetSyntacticDiagnostics and
+ * [this github discussion](https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/257#issuecomment-485414182)
+ * for further information on this problem & solution.
+ */
+function patchSkipGetSyntacticDiagnostics(typescript: typeof ts) {
+  const {
+    createEmitAndSemanticDiagnosticsBuilderProgram: originalCreateEmitAndSemanticDiagnosticsBuilderProgram
+  } = typescript;
+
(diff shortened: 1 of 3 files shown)


phryneas:
(just force-pushed everything because I had the wording of the commit messages wrong :grin: )

piotr-oles:
I'm afraid you are missing colon next to the `type` section (should be `fix(tests): fix linter tests that were doing nothing`) :) The history of commits is clear so I wouldn't like to squash them :)

src/patchTypescript.ts
```diff
@@ -0,0 +1,39 @@
...
+   * This ensures that the compilerHost always assumes that there were no syntactic errors to be found and continues to check for semantic errors.
+   */
+  skipGetSyntacticDiagnostics: boolean;
+}
+
+export function patchTypescript(
+  typescript: typeof ts,
+  config: TypeScriptPatchConfig
+) {
+  const origTypescript = { ...typescript };
```
piotr-oles: Should we copy entire typescript object? I think it would be more elegant to copy just methods that we will use in override :)
```
const { 
  createEmitAndSemanticDiagnosticsBuilderProgram: originalCreateEmitAndSemanticDiagnosticsBuilderProgram
} = typescript;
```

src/patchTypescript.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const patchSkipGetSyntacticDiagnostics: Pick<
+    typeof ts,
+    'createEmitAndSemanticDiagnosticsBuilderProgram'
+  > = {
+    createEmitAndSemanticDiagnosticsBuilderProgram(...args: any[]) {
+      const program = (origTypescript.createEmitAndSemanticDiagnosticsBuilderProgram as any)(
```
piotr-oles: I'm not sure if `this` inside `createEmitAndSemanticDiagnosticBuilderProgram` will be correct (as we have typescript and origTypescript). I would feel more comfortable with explicit `this` binding:
```
const program = originalCreateEmitAndSemanticDiagnosticsBuilderProgram.apply(typescript, args);
```

src/service.ts
```diff
@@ -25,6 +26,13 @@ const rpc = new RpcProvider(message => {
 process.on('message', message => rpc.dispatch(message));
 
 const typescript: typeof ts = require(process.env.TYPESCRIPT_PATH!);
+const wrapperConfig: TypeScriptPatchConfig = {
```
piotr-oles: Isn't this more like `patchConfig`?

test/integration/mocks/ApiIncrementalCheckerWithRpc.js
```diff
@@ -0,0 +1,57 @@
+const mock = require('mock-require');
+
+const origImport = require('../../../lib/ApiIncrementalChecker');
```
piotr-oles: Please change name to the `originalImport` :)

piotr-oles (changes requested):
Great work :)

src/patchTypescript.ts
phryneas: For now yes, but #231 uses this pattern for about 20 methods (the file there is currently named wrapTypescript, I'll rename that there to be the same file as this one), so we already know that going forward, this will make a lot of method copies and take up like 20-40 lines (depending on max char width) that way.
The spread on the other hand is just an object full of references to the original implementations - what isn't overridden doesn't really take extra memory, so I'd suggest to go with this from the beginning.

src/patchTypescript.ts
phryneas: From what I've seen, the typescript code doesn't really make use of the `this` keyword at all, they just have that one reference to `ts` and a lot of nested closures (which is why we can't just wrap it, but have to override some methods directly on the original object).

But yeah, apart from that, this is more future-proof if they should change that at any time in the future. I'll change it to `apply` :+1:

src/service.ts
phryneas: Argh, yes. Remnant of an older #231 approach that still lurks my mind :laughing:  Gonna change it :)

phryneas:
> I'm afraid you are missing colon next to the `type` section (should be `fix(tests): fix linter tests that were doing nothing`) :) The history of commits is clear so I wouldn't like to squash them :)

Whoops! Will do a rebase & rename :)

src/patchTypescript.ts
phryneas: PS: actually, this exact method would need to be overwritten/wrapped a second time for 231, so I guess what I said here doesn't apply there. Gonna move it into a single variable ;)

phryneas: done as requested ;)

piotr-oles (approved):

phryneas:
Okay, I'll rebase this on the current `beta` and then we're good to go.

phryneas:
@piotr-oles  @johnnyreilly  Oh, seeing that I can now merge PRs myself: what is my mode of operation here?

Wait for one of you two to approve the changes, then merge into `beta`? Wait for both of you to approve?

piotr-oles:
For me one approve is enough for fixes and small improvements :) For breaking changes and new features I think it will be better for all of us to get review from two different developers :)

johnnyreilly:
> For me one approve is enough for fixes and small improvements :) For breaking changes and new features I think it will be better for all of us to get review from two different developers :)

That's a sensible general approach I think. 👍  The flip side of that is that it's important that reviews happen in a timely fashion. So if someone has asked me for a review and I haven't come back to them in a reasonable period of time, they should feel free to keep shipping.😁🛳️

BTW feel free to merge this when you've added the comment I requested @phryneas!

src/patchTypescript.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function patchTypescript(
+  typescript: typeof ts,
+  config: TypeScriptPatchConfig
+) {
+  const {
+    createEmitAndSemanticDiagnosticsBuilderProgram: originalCreateEmitAndSemanticDiagnosticsBuilderProgram
+  } = typescript;
+
+  const patchSkipGetSyntacticDiagnostics: Pick<
```
johnnyreilly: Could you put a comment in this module that explains what the purpose of this code is and links back to the GitHub issue discussion that drove this please?

I want to be kind to future maintainers 😁

johnnyreilly (dismissed):

src/patchTypescript.ts
phryneas: that file has now almost as much documentation as code :grin:

johnnyreilly:
This is great work @phryneas!

@piotr-oles what's the process now? We raise a PR to merge from `beta` to `master`?

PR #272, opened 2019-04-28
ongoing work: evaluate jest

phryneas:
This would remove mocha/chai/sinon in favor of jest.

So far, is has not been a performance improvement. On my local machine, this is currently 10 seconds slower (150s vs 140s) than the old solution, on Travis it is about 1-2 Minutes slower.

But, currently tests are running in sequence and not parallel - because currently there are tests that are reading/writing to the same folder. So refactoring those tests and enabling parallel processing *should* make the jest tests faster.

So far, I'm leaving this here until I can do those refactorings or one of you wants to do some experimentation ;)

PS: as far as this is configured, you could just add new tests as *.ts files alongside the *.js tests.


phryneas:
So I did some work, moving the helpers to typescript and doing some refactoring and this is taking on a nice shape.

* every test suite now creates a temporary folder on start
* every createCompiler call creates a subfolder and copies all required test files over there
* after each tests suite, that complete folder is cleaned up

This will allow us to run tests in parallel in the future.

Also, all helpers now rely on each other and have similar arguments. This will make splitting up tests much more easy.

One thing: I noticed that the types in `src/index.ts` (like `Options` etc.) were not exported, so I exported them. As I am not very familiar with the `exports = ...` notation (I guess that is there for compatibility reasons?), I would love if both of you @piotr-oles @johnnyreilly  could take a look that everything is still okay there.

PS: oh lol, apparently something I did breaks the unit tests - didn't check those a second time. Gonna fix those tomorrow :)

src/index.ts
```diff
@@ -22,39 +22,44 @@ import { RunPayload, RunResult, RUN } from './RpcTypes';
 
 const checkerPluginName = 'fork-ts-checker-webpack-plugin';
 
-type Formatter = (message: NormalizedMessage, useColors: boolean) => string;
+namespace ForkTsCheckerWebpackPlugin {
```
phryneas, line 29: I hope these changes are okay? (see the whole file!)

johnnyreilly:
> As I am not very familiar with the exports = ... notation (I guess that is there for compatibility reasons?),

Yeah so TypeScript predates es modules and had it's own syntax for that. I believe `export =` is equivalent to `export default`

The PR looks pretty cool - I haven't been able to get time to look at it as this week is *insanely* busy. Will try and take a look this weekend. Sounds v exciting though!

phryneas:
> > As I am not very familiar with the exports = ... notation (I guess that is there for compatibility reasons?),
> 
> Yeah so TypeScript predates es modules and had it's own syntax for that. I believe `export =` is equivalent to `export default`

From my experiments, changing it to a default export makes it transpile to a different end result and I wanted to avoid that (not sure how many different ways this is being imported). So what I did definitely transpiles the same as before and additionally emits the exported types. Just want to be :100: sure that I didn't miss anything.

> The PR looks pretty cool - I haven't been able to get time to look at it as this week is _insanely_ busy. Will try and take a look this weekend. Sounds v exciting though!

No worries, I'm doing this out of fun and I really like your relaxed approach to open source - apply it to yourself, too ;)

phryneas:
So, performance-wise this goes two directions:

On my local system, I'm from 140 seconds for all tests down to 110 seconds.
On travis, it seems to take longer (from 3-5 minutes to 5-7 minutes) as the machines there don't really profit from multi tasking.

Apart from performance:

Writing the test helpers in TS has been a blast and VSCode even gives me type checking on the .js tests now, which already helps a great deal during the refactoring.

test/types/copy-dir.d.ts
```diff
@@ -0,0 +1,13 @@
+declare module 'copy-dir' {
+  const copydir: {
```
johnnyreilly, line 2: Are there no definitions in definitely typed for unixify / copy-dir? We could fix that! (I'm a maintainer of DT so I could get it merged pretty quick)

babel.config.js
```diff
@@ -0,0 +1,7 @@
+// babel.config.js
+module.exports = {
```
johnnyreilly, line 2: I'm curious what jest needs Babel for? I'm using jest on other projects without Babel so I'm curious what purpose it has. Is this related to the recent jest port to TypeScript? (Think I'm still using the older one)

src/CompilerHost.ts
```diff
@@ -48,7 +48,7 @@ export class CompilerHost
   private readonly tsHost: ts.WatchCompilerHostOfConfigFile<
     ts.EmitAndSemanticDiagnosticsBuilderProgram
   >;
-  private lastProcessing?: Promise<ts.Diagnostic[]>;
+  protected lastProcessing?: Promise<ts.Diagnostic[]>;
```
johnnyreilly, line 51: There's lots of `private` becoming `protected` - I'm guessing there's an overall reason for that but I don't know what it is..  could you enlighten me please? 😄

src/index.ts
johnnyreilly: I think this will possibly be a breaking change for how types are consumed. Looking at code in the PR it looks like we switch from consuming as `Options` to `ForkTsCheckerWebpackPlugin.Options`.

Do I have that right? We'd need to do a new major version if so.

johnnyreilly (commented):
I've a bunch of questions. Obviously this is a big PR because of the nature of switching test frameworks. It seems pretty nice!

Couple of general questions:

- Can we can still debug tests with jest? Is it straightforward?
- Am I right in understanding that this moves both integration / unit tests to jest?

I'm unworried about perf differences. I don't think it's a significant enough factor either way.

test/types/copy-dir.d.ts
phryneas: There aren't - actually, I've never pushed something to DT, I usually just type the part of the API I need from the Documentation.
But it could be a nice exercise to do that for once :)

babel.config.js
phryneas: Jest can either run typescript tests by directly running them with `ts-jest` (which afaik is like `ts-node`) or by transpiling them with babel using `'@babel/preset-typescript'` (afaik jest uses babel anyways to make modern syntax available with old node versions). The latter is the current official approach in their documentation.

In this case, I tried ts-jest at first, but since some tests are mocking the `typescript` import, that wasn't possible because that would break ts-jest.

src/CompilerHost.ts
phryneas: Some tests access these properties. You can still access `private` or `protected` properties directly from the tests by accessing them via `instance['propertyname' ]` instead of `instance.propertyname` from the tests, but if they are `private`, typescript emits them as `any` - if you make them `protected`, typescript emits the correct type and the tests can access that correct type then.

src/index.ts
phryneas: These types weren't exported at all before, so tI think here could not exist anything out there consuming them.

phryneas:
>     * Can we can still debug tests with jest? Is it straightforward?

You mean attach a debugger? Actually, I have never even thought about doing that with any kind of test. How  would you have done that with `mocha`? I can try to reproduce that with `jest` :)

PS: according to [this blog post](https://artsy.github.io/blog/2018/08/24/How-to-debug-jest-tests/) it's simple

>     * Am I right in understanding that this moves both integration / unit tests to jest?

Yup, I don't really think we should maintain two different test runners.

johnnyreilly:
You can attach a debugger - there's some discussion on it here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/198#issuecomment-453742029

Cool - as long as there's a way to do it and that seems covered

test/types/copy-dir.d.ts
johnnyreilly: If you wanted to raise a PR and mention me I can review it and get it merged. If you'd rather keep it in here for now that's fine. I don't want to boil the ocean 😀

phryneas: I'll add it to my TODO list :D

johnnyreilly:
Cool - I'm happy with this generally. 

Could you sync up the branch please? I figure we're then good to merge. @piotr-oles any thoughts?

test/integration/helpers/createCompiler.ts
```diff
@@ -0,0 +1,162 @@
...
+if (!beforeAll || !afterAll) {
+  throw new Error('file should not be included outside of jest!');
+}
+
+const baseTmpDir = path.resolve(__dirname, '../../tmp');
+let tmpDirParent: string;
+let lastInstantiatedPlugin: ForkTsCheckerWebpackPlugin | undefined;
+
+beforeAll(function init() {
+  if (!fs.existsSync(baseTmpDir)) fs.mkdirSync(baseTmpDir);
```
piotr-oles: Does `prettier` and `tslint` operate inside the `test` directory? I think there should be braces in this statement.

test/integration/helpers/createCompiler.ts
```diff
@@ -0,0 +1,162 @@
...
+type CreateCompilerResults = {
+  compiler: webpack.Compiler;
+  readonly compilerConfig: webpack.Configuration;
+  plugin: ForkTsCheckerWebpackPlugin;
+  contextDir: string;
+  outDir: string;
+  tmpDir: string;
+};
+
+function prepareDirectory({ context }: { context: string }) {
```
piotr-oles, line 71: Could you document functions like `prepareDirectory`? The name `prepareDirectory` doesn't tell why it prepares the directory, what is the context :)

test/integration/helpers/createCompiler.ts
```diff
@@ -0,0 +1,162 @@
...
+  const tmpDir = fs.mkdtempSync(path.join(tmpDirParent, 'test'));
+  const contextDir = path.resolve(tmpDir, context);
+  const outDir = path.resolve(tmpDir, 'out');
+  fs.mkdirSync(contextDir);
+  fs.mkdirSync(outDir);
+  copyDir.sync(path.resolve(__dirname, '..', context), contextDir);
+  return { contextDir, outDir, tmpDir };
+}
+
+function doNormalizePaths(
```
piotr-oles: Why the name is **do**NormalizePaths? Do we need `do` here? If it is because of names conflict in the function below, I think it would be better to alias `normalizePaths` variable inside the `createCompiler` function.

piotr-oles: As it gets a list of diagnostics and returns a list of diagnostics, maybe it would be better to change the name to `normalizeDiagnosticsPaths`

piotr-oles (changes requested):

test/integration/helpers/createCompiler.ts
phryneas: good catch, apparently they don't.

test/integration/helpers/createCompiler.ts
phryneas: yup, good idea :)

phryneas:
So, @piotr-oles has left me some good reviews - thank you!

As I don't really have time right now: I'll pick this up in about two weeks, fix everything @piotr-oles has mentioned and convert the other unit tests to typescript as I was only halfway done with those.

I guess then this will be ready enough :)

test/integration/helpers/createCompiler.ts
johnnyreilly: Be careful when fixing this that the tslint auto format test input doesn't get formatted. I can't remember if that file is created from text during the test or if it's copied from an actual file. If it's the latter then you may need an exclusion

phryneas: As it currently stands, I can't even exclude it: the test fixtures include a `tslint.json`, which is found by tslint and overrides all settings from the higher-level tslint.json (including the fact that this should be ignored!).

My only way of solving that would be to move all fixtures out of the tests directory to an own directory `test_fixtures` or something like that. What do you think?

johnnyreilly: That sounds like a decent approach 👍

johnnyreilly: Side note: am currently experimenting with PnP. My mission: to get ts-loader and fork-ts-checker-webpack-plugin both working using PnP and then blog about it.

Haven't finished trying stuff out yet. Current feeling is that people who *hate* webpack configuration are not going to be charmed by PnP. 😁 

I think the performance benefits may win them round though!

phryneas:
One thing I am noticing: in a lot of tests, we are setting `tsconfig: false` - according to our typings this should be `string|undefined`.

I'd go and change that to `tsconfig: undefined` where encounter it - or is there any reason for that?

johnnyreilly:
None that I'm aware of - go for it

phryneas:
Okay, I think I've addressed everything @piotr-oles had requested and I've also converted the remaining integration tests to typescript, so I think this is ready. Could you please take a look?

PS: I had accidentally added some @types as dependencies, fixed that

johnnyreilly (approved):
Looks good to me!  @piotr-oles what do you think?

piotr-oles (approved):

johnnyreilly:
Feel free to merge @phryneas!

Issue #273, opened 2019-04-29
Difficulty with webpack multi-compiler and plugin hooks


swashata:
## Feature motivation

Right now, the way the plugin's hook system and webpack multi-compiler works, it makes it quite difficult to actually get the plugin hooks. For example, let's say we have

```js
const ForkTsCheckerPlugin = require('fork-ts-checker-webpack-plugin');
const compiler = webpack([
	// array of webpack configuration with fork-ts-checker plugin
]);
const pluginHooks = ForkTsCheckerPlugin.getCompilerHooks(compiler);

// This won't work, because ForkTsCheckerPlugin didn't attach itself to
// the multi-compiler, rather the individual compilers
pluginHooks.receive.tap((d, l) => {
	console.log('done');
});
```

In such multi-compiler setup, we can not hook into fork-ts-checker hooks from the multi-compiler instance.

## Feature description

However, we can loop over `compilers` property to get the actual `compiler` where the plugin is attached.

```js
compiler.compilers.forEach(singleCompiler => {
	const pluginHooks = ForkTsCheckerPlugin.getCompilerHooks(singleCompiler);

	// This will work
	pluginHooks.receive.tap((d, l) => {
		console.log('done');
	});
});
```

## Feature implementation

So I was thinking maybe the `ForkTsCheckerPlugin.getCompilerHooks` could be a bit intuitive to support multi-compiler out of the box. Internal implementation could be like this

1. If compiler is a single instance, the don't change anything.
2. If compiler is a multi-compiler instance, then loop over all compilers instance and add the callback on the same hook for every instance.

Maybe this would be a bit out of scope, but I think and helper would be really nice here. Something like

```js
export function getMultiCompilerHooks(compiler) {
	if (! 'compilers' in compiler) {
		throw new Error('Works only with multi-compiler instance');
	}
	const hooks = [];
	compiler.compilers.forEach(singleCompiler => {
		hooks.push(getCompilerHooks(singleCompiler);
	});
	return hooks;
}
```

In any case, a little documentation on usage of the hooks would be really nice. I was trying to implement async typechecking and custom error reporting for a multi-compiler friendly webpack setup on my [tooling](https://github.com/swashata/wp-webpack-script) and I really needed to dig through the source code to findout how and why things work.

johnnyreilly:
[...]

PR #274, opened 2019-04-29
docs: 📖 add hooks documentation

swashata:
Add documentation on how to use hooks on a compiler instance.

See #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,17 +249,91 @@ We hope this will be resolved in future; the issue can be tracked [here](https:/
 
 This plugin provides some custom webpack hooks (all are sync):
 
-| Event name                              | Description                                                                    | Params                                                                     |
-| --------------------------------------- | ------------------------------------------------------------------------------ | -------------------------------------------------------------------------- |
-| `fork-ts-checker-cancel`                | Cancellation has been requested                                                | `cancellationToken`                                                        |
-| `fork-ts-checker-waiting`               | Waiting for results                                                            | `hasTsLint`                                                                |
-| `fork-ts-checker-service-before-start`  | Async plugin that can be used for delaying `fork-ts-checker-service-start`     | -                                                                          |
-| `fork-ts-checker-service-start`         | Service will be started                                                        | `tsconfigPath`, `tslintPath`, `watchPaths`, `workersNumber`, `memoryLimit` |
-| `fork-ts-checker-service-start-error`   | Cannot start service                                                           | `error`                                                                    |
-| `fork-ts-checker-service-out-of-memory` | Service is out of memory                                                       | -                                                                          |
-| `fork-ts-checker-receive`               | Plugin receives diagnostics and lints from service                             | `diagnostics`, `lints`                                                     |
-| `fork-ts-checker-emit`                  | Service will add errors and warnings to webpack compilation ('build' mode)     | `diagnostics`, `lints`, `elapsed`                                          |
-| `fork-ts-checker-done`                  | Service finished type checking and webpack finished compilation ('watch' mode) | `diagnostics`, `lints`, `elapsed`                                          |
+| Event name                              | Hook Access Key      | Description                                                                    | Params                                                                     |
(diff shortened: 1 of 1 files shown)


piotr-oles (approved):
Well done :)

phryneas:
For me, who never has used one of these hooks before, `yourPluginName` is not 100% clear. Do I need to write an own plugin to use this? From the code it seems as if it can be used just from the file that instantiates webpack.

phryneas (commented):
For me, who never has used one of these hooks before, the way you use `yourPluginName` is not 100% clear. Do I need to write an own plugin to use this? From the code it seems as if it can be used just from the file that instantiates webpack.

Could you add some more information on that part please?

swashata:
Thanks for the review. Could you please give me some suggestions?

1. `yourPluginName`, I just borrowed from webpack docs. Perhaps writing `someDescriptor` makes more sense?
2. Perhaps I am missing something here, but I thought the only way to access the hooks is to call `getCompilerHooks` with the compiler passed in as argument. If this is not the case, could you please give me an example?

johnnyreilly:
In case it helps, here's an example of hooking in: 

https://github.com/johnnyreilly/fork-ts-checker-notifier-webpack-plugin/blob/cafba0be10b8151ebf4b45d189129b85357359bf/index.ts#L109

phryneas:
> Thanks for the review. Could you please give me some suggestions?
> 
>     1. `yourPluginName`, I just borrowed from webpack docs. Perhaps writing `someDescriptor` makes more sense?

From looking at the [code](https://github.com/webpack/tapable/blob/master/lib/Hook.js), they call it 'name' internally. I guess something like `yourListenerName` or `yourSubscriptionName` could be appropriate. Having "plugin" in there and adding an example where it is invoked outside of a plugin is a bit confusing I think.

>     2. Perhaps I am missing something here, but I thought the only way to access the hooks is to call `getCompilerHooks` with the compiler passed in as argument. If this is not the case, could you please give me an example?

No, I believe your example is 100% correct. I just wanted to display the confusion that the naming `yourPluginName` could invoke in the person reading the documentation. Obviously, it confused everyone :laughing:

swashata:
Got your point 😅. I will make the changes tomorrow and update this PR

swashata:
Ready for another round of review @phryneas, @johnnyreilly, @piotr-oles 😄

phryneas (approved):
looks good to me

johnnyreilly:
Cool! Feel free to merge!

johnnyreilly:
Awesome! Do we need to do a beta -> master PR now so these changes show up, when people are reading the docs?

phryneas:
@johnnyreilly yup, see #275.

allen-munsch:
@swashata 

Do you happen to know off hand why something like this might not work:

```js
let config = {...}
let compiler = webpack(config);
new ForkTsCheckerWebpackPlugin().apply(compiler);
const pluginHooks = ForkTsCheckerWebpackPlugin.getCompilerHooks(compiler);
pluginHooks.receive.tap('some-receive', (d, l) => {
    console.log(d);
    console.log(l);
});

pluginHooks.emit.tap('some-emit', (d, l) => {
    console.log(d);
    console.log(l);
});
module.exports = config;
```

Or know off hand how I can bubble up any tslint errors from the forked process?

Right now I'm looking at using this in the plugins sections of the webpack config:

```js
    {
      apply: (compiler) => {
        compiler.hooks.done.tap('pm-on-emit', (stats) => {
          if (stats.compilation.errors.length) {
            stats.compilation.errors.forEach(err => console.error(err.message));
            process.exit(1)
          }
        });
      }
    }
```

swashata:
@allen-munsch yes. You need to get plugin hooks from individual compiler instances, not the whole compiler. Do check the readme and multicompiler section

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/README.md#accessing-hooks-on-webpack-multi-compiler-instance

PR #276, opened 2019-05-01
Semantic release improvements

piotr-oles:
To avoid issues that we had in the #275, we drop `CHANGELOG.md` generation support. We can regenerate it  in any time using existing commit messages or by copying it from the [GitHub Releases](https://github.com/Realytics/fork-ts-checker-webpack-plugin/releases) page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,4 +34,7 @@ jobs:
         script:
           - yarn exec semantic-release
         on:
-          all_branches: true
+          branches:
+            only:
+              - beta
+              - master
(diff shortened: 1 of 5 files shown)


phryneas:
Gonna miss those emoji, now that I finally figured out how to use them ^^. 

I guess we should delete the CHANGELOG before merging as it won't be consistent in the future?

Also, yes, :+1: for rebasing.

According to (this page)[https://help.github.com/en/articles/configuring-commit-rebasing-for-pull-requests] we could disable merging pull requests and only allow squash/rebase - but I guess we would still need merge for some other PRs - does anyone know if we could change the *default*  to rebase? I found no info on that.

phryneas (approved):

piotr-oles:
I've added info to the `CHANGELOG.md` that next versions will be available on the GitHub Releases page. I suggest to remove this in the future and keep it as a *deprecated* file.

PR #281, opened 2019-05-08
fix: add @types dependencies for tapable + webpack

johnnyreilly:
Fix this error when using strict package managers like pnpm/yarn pnp:

    ERROR in /path/to/project/node_modules/.registry.npmjs.org/fork-ts-checker-webpack-plugin/1.2.0/node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts(1,26):
    TS7016: Could not find a declaration file for module 'webpack'. '/path/to/project/node_modules/.registry.npmjs.org/webpack/4.30.0/node_modules/webpack/lib/webpack.js' implicitly has an 'any' type.
      Try `npm install @types/webpack` if it exists or add a new declaration (.d.ts) file containing `declare module 'webpack';`
    ERROR in /path/to/project/node_modules/.registry.npmjs.org/fork-ts-checker-webpack-plugin/1.2.0/node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts(48,209):
    TS7016: Could not find a declaration file for module 'tapable'. '/path/to/project/node_modules/.registry.npmjs.org/tapable/1.1.3/node_modules/tapable/lib/index.js' implicitly has an 'any' type.
      Try `npm install @types/tapable` if it exists or add a new declaration (.d.ts) file containing `declare module 'tapable';`
    ERROR in /path/to/project/node_modules/.registry.npmjs.org/fork-ts-checker-webpack-plugin/1.2.0/node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts(48,253):
    TS7016: Could not find a declaration file for module 'tapable'. '/path/to/project/node_modules/.registry.npmjs.org/tapable/1.1.3/node_modules/tapable/lib/index.js' implicitly has an 'any' type.
      Try `npm install @types/tapable` if it exists or add a new declaration (.d.ts) file containing `declare module 'tapable';`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,4 @@ jobs:
         script:
           - yarn exec semantic-release
         on:
-          branches:
-            only:
-              - beta
-              - master
+          all_branches: true
--- a/package.json
+++ b/package.json
@@ -98,6 +98,8 @@
     ]
   },
   "dependencies": {
+    "@types/tapable": "^1.0.4",
+    "@types/webpack": "^4.4.19",
     "babel-code-frame": "^6.22.0",
     "chalk": "^2.4.1",
     "chokidar": "^2.0.4",
@@ -115,7 +117,6 @@
     "@types/minimatch": "^3.0.1",
     "@types/node": "^8.10.38",
     "@types/semver": "^5.5.0",
-    "@types/webpack": "^4.4.19",
     "chai": "^4.2.0",
     "commitlint": "^7.5.2",
     "css-loader": "0.28.11",
--- a/yarn.lock
+++ b/yarn.lock
@@ -381,9 +381,10 @@
   version "5.5.0"
   resolved "https://registry.yarnpkg.com/@types/semver/-/semver-5.5.0.tgz#146c2a29ee7d3bae4bf2fcb274636e264c813c45"
 
-"@types/tapable@*":
+"@types/tapable@*", "@types/tapable@^1.0.4":
   version "1.0.4"
   resolved "https://registry.yarnpkg.com/@types/tapable/-/tapable-1.0.4.tgz#b4ffc7dc97b498c969b360a41eee247f82616370"
+  integrity sha512-78AdXtlhpCHT0K3EytMpn4JNxaf5tbqbLcbIRoQIHzpTIyjpxLQKRoxU55ujBXAtg3Nl2h/XWvfDa9dsMOd0pQ==
 
 "@types/uglify-js@*":
   version "3.0.4"


johnnyreilly:
Any idea why this says "This branch is out of sync with the base branch.
Changes from the base branch must be synced with this branch before merging."?

It looks like there's still something not quite right with our setup .... Any ideas @piotr-oles / @phryneas?

Are we safe to hit the "update" button or will that cause other problems?

piotr-oles:
The only problem with "Update branch" will be that it will change the commit hash. 
I can change a link in the "Realseses" manually to point to the proper commit.

There is one merge commit more on the master branch: e9ee9a from #275. I had to use a merge commit but that the last time it was used. Now merge commits are blocked and I believe that if we update this branch and merge with rebase it will work well 😀

johnnyreilly:
> Now merge commits are blocked and I believe that if we update this branch and merge with rebase it will work well 😀

We'll get there! So by the sounds of it, after this  we'll be at the point where semantic release works without extra attention? 🤞

piotr-oles (approved):

vakrilov:
This PR is causing problems with non-node projects.

`@types/webpack` has a dependency to `@types/node`. Having `@types/webpack` as a direct dependency of `fork-ts-checker-webpack-plugin` means that you will always get `@types/node` in your project (because of dependency chain).

It is OK for node project, but for non-node project you have all the node typings which might cause build time TS error because of conflicting types.

Would you consider moving the `@types/webpack` dependency as a `devDependency` - this way it won't be installed automatically.

johnnyreilly:
@arcanis - would this change suggested by @vakrilov work for yarn PnP?

arcanis:
I'm not sure it would since you depend on webpack in the [`index.d.ts` file](https://cdn.jsdelivr.net/npm/fork-ts-checker-webpack-plugin@1.3.1/lib/index.d.ts). TypeScript would emit errors if it couldn't be imported, right (even without considering PnP)?

johnnyreilly:
You're right.... Hmmmm... Not sure what to suggest here. No `@types/webpack`, no yarn PnP support. But the alternative (current) situation is potential type conflicts.

I suppose this could be worked around in `tsconfig.json` with `excludes` or perhaps `types` 

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

But that is not a great default experience... I'm not too sure what to suggest here. Any ideas gratefully received ...

arcanis:
Just to be sure - regular (non-PnP) installs would work without `@types/webpack` being available in the node_modules? How would TypeScript know that it's to be expected and not emit errors?

phryneas:
It just doesn't check these, as these only describe the .js files we're shipping and most users don't write their own code against these, so the .d.ts are never evaluated.

I guess if we would `any`-type `webpack.Compiler`, our emmited types would not depend on `webpack` at all.

Otherwise, we could just copy over the definition of `webpack.Compiler`, `webpack.Stats`, `typeof webpack.compilation.Compilation`, `tapable.AsyncSeriesHook` and `tapable.SyncHook`  (those five interfaces are all we use) into a .ts file of our own and just drop the dependencies to `@types/webpack` and `@types/tapable` completely, so that we don't accidentally add that dependency again in the future..

Or we just don't ship our own type definitions at all. (Or only a select subset, like types for the options.. although those weren't part of the types until recently). That way, there would be not dependency as well.

phryneas:
Or we could just mark the `ForkTsCheckerWebpackPlugin.getCompilerHooks` & `ForkTsCheckerWebpackPlugin.apply` properties as `@internal` and activate `compilerOptions.stripInternal` for release builds.

That way, only these two properties (which depend on `webpack`) would be removed from the built types.

Edit: not a good idea. `getCompilerHooks` is a documented API, we shouldn't make that go away in the types.

phryneas:
I tried some different things and I think for now the hotfix with minimal changes is over in #287. Please take a look if this works for everyone!

PR #282, opened 2019-05-08
chore: fix travis deployment - revert to all_branches: true

piotr-oles:
It's hard to test travis deployment on protected branches - that's why
we have to revert changes and use "all_branches: true" - specyfing more
than one branch won't work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,4 @@ jobs:
         script:
           - yarn exec semantic-release
         on:
-          branches:
-            only:
-              - beta
-              - master
+          all_branches: true


piotr-oles:
I'm sorry guys - my fault. The Travis documentation is not always clear. I checked the `.travis.yml` with a Travis lint tool but there were no errors.

On the beta branch build we've got a message:
> Skipping a deployment with the script provider because this branch is not permitted: beta

piotr-oles:
I'll merge it to move forward with #282 as this is only Travis related change.

Issue #284, opened 2019-05-09
Discussion: releasing


phryneas:
CC @piotr-oles  @johnnyreilly 

So as I noticed in #283, our current way of rebasing `beta` into `master` won't work. Github rewrites hashes even if there is no need to and this means, with every release, `master` and `beta` will be out of sync history-wise even if they contain commits with the same message etc.

Especially for future PRs, this is super-confusing.

So my idea: why don't we trigger this from the CI?

To release, we could use [a custom build trigger from travis](https://blog.travis-ci.com/2017-08-24-trigger-custom-build) - we'd trigger a build from there, setting `RELEASE=true` and then in the build job, we could have [a conditional stage](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) `release`.

That stage would essentially be a script like this:

```
git fetch origin
AHEAD=$(git rev-list --left-right --count origin/master...origin/beta | cut -f1)
[ "$AHEAD" = "0" ] || { echo "master is ahead of beta, skipping."; exit 1; }
git checkout master
git reset --hard origin/beta
git push -f origin master
```

Note that this is not my favourite way of doing this, just a practical one. I'd very much prefer a UI that is made for this specific task - maybe you could [share this tweet](https://twitter.com/phry/status/1126562273626066946) (or just ask around yourselves :) ) and maybe someone knows of a UI exactly for that purpose

johnnyreilly:
Are you saying that semantic release will never work without manual intervention?

I'm surprised if that's the case as that would kind of defeat the purpose of it to a certain extent. I mean to say: semantic release is supposed to remove friction from the release process. If I understand you correctly, given the way GitHub works, there is always going to be this added friction; literally having to manually update commits for each release. Do I understand you correctly?

I think it's super cool that you've come up with a solution. That's amazing! Well done!

But where this leaves me is here: are we using semantic release correctly? Surely it can't be the case that everyone is doing this? Can they? Why would they be willing to put up with the inconvenience?

Is our setup in some way custom? If so, can we adjust it so we're going with the default flow instead? I'm very much in favour of not deviating too much when you're using someone else's service / tool. Whenever I've done that I've just made future me angry and sad with past me. 😁

If things are as described then I don't see us getting value out of semantic release. And I'd dearly love for us to crack this! 

[...]

PR #285, opened 2019-05-09
minor release - update docs

phryneas:
Let's get this into master :)

Also, let's test what GH does with commit hashes on a rebase ;)

Edit: oh, and @piotr-oles, please you do the rebase - the current commit is already authored by me, so maybe in my case it wouldn't change the hash no matter what.


piotr-oles (approved):

piotr-oles:
You were right :(

phryneas:
Yup :( 
See #284 - I hope we can find a more elegant solution but that might be a last resort.

phryneas:
Oh, could you please work your magic again and push the new master onto beta? Before the next PR get's based of one of the duplicate commits ^^

PR #287, opened 2019-05-10
loosen type dependencies (see #281)

phryneas:
This should at least provide a hotfix for the problems in #281.

This is the new [index.d.ts](https://pastebin.com/QBTQXauF) that would be emitted by these changes. 

@vakrilov @elliottsj @arcanis  @johnnyreilly @piotr-oles - could you please take a look that this still works for everyone? Would hate to solve one problem and add a new one in the same moment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,8 +98,6 @@
     ]
   },
   "dependencies": {
-    "@types/tapable": "^1.0.4",
-    "@types/webpack": "^4.4.19",
     "babel-code-frame": "^6.22.0",
     "chalk": "^2.4.1",
     "chokidar": "^2.0.4",
@@ -117,6 +115,7 @@
     "@types/minimatch": "^3.0.1",
     "@types/node": "^8.10.38",
     "@types/semver": "^5.5.0",
+    "@types/webpack": "^4.4.19",
     "chai": "^4.2.0",
     "commitlint": "^7.5.2",
     "css-loader": "0.28.11",
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -2,7 +2,7 @@
 import * as webpack from 'webpack';
 import { AsyncSeriesHook, SyncHook } from 'tapable';
 
-type ForkTsCheckerHooks =
+export type ForkTsCheckerHooks =
   | 'serviceBeforeStart'
   | 'cancel'
   | 'serviceStartError'
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,11 @@ import { createCodeframeFormatter } from './formatter/codeframeFormatter';
 import { FsHelper } from './FsHelper';
 import { Message } from './Message';
 
-import { getForkTsCheckerWebpackPluginHooks, legacyHookMap } from './hooks';
+import {
+  getForkTsCheckerWebpackPluginHooks,
+  legacyHookMap,
+  ForkTsCheckerHooks
+} from './hooks';
 import { RunPayload, RunResult, RUN } from './RpcTypes';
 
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -117,6 +115,7 @@
     "@types/minimatch": "^3.0.1",
     "@types/node": "^8.10.38",
     "@types/semver": "^5.5.0",
+    "@types/webpack": "^4.4.19",
     "chai": "^4.2.0",
```
johnnyreilly, line 119: It looks that `@types/webpack` moves to being a `devDependency` here. But it looks like we're using `any` in the codebase. Or is that not always the case? (I'm on my phone so I may have missed something)

johnnyreilly (approved):
Looks good to me! Side note: at some point I'll write some definitely typed compilation tests to check that our exported definitions don't break in a release. 

Incidentally, I wonder how many people actually use them? I say this as a person who has been using TypeScript since it was released, but still happily writes his `webpack.config.js` files in JS 🤗

package.json
phryneas: We're using it in a few dozen other places internally. These two declarations were just the only two that were exposed externally, and thus the only ones to end up in index.d.ts upon compilation.

phryneas:
> Looks good to me! Side note: at some point I'll write some definitely typed compilation tests to check that our exported definitions don't break in a release.

:+1: 
 
> Incidentally, I wonder how many people actually use them? I say this as a person who has been using TypeScript since it was released, but still happily writes his `webpack.config.js` files in JS hugs

I have done so in the past, but not always - so I guess there are people out there ;)

vakrilov:
This looks good to me! I will be able to test it thoroughly next week, but this should solve the problem.

johnnyreilly:
It would be cool to merge and release this... @piotr-oles would you be able to help out with that? I'm afraid I don't know how to ⛴️ successfully with our current setup.

No particular rush - but it would be good to get this out there when you have a moment.

piotr-oles (commented):
Do we want to release this directly on the stable channel?

johnnyreilly:
I think it's fine yes - though if you'd rather take another tack that's fine

elliottsj:
Unfortunate that this is needed; npm's package hoisting seems to be the source of a lot of headaches 😅

With release 1.3.2, this issue has returned:

```
$ npm i typescript@3.4.5 fork-ts-checker-webpack-plugin@1.3.2
$ ./node_modules/.bin/tsc --noEmit ./node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts
node_modules/fork-ts-checker-webpack-plugin/lib/hooks.d.ts:1:26 - error TS2307: Cannot find module 'webpack'.

1 import * as webpack from 'webpack';
                           ~~~~~~~~~


Found 1 error.
```

Note that this is using the normal npm client, not pnpm or yarn PnP. I can create a new issue if that's preferred.

@vakrilov @phryneas Would you be able to provide a small example of a non-node project which conflicts with `@types/node`? I'd like to create an issue on https://github.com/microsoft/TypeScript to see if TypeScript's maintainers have any opinions about this default behaviour of typeRoots.

vakrilov:
Sure. I have added a test project [here](https://github.com/vakrilov/fork-ts-checker-example). It was generated with NativeScript CLI, but you don't need NativeScript to get the issue - it is build time.

You can reproduce in any web project too if you have code like this 
`const id: number = setTimeout(() => { ... });`

elliottsj:
@vakrilov Thanks! Submitted an issue here: https://github.com/microsoft/TypeScript/issues/31373

PR #290, opened 2019-05-15
fix: test master branch release process works-no code changes

johnnyreilly:
This PR is intended to test the new release process on the master branch - see here: https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/284#issuecomment-491811698

There's no code changes in here; just a test that the release process works when we merge a PR into master. This should trigger a release which is codewise identical to the current live release.

#### Side discussion 😄 

@piotr-oles it looks like I'm no longer able to merge a PR that I've raised.  I think it's great that we're careful about releasing something. However I'm a little concerned that this change will mean that we're unable to ship speedily when we need to.

I am mindful that there have been fairly long stretches when it's only been me working on the `fork-ts-checker-webpack-plugin`.  People come and go in open source, this is people's spare time and that's great!  But I want to guard against the situation where we won't be able to merge and ship.

I think we have a key person dependency at the moment.  As far as I know, you're our only admin.  It's generally good to share this around to plan for the future - both the expected and the unexpected!  Certainly on `ts-loader` I've made sure that, should I not be available all of a sudden, others are able to take things on.  It would be good to do the same on `fork-ts-checker-webpack-plugin`.  It's worth saying that the initial maintainer of `ts-loader` stepped away from the project a long time ago - if he hadn't taken the steps of powering up others the project would either of died or been forked.  I'd like to give `fork-ts-checker-webpack-plugin` that same insurance / redundancy.

As things stand I'm not an admin on the fork-ts-checker-webpack-plugin.  Would you be able to make me an admin so that, in the event it's only me working on the plugin I'll still be able to merge and ship?  Alternately, could we drop the 1 review minimum?  I'm super keen that the project can be depended upon.  At the moment I feel that if there were problems I would be under equipped to deal with them.  

Apologies if this comes across as me being naggy or difficult.  I care a great deal about this project and I want it to have a bright future. ❤️ 

You may have a different perspective - happy to discuss if so! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,24 @@
 # Fork TS Checker Webpack Plugin
 
 [![npm version](https://img.shields.io/npm/v/fork-ts-checker-webpack-plugin.svg)](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin)
+
 [![npm beta version](https://img.shields.io/npm/v/fork-ts-checker-webpack-plugin/beta.svg)](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin)
+
 [![build status](https://travis-ci.org/Realytics/fork-ts-checker-webpack-plugin.svg?branch=master)](https://travis-ci.org/Realytics/fork-ts-checker-webpack-plugin)
+
 [![downloads](http://img.shields.io/npm/dm/fork-ts-checker-webpack-plugin.svg)](https://npmjs.org/package/fork-ts-checker-webpack-plugin)
+
 [![commitizen friendly](https://img.shields.io/badge/commitizen-friendly-brightgreen.svg)](http://commitizen.github.io/cz-cli/)
+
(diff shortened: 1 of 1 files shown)


piotr-oles:
You're right - I do not want to block this project in any way.
That's why you've just become an admin 🎉 

> With great power comes great responsibility

piotr-oles (approved):

johnnyreilly:
Thanks @piotr-oles! I really appreciate the trust 🥰

I'll now do a simple merge and hopefully that will result in a new version being published 🤞😁

johnnyreilly:
It's done - I "squash and merge"d - let's see what happens 😁

By the way the docs in contributing.md say:

> Releasing a stable version
> Fixes and tested features from the beta branch can be merged with the master branch. Please ensure your branch is up to date with the master and then open a Pull Request.

Should this be updated?

johnnyreilly:
It worked!

PR #293, opened 2019-05-21
fix: emit global errors even when checkSyntacticErrors: false

### LINKED ISSUE #289: Build passes if some import has filename with different case
## Current behavior
Consider we have `src/SomeLib.ts` and import it as `import {something} from "src/someLib"`.
If SomeLib has type errors fork-ts-checker will give no hint about this. Even more, it completely misses all type errors in project and always builds with "no type errors found". Standard ts compiler will give errors and will warn that import filename case differs from original.

## Expected behavior
Fork-ts-checker should give same errors as ts compiler

Create index.ts and lib.ts, put content to index.ts: `import {Something} from "Lib.ts"`, put content to lib.ts
```
type Obj = {something: number};
export const Something: Obj = {something: "notnumber"};
```
build with webpack.

## Test Repository
https://github.com/Reeywhaar/fork-ts-checker-289

## Environment
 - **fork-ts-checker-webpack-plugin**: 1.3.1
 - **typescript**: 3.4.5
 - **webpack**: 4.30.0
 - **os**: MacOS 10.14.4

## Additional
<details>
<summary>tsconfig.json</summary>

```
{
  "compilerOptions": {
    "target": "es2017",
    "module": "esnext",
    "lib": ["es2017", "dom"],
    "incremental": true,
    "jsx": "preserve",
    "jsxFactory": "React.createElement",
    "strict": true,
    "esModuleInterop": true,
    "moduleResolution": "node",
    "alwaysStrict": true,
    "sourceMap": true,
    "noImplicitAny": false,
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true,
    "allowSyntheticDefaultImports": true,
    "skipLibCheck": true,
    "allowJs": true,
    "baseUrl": "./src",
    "outDir": "./build",
    "paths": {
      "@app/*": ["./*"]
    }
  },
  "include": ["./src/**/*"],
  "exclude": ["node_modules", "public"],
}
```
</details>

phryneas:
This should close #289

Previously, global errors (judging from the TS source: errors without a file name) were suppressed when only syntactic errors should have been suppressed.
Unfortunately, `ts-loader` with `transpileOnly: true` would not report those either.

This lead to this error not being reported:

```
ERROR in undefined(undefined,undefined):
TS1149: File name 'removed/DispatchBinder.ts' differs from already included file name 'removed/dispatchBinder.ts' only in casing.
```

Thanks to @Reeywhaar for providing a reproduction repo: https://github.com/Reeywhaar/fork-ts-checker-289

(this reproduction only works on case-insensitive file-systems, so in linux it has to be checked out on a VFAT mount. As that in return won't support symlinks, *node_modules*  has to be mounted as tmpfs or something similar)

### DIFF
--- a/src/CompilerHost.ts
+++ b/src/CompilerHost.ts
@@ -82,7 +82,12 @@ export class CompilerHost
       typescript.sys,
       typescript.createEmitAndSemanticDiagnosticsBuilderProgram,
       (diag: ts.Diagnostic) => {
-        if (!checkSyntacticErrors && diag.code >= 1000 && diag.code < 2000) {
+        if (
+          !checkSyntacticErrors &&
+          diag.code >= 1000 &&
+          diag.code < 2000 &&
+          diag.file // if diag.file is undefined, this is not a syntactic error, but a global error that should be emitted
+        ) {
           return;
         }
         this.gatheredDiagnostic.push(diag);
--- a/test/fixtures/caseSensitiveProject/src/index.ts
+++ b/test/fixtures/caseSensitiveProject/src/index.ts
@@ -1,0 +1,3 @@
+import y from './Lib';
+
+console.log(y);
--- a/test/fixtures/caseSensitiveProject/src/lib.ts
+++ b/test/fixtures/caseSensitiveProject/src/lib.ts
@@ -1,0 +1,1 @@
+export default 'test';
--- a/test/fixtures/caseSensitiveProject/tsconfig.json
+++ b/test/fixtures/caseSensitiveProject/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "lib": ["es2016"],
+    "forceConsistentCasingInFileNames": true
+  },
+  "import": ["./src/**/*"]
+}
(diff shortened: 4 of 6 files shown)


phryneas:
Due to this being only reproducible on a non-linux filesystem I can't really write a test for this. I imagine, we could have a test that would be skipped in a Linux environment, but it would be very fiddly for me to write that test.

phryneas:
Ah well, let's test how well that test runs :D

phryneas:
Hm, fails :/

@Reeywhaar as you're a Mac user, could you please take a look at that test? I thought I had reproduced the case-insensitivity in my Linux system, but apparently not good enough.

reeywhaar:
Dunno, if it's the case, but `"forceConsistentCasingInFileNames": true` is absent in `test/fixtures/caseSensitiveProject/tsconfig.json`, with it test passes with both `checkSyntacticErrors: false/true`

phryneas:
That makes a lot of sense, guess I just forgot that :) Let's give it a try :+1:

johnnyreilly (approved):
Great work! 🤗

piotr-oles (approved):

Issue #203, opened 2019-01-19
Support typescript-eslint when it becomes available


johnnyreilly:
TypeScript are planning to move to using eslint in future to lint TypeScript: https://github.com/Microsoft/TypeScript/issues/29288

See also: https://eslint.org/blog/2019/01/future-typescript-eslint

Given this plugin supports running tslint side by side with the compiler it makes sense to allow the same functionality with eslint when that becomes ready.  It will likely involve integrating with this project:

https://github.com/typescript-eslint/typescript-eslint

Given `typescript-eslint` requires a fairly high version to work we will only be able to support this with newer versions of TypeScript; but that's fine.

If someone wants to look at integrating this into the plugin when `typescript-eslint` is sufficiently mature that would be *amazing*! :heart:

https://medium.com/palantir/tslint-in-2019-1a144c2317a9

johnnyreilly:
This PR that adds support for `typescript-eslint` to `create-react-app` could well prove useful as inspiration when someone is looking into adding eslint support directly to `fork-ts-checker-webpack-plugin`

https://github.com/facebook/create-react-app/pull/6513

😀🌻

ux-engineer:
As a new comer with this package, I'm wondering what's the current state of ESLint support?

TypeScript's future is with ESLint instead of TSLint (as linked above). README doesn't even mention ESLint so it seems I'd need to wait before I can give this a try. Although linting integration is optional, I'm not sure if the package would be fully useful without it.

johnnyreilly:
As yet it hasn't been added - eslint TypeScript support is still a very new thing. If someone would like to submit a PR adding support we'd be super grateful.

In the longer term given that tslint is being deprecated, I'd imagine support for tslint with the plugin will be too. But that's going to be some time away.

The main purpose of the plugin is type checking. Linting comes a close second. Until support is added to the plugin there's nothing stopping you doing what `create-react-app` are doing currently; that is to say running `eslint` alongside the `fork-ts-checker-webpack-plugin.`

johnnyreilly:
# A Plan!

Yesterday I took the functionality in place in `create-react-app` for a spin. Based upon that experience I think that TypeScript eslint has got to a pretty good place.

I'm not sure whether I'm going to have time to attempt this in the near future, but I think implementation of this functionality may actually prove fairly straightforward. I've certainly an idea of a rough plan - if someone wants to take have a crack at this and report back that'd be amazing!  Here's the plan:

[...]

PR #305, opened 2019-06-30
add eslint typescript support

johnnyreilly:
This PR is intended to start the journey towards adding eslint TypeScript support: https://github.com/Realytics/fork-ts-checker-webpack-plugin/issues/203

I'm not sure how long it will take to do this as I'm somewhat time poor right now, so all help is appreciated ❤️ 😄 

So, what's happened so far? This:

1. Disabled `tslint --fix` during commits - in future we should consider using `eslint --fix`, but I'd rather do that in a subsequent PR to make this one less noisy.
2. Added a new test project called `project_eslint`.  This contains some simple eslint errors which are visible in VS Code with the ESLint VS Code plugin. 
3. Added an `eslint` option to the plugin and put the mechanism in place in `service.ts` to call it.
4. Added a basic implementation to both `IncrementalChecker` (for Vue users) and `ApiIncrementalChecker`.  
5. Added a first test
6. Tested it on a simple project. All looks good!

You can see what usage of this looks like by looking at this PR: https://github.com/TypeStrong/ts-loader/pull/960 

The plan to deprecate tslint support and remove it in future still stands.  However, before we get to that point I'd like to ship eslint support and ensure it's doing what we need.

Having looked at the code around tslint support, it looks like removing tslint support will be a nice simplification of the codebase.


phryneas:
Heyo, I'll try to take a look at this at the weekend - I'm very preoccupied right now. But I'm watching :)

johnnyreilly:
Cool - I'll see if I can take it for a test drive on a real codebase in the meantime. That said, busy week, might have to wait - just wanted to give a heads up that it was happening 😁

Oh side question:

I noticed that `rpc.js` triggers some errors in the console in vs code sometimes. It doesn't seem to stop tests working but it did get me wondering:

1. Could / should this file be TypeScript instead of js?
2. `worker-rpc` has received updates - should we update fork-ts-checker-webpack-plugin to accommodate at some point?

https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/master/test/integration/helpers/rpc.js

README.md
```diff
@@ -1,445 +1,479 @@
...
-// In a .ts or .vue file...
-import Hello from '@/components/hello.vue'
-```
-
-7. If you are working in **VSCode**, you can get extensions [Vetur](https://marketplace.visualstudio.com/items?itemName=octref.vetur) and [TSLint Vue](https://marketplace.visualstudio.com/items?itemName=prograhammer.tslint-vue) to complete the developer workflow.
-
-## License
-
-MIT
+# Fork TS Checker Webpack Plugin
```
johnnyreilly: sometimes prettier is just evil - I added a few lines and it redoes the whole file 😢

johnnyreilly:
Tested this on a large codebase. It works fine, however it's subject to the performance issues that typescript-eslint seems to have; see here: https://github.com/typescript-eslint/typescript-eslint/issues/389

johnnyreilly:
Would someone be able to review this PR please? I think we're ready to go!

phryneas:
Skimmed it so far, looks good but I might have some suggestions. 
I'll try to set aside some time this evening for a thorough review.

johnnyreilly:
The promise of a review made me do some of the mini refactorings I hadn't yet got round to 😁

.gitignore
```diff
@@ -25,4 +25,3 @@ package-lock.json
 
 # Editor directories and files
 .idea
-.vscode
```
phryneas, line 28: was this deliberate or an accident?

phryneas: (forget it, saw the  .vscode/settings.json below)

README.md
```diff
@@ -61,6 +61,37 @@ to compile files (which traverses dependency graph during compilation) - we have
...
+## ESLint
+
+ESLint is the future of linting in the TypeScript world. ([TSLint is being replaced by ESLint](https://medium.com/palantir/tslint-in-2019-1a144c2317a9).
+https://eslint.org/blog/2019/01/future-typescript-eslint)
+
+If you'd like to use eslint with the plugin, supply this option: `eslint: true` and ensure you have the relevant dependencies installed:
+
+`yarn add eslint @typescript-eslint/parser @typescript-eslint/eslint-plugin --dev`
+
+You should have an `.eslintrc.js` configuration file in your root project directory. Here is a sample configuration for a TypeScript project:
```
phryneas: Maybe "eslint configfile" instead of `.eslintrc.js`, as there could also be a `.eslintrc`?

package.json
```diff
@@ -67,7 +67,6 @@
         "git add"
       ],
       "*.ts": [
-        "tslint --fix",
```
phryneas, line 70: This would mean that .ts files would not get a linter treatment right now.
`eslint --fix`?

src/IncrementalChecker.ts
```diff
@@ -409,4 +417,63 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
...
     );
   }
+
+  public getEsLints(cancellationToken: CancellationToken) {
+    // select files to lint
+    const filesToLint = this.files
+      .keys()
+      .filter(
+        filePath =>
+          !this.files.getData(filePath).linted &&
```
phryneas, line 456: I haven't really worked with the files register but wouldn't that mean that all files are only linted once?

src/service.ts
```diff
@@ -50,30 +52,37 @@ const resolveTypeReferenceDirective = process.env
...
 
+const eslinter =
+  process.env.ESLINT === 'true'
+    ? makeEslinter(JSON.parse(process.env.ESLINT_OPTIONS!))
+    : undefined;
+
 const checker: IncrementalCheckerInterface =
   process.env.USE_INCREMENTAL_API === 'true'
     ? new ApiIncrementalChecker(
         typescript,
```
phryneas: instead of adding even more & re-ordering these options yet again, would it maybe make sense to just create one big "configuration" object with an interface that would satisfy both IncrementalChecker and ApiIncrementalChecker?

phryneas:
Yeah, sorry, I'm quite tired today and not really up for the task. I commented what I could see right of the bat, not everything might be correct and most of it definitely debatable. I'll give it a second look tomorrow.

Props for the "share eslint between checkers" commit, I was definitely going to suggest that after having read it this morning :)

johnnyreilly:
> Yeah, sorry, I'm quite tired today and not really up for the task. I commented what I could see right of the bat, not everything might be correct and most of it definitely debatable. I'll give it a second look tomorrow.

Dude, take your time! I appreciate you looking at all. There's no rush; nothing is on fire 😁

> Props for the "share eslint between checkers" commit, I was definitely going to suggest that after having read it this morning :)

Yeah, my DRY-sense was tingling there; had to deal with it.  Went slightly further in my last commit to share the logic for using ESLint as well. Felt better about myself after 😉

src/ApiIncrementalChecker.ts
```diff
@@ -165,4 +176,23 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
       allLints.map(this.createNormalizedMessageFromRuleFailure)
     );
   }
+
+  public getEsLints(_cancellationToken: CancellationToken) {
+    for (const updatedFile of this.lastUpdatedFiles) {
+      if (this.isFileExcluded(updatedFile)) {
+        continue;
+      }
+
```
piotr-oles, line 207: why don't you use cancellationToken here?

src/ApiIncrementalChecker.ts
```diff
@@ -165,4 +176,23 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
...
+      const lints = this.eslinter!.getLints(updatedFile);
+      if (lints !== undefined) {
+        this.currentEsLintErrors.set(updatedFile, lints);
+      }
+
+      for (const removedFile of this.lastRemovedFiles) {
```
piotr-oles: shouldn't this loop be outside of the `for (const updatedFile of this.lastUpdatedFiles)` loop? :)

src/FilesRegister.ts
```diff
@@ -2,11 +2,12 @@
 import * as ts from 'typescript'; // import for types alone
 // tslint:disable-next-line:no-implicit-dependencies
 import { RuleFailure } from 'tslint'; // import for types alone
+import { CLIEngine } from 'eslint'; // import for types alone
 
 export interface DataShape {
   source?: ts.SourceFile;
   linted: boolean;
-  lints: RuleFailure[];
+  lints: (RuleFailure | CLIEngine.LintReport)[];
```
piotr-oles: Maybe it would be better to have another `eslints` key? We can change it to lints after dropping support for the tslint :)

src/IncrementalChecker.ts
```diff
@@ -400,7 +407,8 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       .keys()
       .reduce(
         (innerLints, filePath) =>
-          innerLints.concat(this.files.getData(filePath).lints),
+          innerLints.concat(this.files.getData(filePath)
+            .lints as RuleFailure[]),
```
piotr-oles: `CLIEngine.LintReport` is same as `RuleFailure`?

src/makeEslinter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  function getLintsForFile(filepath: string) {
+    try {
+      const lints = eslinter.executeOnFiles([filepath]);
+      return lints;
+    } catch (e) {
+      if (
+        FsHelper.existsSync(filepath) &&
+        // check the error type due to file system lag
+        !(e instanceof Error) &&
```
piotr-oles: I saw this code somewhere else - can we extract it to the function? I would call it `isInvalidSourceFileError` or similar

src/makeEslinter.ts
```diff
@@ -0,0 +1,50 @@
+import * as eslinttypes from 'eslint'; // import for types alone
+import { NormalizedMessage } from './NormalizedMessage';
+import { FsHelper } from './FsHelper';
+import { makeCreateNormalizedMessageFromEsLintFailure } from './NormalizedMessageFactories';
+
+export function makeEslinter(eslintOptions: object) {
```
piotr-oles: maybe we could call it `createEslinter` as it's a more common name for factory function in the javascript world :) I see that we use `make` word for a factory of factory function to not double the `create` word :)

piotr-oles (changes requested):
Good job! Just a few comments from me :)

src/makeEslinter.ts
johnnyreilly: Happy to switch it 😄

src/IncrementalChecker.ts
johnnyreilly: No but in this context it can only be that; if I switch to the different key approach that you suggested then this will go away. 😄

src/FilesRegister.ts
johnnyreilly: Yeah I went back and forth on this. Initially I was going for a generics based approach but that ended up being tricky. I think I might go with your suggestion as it's clearer. Will have a play and report back

src/ApiIncrementalChecker.ts
johnnyreilly: Good spot! Will fix

src/ApiIncrementalChecker.ts
johnnyreilly: The `ApiIncrementalChecker` has never used the `cancellationToken`; I can't remember if there's a good reason; but it's always been this way: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/198/files

I just wanted to be consistent :smile:

I guess I could add in a `cancellationToken.throwIfCancellationRequested();` call at the start of the method?

src/makeEslinter.ts
johnnyreilly: ~yeah maybe~

Done

johnnyreilly:
Thanks for the review @piotr-oles - I've made changes 😄

src/ApiIncrementalChecker.ts
piotr-oles: Oh, I didn't know that. I think we should use it - it's a waste of resources to lint stale files. I think we should always iterate over lastRemovedFiles:
```
try {
  for (const updatedFile of this.lastUpdatedFiles) {
    cancellationToken.throwIfCancellationRequested();
    ...
  }
} catch (e) {
  throw e;
} finally {
   for (const removedFile of this.lastRemovedFiles) {
     ...
   }
}
```
or if it doesn't break anything, we can also move iteration over lastRemovedFiles to the top of this function.

johnnyreilly: > or if it doesn't break anything, we can also move iteration over lastRemovedFiles to the top of this function.

Excellent idea - will do that!

phryneas (commented):
So apparently my comments from yesterday were held back by me not "finishing" my review (argh).

So here's what I had written yesterday, I'll come check around in the next few days with a more thorough review.

README.md
johnnyreilly: Yeah good shout

package.json
johnnyreilly: Yeah that's intentional. I do want to do the `eslint --fix` but probably in a subsequent PR; there's a bunch of changes ESLint is suggesting and I'd rather tackle them as a standalone piece of work. I'm okay with us not linting for now.

src/IncrementalChecker.ts
johnnyreilly: I don't think that's the case; @piotr-oles can possibly provide more details.

src/service.ts
johnnyreilly: Yeah I was thinking of doing something like that.... Maybe we should

johnnyreilly:
Thanks for the review @phryneas - I've made changes in line with your suggestions

src/service.ts
johnnyreilly: done!

src/IncrementalChecker.ts
piotr-oles: This plugin will reset all flags and AST cache when mtime of the file will change :) (I think this should be more transparent and clear but we will drop support for IncrementalChecker anyway)

piotr-oles (dismissed):

johnnyreilly:
I've just fixed up the merge conflict; so I think this means I technically need another approval 😄

johnnyreilly:
Hey @piotr-oles / @phryneas,

Are either of you good to approve this PR? I'd like to ship and make sure our release infrastructure is still good now we've moved

package.json
```diff
@@ -67,11 +67,6 @@
...
-        "prettier --write",
-        "git add"
-      ],
-      "*.{yml,json,md,html}": [
```
piotr-oles, line 74: why did you remove yml, json, md, and html from being formatted by prettier?

src/ApiIncrementalChecker.ts
```diff
@@ -165,4 +193,25 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
       allLints.map(this.createNormalizedMessageFromRuleFailure)
     );
   }
+
+  public getEsLints(cancellationToken: CancellationToken) {
+    cancellationToken.throwIfCancellationRequested();
```
piotr-oles: I think this line should be inside `for (const updatedFile of this.lastUpdatedFiles)` on line 205 (before `if` statement)

test/integration/incrementalApi.spec.ts
```diff
@@ -7,6 +7,7 @@
  * */
 import fs from 'fs';
 import * as helpers from './helpers';
+import unixify from 'unixify';
```
piotr-oles: did you use it somewhere?

test/fixtures/project_eslint/src/lib/func.ts
```diff
@@ -0,0 +1,3 @@
+// This code should trigger this error
+// Array type using 'Array<string>' is forbidden. Use 'string[]' instead.eslint(@typescript-eslint/array-type)
+export const arrayOfStringTriggeringError: Array<string> = [];
```
piotr-oles, line 3: I think we are missing test that lint js + ts files

piotr-oles (changes requested):
I just did another round as there are a lot of changes :)

package.json
johnnyreilly: The formatter doesn't work on Windows and messes up the formatting every time. I can add it back in if you like; just I have to comment it out when I'm committing as I can't use it.

src/ApiIncrementalChecker.ts
johnnyreilly: I don't think so - the value that `cancellationToken.throwIfCancellationRequested` depends upon can't change mid loop iteration. At least as I understand it.  It's a `for ... of` loop but given the underlying type is an `Array` JS will not yield control on loop iteration.

Unless there's something I've missed?

test/integration/incrementalApi.spec.ts
johnnyreilly: no - will fix

src/ApiIncrementalChecker.ts
piotr-oles: The thing is that CancellationToken uses synchronous file system calls - the file system is a very simple/primitive way to communicate between processes. Thanks to that we can cancel type-checking/linting in the middle of the process (which is basically a loop over files). The idea of file-base cancellation token came from typescript implementation of CancellationToken :)

package.json
piotr-oles: that's weird - but if you are saying so, I'm okay with removing it

test/fixtures/project_eslint/src/lib/func.ts
johnnyreilly: ~Will do~

Done!

test/integration/incrementalApi.spec.ts
johnnyreilly: done!

johnnyreilly:
I've addressed the review comments!

piotr-oles:
What about CancellationToken? :)

johnnyreilly:
I replied!

src/ApiIncrementalChecker.ts
johnnyreilly: FIxed!

johnnyreilly:
Fixed!

piotr-oles (approved):
Great!

johnnyreilly:
Okay - merged.... Let's see if we get a new version published!

johnnyreilly:
Amazing!

zackschuster:
could `tslint` support be kept as a plugin somehow? i have no plans to migrate as `eslint` is 4x the install size _before_ typescript support & doesn't provide any extra value for me.

johnnyreilly:
See how to use this here: https://blog.johnnyreilly.com/2019/07/typescript-and-eslint-meet-fork-ts-checker-webpack-plugin.html

PR #306, opened 2019-07-09
fix: handle unhandled "write EPIPE" error

deftomat:
Fix the issue when killing Webpack during the running typechecks on Node 12 could cause an unhandled error.

Basically, you can kill Webpack and then, after a few seconds, probably when typecheck finnish, error occurs in terminal.

Untitled 2


deftomat:
Unfortunately, we probably need to upgrade to at least `@types/node: ^11.0.0` to make compiler happy as `process.send()` is missing the last 2 options in `@types/node: ^8.0.0` 😞

johnnyreilly:
Awesome work!   Thanks!

Right now I'm pretty focussed on landing ESLint support in the checker.  Once that's done I'll loop back round to this.  Thanks again!

johnnyreilly:
Could you add node 12 to our travis build matrix please?

johnnyreilly:
~Can you fix the PR up with master please?~

Oooh there's a button that does it!

src/cluster.ts
```diff
@@ -15,7 +15,7 @@ for (let num = 0; num < division; num++) {
   workers.push(
     childProcess.fork(path.resolve(__dirname, './service.js'), [], {
       execArgv: ['--max-old-space-size=' + process.env.MEMORY_LIMIT],
-      env: { ...process.env, WORK_NUMBER: num },
+      env: { ...process.env, WORK_NUMBER: num.toString() },
       stdio: ['inherit', 'inherit', 'inherit', 'ipc']
```
johnnyreilly, line 19: Should this always have been a `string`?

johnnyreilly (approved):

johnnyreilly:
Think we're ready to go @deftomat - I was just wondering if you could explain the implications of `WORK_NUMBER: num.toString()` - why the change?

deftomat:
Hi, `@types/node=11.0.0.` requires `process.env` properties to be string which makes sense as ENV variables cannot be anything else. You can use other types, but just in a process itself. They always be string whey they are passed from environment.

johnnyreilly:
Makes sense; now that I think about it, @arcanis made similar changes here: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/f292028733f87a341c10f41c4d45b713fedc87ad/src/index.ts#L654

Thanks! Should be released soon - just need Travis to do it's thing 😄

deftomat:
AWESOME! Thanks.

JuanmaMenendez:
I had this error and the problem was a domain that was missing on my 'host' file

VickyHax:
> I had this error and the problem was a domain that was missing on my 'host' file

what did you do to fix it @JuanmaMenendez

1714080902120:
Excuse me, what is the trigger for this problem? Is it a nodejs version issue?

PR #320, opened 2019-08-01
feat: ctrl click on error filename sends to the file line

prevostc:
When an error message is shown on an IDE console, most IDEs will allow
you to ctrl+click the file name to go directly to the proper file, line
and character. In order for this to work properly, the error filename,
line and character must have a specific format: "<filename>:<line
number>:<character number>".

### DIFF
--- a/src/formatter/codeframeFormatter.ts
+++ b/src/formatter/codeframeFormatter.ts
@@ -24,7 +24,11 @@ export function createCodeframeFormatter(options: any) {
 
     if (message.code === NormalizedMessage.ERROR_CODE_INTERNAL) {
       return (
-        messageColor(`INTERNAL ${message.severity.toUpperCase()}: `) +
+        messageColor(
+          `INTERNAL ${message.severity.toUpperCase()}(${message.line},${
+            message.character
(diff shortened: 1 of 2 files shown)


johnnyreilly:
This is already supported by the default formatter but the syntax is different; it's like this:

`file.ts(2,6)`

This is a generally supported format for clickable errors. Certainly works in vs code etc. What are you using?

prevostc:
Oh nice this syntax works too \o/ Hold my beer.

johnnyreilly:
Yeah 😀

There's some relevant discussion here: https://github.com/facebook/create-react-app/pull/5624#issuecomment-442418402

prevostc:
@johnnyreilly I just pushed force a new commit to use this format, thanks for the link btw, very interesting.

johnnyreilly:
So this fixes up - the other formatter; makes it consistent with the default one?

prevostc:
@johnnyreilly Exactly. I came here from a frustrating next.js experience :)

johnnyreilly (approved):
Looks good to me!

PR #322, opened 2019-08-01
Log everything to stderr, not stdout

### LINKED ISSUE #321: Info messages are written to stdout, not stderr, which breaks --stats
## Current behavior

The

> Starting type checking service...
> Using 1 worker with 2048MB memory limit

progress/info messages are written to the stdout stream (since the default behavior is to basically call `console.log`).

This breaks using Webpack's `--stats` option, which dumps a blob of JSON statistics to standard output at the end of compilation (which can then be used by e.g. (tooting my own horn a bit here 😄) https://github.com/akx/webpack-xray for analysis).

## Expected behavior

Those status messages should be output to the stderr stream so `webpack ... --stats > stats.json` works correctly without having to manually edit 2 lines out of the blob of JSON.

In a project with this plugin enabled, see how `webpack --progress > foo.txt` writes those two lines into `foo.txt` while other progress is printed out to stderr and not into the file.

## Environment
 - **fork-ts-checker-webpack-plugin**: 1.4.3
 - **typescript**: 3.5.3
 - **tslint**: n/a
 - **webpack**: 4.38.0
 - **os**: all

akx:
Fixes #321.

### DIFF
--- a/src/defaultLogger.ts
+++ b/src/defaultLogger.ts
@@ -1,0 +1,13 @@
+// tslint:disable:no-console
+import { Logger } from '.';
+
+const defaultLogger: Logger = {
+  // since `console.warn` is an alias to `console.error`,
+  // and `console.error` is really only a `console.info` for stderr
+  // instead of stdout, these are all equivalent by default.
+  error: console.error,
+  warn: console.error,
+  info: console.error
+};
+
+export default defaultLogger;
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ import { createDefaultFormatter } from './formatter/defaultFormatter';
 import { createCodeframeFormatter } from './formatter/codeframeFormatter';
 import { fileExistsSync } from './FsHelper';
 import { Message } from './Message';
+import defaultLogger from './defaultLogger';
 
 import {
   getForkTsCheckerWebpackPluginHooks,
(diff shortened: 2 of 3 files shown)


johnnyreilly:
I'm not sure how I feel about this... There is the `silent` option which should allow you to use the plugin without logging to `stderr`

Also, you're able to supply your own logger as is.

akx:
I don't think users should need to configure `logger` or `silent` to be able to use Webpack's `--stats` option.

Why do you feel negatively about this?

piotr-oles:
Looks good for me, I agree with @akx that by default, a developer shouldn't have to configure anything to make webpack's stats work. Also, it doesn't break public API and is a very clean and simple solution.
Am I missing something @johnnyreilly ?  😃

johnnyreilly:
I don't feel super strongly to be honest... If you'd like to take this forward I'm fine with that 😁

jbrantly:
A little bit of related history: https://github.com/TypeStrong/ts-loader/issues/95

johnnyreilly (approved):

johnnyreilly:
I've updated the branch and marked it as approved. If you'd like to merge this and get it out there that's fine with me @piotr-oles 🥰

piotr-oles (approved):

piotr-oles:
It should be released as a new patch bump in the next few hours (there is a lot of tests on different versions of Node, Webpack, TypeScript, and OS). Thank you for your work 👍

PR #337, opened 2019-09-25
feat: support for noEmitOnErrors property in webpack.config

Fatme:
Currently [noEmitOnErrors property](https://webpack.js.org/configuration/optimization/#optimizationnoemitonerrors) is not respected from webpack compilation in case it is set in `webpack.config.js`. As a result, the files are emitted even when there is an error in the source code. However, it turns out that `ts-loader` is working correctly in `transpileOnly: false` mode and the problem is only in case when `ts-loader` is started in `transpileOnly: true` + `ForkTsCheckerWebpackPlugin`.

After investigating the issue, it turns out that the problem is [here](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/4e8c6f98fb14dced5f40a56723b459e519d59362/src/index.ts#L574). Webpack has a logic that checks if the files should be emitted based on `compilation.errors` array. This logic is executed on `shouldEmit` hook as can be seen [here](https://github.com/webpack/webpack/blob/95d21bb39a1e700f28be776b300492c50a169fc7/lib/NoEmitOnErrorsPlugin.js#L9-L11). So, we need to put our reported `diagnostics` and `lints` in `compilation.errors` **before** `shouldEmit`. 

The `afterCompile` hook seems as a perfect solution as it is executed before `shouldEmit` and is an `async` hook. As supporting this feature is important for our error handling story, we implemented the change, spent a little time testing it and it works for our cases.

However, we're not absolutely sure for the impact of the mentioned change so we'd be glad to hear your opinion and suggestions.

Rel to: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/57

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ This plugin provides some custom webpack hooks (all are sync):
 | `fork-ts-checker-service-start-error`   | `serviceStartError`  | Cannot start service                                                           | `error`                                                                    |
 | `fork-ts-checker-service-out-of-memory` | `serviceOutOfMemory` | Service is out of memory                                                       | -                                                                          |
 | `fork-ts-checker-receive`               | `receive`            | Plugin receives diagnostics and lints from service                             | `diagnostics`, `lints`                                                     |
-| `fork-ts-checker-emit`                  | `emit`               | Service will add errors and warnings to webpack compilation ('build' mode)     | `diagnostics`, `lints`, `elapsed`                                          |
+| `fork-ts-checker-after-compile`         | `after-compile`      | Service will add errors and warnings to webpack compilation ('build' mode)     | `diagnostics`, `lints`, `elapsed`                                          |
 | `fork-ts-checker-done`                  | `done`               | Service finished type checking and webpack finished compilation ('watch' mode) | `diagnostics`, `lints`, `elapsed`                                          |
 
 The **Event name** is there for backward compatibility with webpack 2/3. Regardless
--- a/src/index.ts
+++ b/src/index.ts
@@ -571,7 +571,7 @@ class ForkTsCheckerWebpackPlugin {
 
     if ('hooks' in this.compiler) {
       // webpack 4+
-      this.compiler.hooks.emit.tapAsync(checkerPluginName, emit);
+      this.compiler.hooks.afterCompile.tapAsync(checkerPluginName, emit);
     } else {
       // webpack 2 / 3
       this.compiler.plugin('emit', emit);
--- a/src/index.ts
+++ b/src/index.ts
@@ -135,7 +135,7 @@ class ForkTsCheckerWebpackPlugin {
   private diagnostics: NormalizedMessage[] = [];
   private lints: NormalizedMessage[] = [];
 
-  private emitCallback: () => void;
+  private afterCompileCallback: () => void;
   private doneCallback: () => void;
   private typescriptPath: string;
   private typescript: typeof ts;
@@ -184,7 +184,7 @@ class ForkTsCheckerWebpackPlugin {
             options.formatterOptions || {}
           );
 
-    this.emitCallback = this.createNoopEmitCallback();
+    this.afterCompileCallback = this.createNoopAfterCompileCallback();
     this.doneCallback = this.createDoneCallback();
 
     const {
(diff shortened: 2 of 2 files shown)


piotr-oles:
Thank you for your findings and for the PR :)
Unfortunately, changing the hook from `emit` to `afterCompile` is a breaking change in my opinion. There is a custom hook: `fork-ts-checker-emit` - and as it's a part of public API, some developers could assume that it will run on the `emit` event.
Because of that, we need to find a method to deprecate the `fork-ts-checker-emit` hook and add a new one (`fork-ts-checker-after-compile`) + keep default behavior of using webpack's `emit` event (probably we will have to add a new configuration option to control that behavior).

I know that it can be frustrating that we cannot just fix this bug and change the hook, but our responsibility is to introduce breaking changes gracefully :)

Fatme:
Hey @piotr-oles,

Thank you very much for taking a look at this PR.

> but our responsibility is to introduce breaking changes gracefully

I totally agree that we have to ensure seamless transition for the users.

> keep default behavior of using webpack's emit event (probably we will have to add a new configuration option to control that behavior)

Yep, maybe introducing a new configuration option is a good solution. Having that option with default value false will ensure the new version of the plugin will not break existing users and applications. So we need a name for the new option 😸 - maybe `useAfterCompile` or `noEmitOnErrors` or something else? I'm opened to any ideas and suggestions.

As this PR is really important for us and our error handling story, I'll prepare the needed changes once we have an agreement about the name of the new option.

piotr-oles:
I would suggest `useAfterCompileHook`, @johnnyreilly, what do you think?

johnnyreilly:
We could do... I'm just wondering if it might make more sense to do a breaking changes version? After all, do we want to maintain two behaviours in the long term? If we all agree this is a more sensible behaviour then having a breaking changes version seems like a way to not break existing users and move the plugin in the direction we want to go... Thoughts?

Fatme:
Hey @piotr-oles, @johnnyreilly

It seems we have 2 options:
* Release a breaking change version of the plugin - If we agree on that, I suppose it'll be released as a major version e.g `2.0.0`.
* Add a configuration option and support the both behaviors.

The both solutions will work for us and will address our problem. However, personally to me, I prefer releasing a major version and introducing it as a breaking change. That will be the most beneficial solution in the long term.

Let me know about your decision, so I can apply the required changes on the PR.

piotr-oles:
Ok, maybe breaking change is a better option - it will probably not break anything for 99% of users.

Please update the documentation (Plugin hooks section) and add [BREAKING CHANGE](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/CONTRIBUTING.md#examples) section to the commit message :)

Fatme:
@piotr-oles,

I changed the commit message with BREAKING CHANGE: footer and updated the plugin section in the documentation.

Let me know if I've missed something or there is something else that I can do.

johnnyreilly:
Could you catch up this branch with master please?

Fatme:
@piotr-oles, @johnnyreilly,

I've rebased the PR on the latest master branch.

johnnyreilly (dismissed):
Looks good to me!

Fatme:
@johnnyreilly, @piotr-oles,

It seems there are failing CI builds https://travis-ci.org/TypeStrong/fork-ts-checker-webpack-plugin/jobs/602424646?utm_medium=notification&utm_source=github_status:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/babel-code-frame/-/babel-code-frame-6.26.0.tgz: Request failed \"503 Service Unavailable\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/TypeStrong/fork-ts-checker-webpack-plugin/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command "yarn install --frozen-lockfile" failed and exited with 1 during .
```

Maybe we should rerun the travis builds.

johnnyreilly:
Retriggered. Travis has been pretty flaky this week

Fatme:
@johnnyreilly, @piotr-oles 

We have a green build! 🎉 

Do I need to rebase it again on master as there are additional commits merged into master or you can merge the PR directly?

johnnyreilly:
Yes please - update the branch!

johnnyreilly (approved):

piotr-oles (approved):

Fatme:
@piotr-oles, @johnnyreilly,

Thank you very much for merging the PR!
When do you plan to release it officially?

piotr-oles:
It will be released automatically by [`📦 🚀 semantic-release`](https://github.com/semantic-release/semantic-release) when CI build finish :)

Fatme:
@piotr-oles,

What about the version - will it be updated to `2.0.0` as we're releasing a breaking change?

piotr-oles:
The CI is performing very badly last time - I had to re-run the release job. And yes - as you can see semantic-release do all the magic ✨ 😃

Fatme:
Yeep, it's really cool magic! 😃

Issue #323, opened 2019-08-07
How can add the --fix option to eslint?


f-jost:
For tslint we can enable the fix option with `tslintAutoFix` but how use the `fix` option for eslint?

johnnyreilly:
Is it possible to to pass an option to `eslintOptions` to trigger this? https://eslint.org/docs/1.0.0/developer-guide/nodejs-api#cliengine

Or can it be driven from `.eslintrc.js` at all?

viljark:
At the moment it is possible to pass fix option to eslint: https://eslint.org/docs/developer-guide/nodejs-api#cliengine

But as the docs state: 

> Files on disk are never changed regardless of the value of fix. To persist changes to disk, call outputFixes().

Which is not implemented in the codebase yet.

johnnyreilly:
So there's no way to do it right now?

viljark:
What I meant is that this is not implemented in `fork-ts-checker-webpack-plugin` codebase.
For my use case I patched the createEslinter.js with patch-package until official support:
```
diff --git a/node_modules/fork-ts-checker-webpack-plugin/lib/createEslinter.js b/node_modules/fork-ts-checker-webpack-plugin/lib/createEslinter.js
index bb6c1f9..6b8a2f6 100644
--- a/node_modules/fork-ts-checker-webpack-plugin/lib/createEslinter.js
+++ b/node_modules/fork-ts-checker-webpack-plugin/lib/createEslinter.js
@@ -19,6 +19,9 @@ function createEslinter(eslintOptions) {
                 return undefined;
             }
             const lints = eslinter.executeOnFiles([filepath]);
+            if (eslintOptions && eslintOptions.fix) {
+                eslint.CLIEngine.outputFixes(lints);
+            }
             return lints;
         }
         catch (e) {

```

johnnyreilly:
Would you consider sending a PR and we could add that support officially?

viljark:
> Would you consider sending a PR and we could add that support officially?

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/342

PR #342, opened 2019-10-11
feat: support eslint fix (#323)

viljark:


src/createEslinter.ts
```diff
@@ -5,11 +5,11 @@ import { throwIfIsInvalidSourceFileError } from './FsHelper';
 import { makeCreateNormalizedMessageFromEsLintFailure } from './NormalizedMessageFactories';
 import * as path from 'path';
 
-export function createEslinter(eslintOptions: object) {
+export function createEslinter(eslintOptions: eslinttypes.CLIEngine.Options) {
```
johnnyreilly, line 8: Whilst I'd love to take the ESLint options type here I don't believe we can as we then make ESLint a dependency of the plugin. Not everyone uses ESLint and hence that's why we haven't done that. The `object` is necessary to prevent that I believe. Could you revert this back please?

o-alexandrov:
Would be greatly appreciated when released, thank you :)

johnnyreilly:
I'd like to get this out there but we need the points I raised to be addressed

phryneas:
@johnnyreilly  At least in this PR, there are no comments visible - did you start a review and never click finish? Happened to me once.

johnnyreilly (changes requested):

johnnyreilly:
You may be right...

o-alexandrov:
@johnnyreilly why not making that as an `optionalDependency` or `peerDependency` and doing [conditional require](https://stackoverflow.com/a/40277993), not a top-level import?
- while ES6 [dynamic imports](https://github.com/tc39/proposal-dynamic-import) will be available soon, it's still a proposal that's as of now in [stage 4](https://tc39.es/process-document/) 

That way:
- those, who are not using ESLint would not need to install additional deps
- those, who are using ESLint, would need to install an additional dependency

As of now, this issue is a major drawback of using this plugin with `typescript-eslint`

johnnyreilly:
> while ES6 dynamic imports will be available soon, it's still a proposal that's as of now in stage 4

I'm not sure I follow what this refers to?

>  not making that as an optionalDependency or peerDependency

That's worth considering; what would the implications of each choice be?

o-alexandrov:
@johnnyreilly to achieve what you mentioned above (pls see highlighted portion of the quote):
> Whilst I'd love to take the ESLint options type here **`I don't believe we can as we then make ESLint a dependency of the plugin`**. Not everyone uses ESLint and hence that's why we haven't done that. The object is necessary to prevent that I believe. Could you revert this back please?

You can conditionally require ESLint. That way, who doesn't use ESLint, won't need it as a dependency.
Whereas those, who are using ESLint, if you say the plugin then needs ESLint as a dependency, then those users would need to install ESLint as `peerDependencies` or `optionalDependencies` suggest.

> while ES6 dynamic imports will be available soon, it's still a proposal that's as of now in stage 4
> > I'm not sure I follow what this refers to?

In the future, we might be able to use dynamic imports instead of conditional require statement.
- please ignore it, as of now this particular option would add additional complexity

o-alexandrov:
@viljark 
cc @johnnyreilly 
Based on the code changes, the `eslint` module is imported for `typescript` only.

So it would be removed from the runtime, if not used as a value (not for `typescript`), therefore, conditional require in this particular case shouldn't be needed, unless you have other plans (use `eslint` as a value, not a type).

Please correct me, if I'm wrong @viljark

phryneas:
The problem is that from that point on, our types would contain a reference to eslint, which would require people using our types to also install eslint. (And since we're distributing the types with the package, they essentially become dependencies.)
We've had a similar problem with webpack's types, which in the end lead to `@types/node` being installed as a requirement and subsequently polluting the global scope. That's why we're trying to keep our types as independent from other libraries as possible.

Also, as far as I can tell, types cannot be `import`ed dynamically. Either they're in, or not.

o-alexandrov:
@phryneas based on my experience, if you don't have a `@types/eslint` dependency installed, the typescript compiler would process it as `any` type.
Have you actually tried it without the dependency `@types/eslint` installed?

I will also check it later today.

> Also, as far as I can tell, types cannot be imported dynamically. Either they're in, or not.

Please be informed that I never talked about the typings importing dynamically. Let me quote myself from the previous posts:
> Based on the code changes, the eslint module is imported for typescript only.
So it would be removed from the runtime, if not used as a value (not for typescript), therefore, **`conditional require in this particular case shouldn't be needed unless you have other plans (use eslint as a value, not a type).`**,

src/createEslinter.ts
```diff
@@ -5,11 +5,11 @@ import { throwIfIsInvalidSourceFileError } from './FsHelper';
...
   // tslint:disable-next-line:no-implicit-dependencies
   const eslint: typeof eslinttypes = require('eslint');
```
o-alexandrov, line 10: @johnnyreilly you conditionally require `eslint` here, your argument that you then need to make `ESLint` a dependency isn't valid as it is not a top-level import and eslint above is used as the type.

o-alexandrov:
@phryneas just checked the type of `eslinttypes`, if no `@types/eslint` is installed, it works as I mentioned before, it has the `any` type.

Hopefully the following clears the concern, you could refer to the screenshot below:
Screen Shot 2019-10-23 at 3 33 51 PM

But, [you already have](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/package.json#L111) the `@types/eslint` `devDependency` and @viljark doesn't mess with the deps according to the code changes in the PR.

Sorry for pinging too much, I'm just interested to see this PR being released 😆

johnnyreilly:
`devDependency`s are fine, as are internal type usage which is what `createEslinter` is. For the reasons stated, I don't think we're going to be able to make `@types/eslint` a prod dependency of the plugin. We did similar previously and had to back it out because of the issues it caused.

Are you interested in proceeding with the feature without making `@types/eslint` a prod dependency?

phryneas:
So I just created a new project, installed typescript and 'fork-ts-checker-webpack-plugin', set `skipLibCheck: false` and edited `import X from 'eslint';`into the `index.d.ts` (also I used it further down in one of the types).

Result when running `tsc`:

```

node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts:5:15 - error TS2307: Cannot find module 'eslint'.

5 import X from 'eslint';
      
```

(And yes, we have two more of those errors already in there for webpack and tapable, that's over in https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/339 - I hopefully get around to removing those references this weekend.

linxiaowu66:
this feature has any update? @johnnyreilly

piotr-oles:
Resolved in #395

Issue #340, opened 2019-10-08
add option to enable --quiet in eslint


jeffshaver:
In eslint-loader, you have the option to pass `quiet: true` and it will only show errors in the output. When I was trying to switch over to this instead of eslint-loader, there is no option to pass the quiet option. You cannot pass this option via eslintOptions because the CLIEngine doesn't actually accept a quiet option.

johnnyreilly:
That sounds like a potentially useful option!  Would you like to submit a PR?  I suspect this is probably not too difficult to implement.

jeffshaver:
@johnnyreilly i could definitely look to help. I haven't dove into the repo. any suggestions on where to start?

jeffshaver:
@johnnyreilly looks like this is what we want: https://eslint.org/docs/developer-guide/nodejs-api#clienginegeterrorresults

And it looks like all i would need to do is add a quiet option to this plugin, and then check that new option after we call `executeOnFiles`. If true, set the `lints` variable to the result of `getErrorResults`. does that sound about right to you?

jeffshaver:
@johnnyreilly Actually, since we pass around a LintReport object most of the time, we would need to filter right before outputting to the console... One option is to put a filter here before the map: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/1621fd501353cd40ba1bfe5aac96c697ad8f9f74/src/createEslinter.ts#L44

if message.severity !== 2 and the new option is true, then filter it out.

the other option is to use `value.results` if the option is false, and use the result of `CLIEngine.getErrorResults(value.results)` if the option is true

I am leaning toward the latter since eslint is already providing us a way to only get the error results

johnnyreilly:
The latter seems like a better choice; why reimplement what ESLint already has.  Question though; is this not configurable with the `.eslintrc`?

Also, could you expand a little on why a quiet option would be useful? It's not clear to me why you'd like to see warnings inside VS Code (or your editor of choice) but not in the console.

https://github.com/eslint/eslint/issues/905

It's interesting that the option is supported in the CLI but not elsewhere. Wonder why?

https://github.com/webpack-contrib/eslint-loader/blob/master/README.md#quiet

ESLint loader implementation for context:

https://github.com/webpack-contrib/eslint-loader/blob/2dd8b43b2eacbe86b8a0bd984e0888d030263369/src/Linter.js#L99

jeffshaver:
[...]

PR #346, opened 2019-10-21
feat: add eslintQuiet option (#340)

jeffshaver:
This PR adds a quiet option so that warnings get suppressed in the output.

I definitely would like feedback on if this was the right approach. But also, when I went to write an integration test for this, I couldn't get it working correctly. when incrementalTypsciptApi is on, i get warnings like: 

```
{
   rawMessage: '\u001b[1m\u001b[31mERROR in \u001b[39m\u001b[22m\u001b[
1m\u001b[36mundefined(undefined,undefined)\u001b[39m\u001b[22m\u001b[1m\u001b[
31m:\u001b[39m\u001b[22m\n' +
    "\u001b[90mTS5055: \u001b[39mCannot write file '/test-context/src/lib/otherFunc.js' because it would overwrite input file.",
  message: '\u001b[1m\u001b[31mERROR in \u001b[39m\u001b[22m\u001b[1m\u001b[36mundefined(undefined,undefined)\u001b[39m\u001b[22m\u001b[1m\u001b[31m:\u001b[39m\u001b[22m\n' +
    "\u001b[90mTS5055: \u001b[39mCannot write file '/test-context/src/lib/otherFunc.js' because it would overwrite input file.",
  location: { line: undefined, character: undefined },
  file: undefined
}
```


src/createEslinter.ts
```diff
@@ -37,9 +37,13 @@ export function createEslinter(eslintOptions: object) {
       | IterableIterator<eslinttypes.CLIEngine.LintReport>
       | eslinttypes.CLIEngine.LintReport[]
   ) {
+    const eslint: typeof eslinttypes = require('eslint');
```
johnnyreilly, line 40: Is it necessary to re-`require` ESLint here? We require it earlier in `createEslinter`. Is that no longer used with the new approach?

jeffshaver: No, this was an accidental add i think. My bigger questions are with how to test any of this since just adding a test caused tests to start failing.

johnnyreilly:
The approach seems fine; you're using https://eslint.org/docs/developer-guide/nodejs-api#clienginegeterrorresults which looks to be a good call.  Why the integration test is failing is mysterious...  It seems to be an issue with writing to disk:

> `Cannot write file '/test-context/src/lib/otherFunc.js' because it would overwrite input file.`

johnnyreilly:
I thought my change (creating a new fixture) would fix the tests - it seems not 😢 

The issue seems to be related to files in the test context not being cleaned up.  Most perplexing.

piotr-oles:
I'm closing it as I was able to release [v5.0.0-alpha.1](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/tree/v5.0.0-alpha.1) version which allows you to specify issues predicates :)

PR #352, opened 2019-10-28
fix: revert 2.0.0 afterCompileCallback changes

### LINKED ISSUE #351: v2.0.0 breaks async: false
## Current behavior
If I have `async: false`, our `webpack` build (webpack-dev-server` will not finish and just stuck at `wait until bundle finished`

## Expected behavior
In `v1.6.0`, this does not happen.

I am using this plugin with `eslint`, `typescript-eslint`, `webpack` and `ts-pnp` (similar to the setup of `create-react-app`).
I have `async: false` in webpack config and that's when I see problem with 2.0.0. `async:false` is kind of important to me because I want to fail the webpack build when there are type issues (warning or error), also I want webpack to be the reporter of error because I have a decent error format reporter there :). As such I want to block webpack build to finish until this plugin's done with checking type, but somehow after upgrading to 2.0.0 that nolonger happens.

Really sorry the project I'm working on doesn't have a public repo so it would take me time to create a repo for this

## Environment
 - **fork-ts-checker-webpack-plugin**: 2.0.0
 - **typescript**: 3.6.4
 - **webpack**: 4.41.2
 - **os**: Windows 10

P/S: Honestly though the more I think about this the more I want to ask you guys what's your view on this? Do you want to fail a build when there are typescript/linting errors?
I don't know of a way to have `async:true` and still fail webpack compilation build using this plugin` :(, let me know if I miss some config. Thanks!

johnnyreilly:
BREAKING CHANGE: 🧨 afterCompileCallback replaced with emitCallback

✅ Closes: #351

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ This plugin provides some custom webpack hooks (all are sync):
 | `fork-ts-checker-service-start-error`   | `serviceStartError`  | Cannot start service                                                           | `error`                                                                    |
 | `fork-ts-checker-service-out-of-memory` | `serviceOutOfMemory` | Service is out of memory                                                       | -                                                                          |
 | `fork-ts-checker-receive`               | `receive`            | Plugin receives diagnostics and lints from service                             | `diagnostics`, `lints`                                                     |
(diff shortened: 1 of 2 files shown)


johnnyreilly:
Merging without approvals as it's important we get the fix out.

Issue #354, opened 2019-11-01
Support for nativescript-vue-template-compiler


cainrus:
## Feature motivation
Compatibility with nativescript-vue. It uses nativescript-vue-template-compiler instead of vue-template-compiler

## Feature description
Allow to set ```{ vue: {compiler: 'nativescript-vue-template-compiler'}}``` instead of ```vue: true```
To be compatible with nativescript-vue based projects.

## Feature implementation
We need to fix types for vue option from ```boolean``` to ```boolean | { compiler: string }```
Pass vue options to VueProgram.ts if they are available, then update ```require('vue-template-compiler')``` and error message in catch block.

johnnyreilly:
If you'd like to work a PR we'd certainly take a look!

PR #355, opened 2019-11-02
feat: add nativescript-vue-template-compiler support (#354)

cainrus:
PR for #354 

* feat: add nativescript-vue-template-compiler support

* test: add vue related unit and integrations tests

* docs: add information about nativescript-vue-template-compiler support

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,8 +206,8 @@ new ForkTsCheckerWebpackPlugin({
   node doesn't share memory between workers - keep in mind that memory usage will increase. Be aware that in some scenarios increasing workers
   number **can increase checking time**. Default: `ForkTsCheckerWebpackPlugin.ONE_CPU`.
 
-- **vue** `boolean`:
-  If `true`, the linter and compiler will process VueJs single-file-component (.vue) files. See the
+- **vue** `boolean | { enabled: boolean, compiler: string }`:
+  If `true` or `enabled: true`, the linter and compiler will process VueJs single-file-component (.vue) files. See the
   [Vue section](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#vue) further down for information on how to correctly setup your project.
 
 - **useTypescriptIncrementalApi** `boolean`:
@@ -406,6 +406,13 @@ if ('compilers' in compiler) {
 new ForkTsCheckerWebpackPlugin({
   tslint: true,
   vue: true
+});
+```
+Optionally change default [vue-template-compiler](https://github.com/vuejs/vue/tree/dev/packages/vue-template-compiler) to [nativescript-vue-template-compiler](https://github.com/nativescript-vue/nativescript-vue/tree/master/packages/nativescript-vue-template-compiler) if you use [nativescript-vue](https://github.com/nativescript-vue/nativescript-vue)
+```
+new ForkTsCheckerWebpackPlugin({
+  tslint: true,
+  vue: { enabled: true, compiler: 'nativescript-vue-template-compiler' }
 });
 ```
 
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -29,6 +29,7 @@ import {
   IncrementalCheckerParams
 } from './IncrementalCheckerInterface';
 import { createEslinter } from './createEslinter';
+import { VueOptions } from './types/vue-options';
 
 export class IncrementalChecker implements IncrementalCheckerInterface {
   // it's shared between compilations
@@ -70,7 +71,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
   private readonly watchPaths: string[];
   private readonly workNumber: number;
   private readonly workDivision: number;
-  private readonly vue: boolean;
+  private readonly vue: VueOptions;
   private readonly checkSyntacticErrors: boolean;
   private readonly resolveModuleName: ResolveModuleName | undefined;
   private readonly resolveTypeReferenceDirective:
@@ -90,7 +91,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
     watchPaths,
     workNumber = 0,
     workDivision = 1,
-    vue = false,
+    vue,
     checkSyntacticErrors = false,
     resolveModuleName,
     resolveTypeReferenceDirective
@@ -253,7 +254,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
 
   public nextIteration() {
     if (!this.watcher) {
-      const watchExtensions = this.vue
+      const watchExtensions = this.vue.enabled
         ? ['.ts', '.tsx', '.vue']
         : ['.ts', '.tsx'];
       this.watcher = new FilesWatcher(this.watchPaths, watchExtensions);
@@ -285,14 +286,16 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       }
     }
 
-    this.program = this.vue ? this.loadVueProgram() : this.loadDefaultProgram();
+    this.program = this.vue.enabled
+      ? this.loadVueProgram(this.vue)
+      : this.loadDefaultProgram();
 
     if (this.linterConfigFile) {
       this.linter = this.createLinter(this.program!);
     }
   }
 
-  private loadVueProgram() {
+  private loadVueProgram(vueOptions: VueOptions) {
     this.programConfig =
       this.programConfig ||
       VueProgram.loadProgramConfig(
@@ -309,7 +312,8 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       this.watcher!,
       this.program!,
       this.resolveModuleName,
-      this.resolveTypeReferenceDirective
+      this.resolveTypeReferenceDirective,
+      vueOptions
     );
   }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,6 +23,7 @@ import {
   ForkTsCheckerHooks
 } from './hooks';
 import { RunPayload, RunResult, RUN } from './RpcTypes';
+import { VueOptions } from './types/vue-options';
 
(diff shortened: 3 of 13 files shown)


cainrus:
I don't have much experience with yarn.
Do I need to do something with yarn lock file to pass the ci tests?

johnnyreilly:
I believe you've changed the package.json but you didn't do so with yarn and so the yarn.lock file wasn't regenerated. You're going to need to use yarn to regenerate the lock file. Simply executing `yarn` where you would otherwise `npm install` should do it

cainrus:
@johnnyreilly 
Can you take a look for PR?

johnnyreilly:
In holiday right now - will check soon. Thanks!

README.md
```diff
@@ -408,6 +408,13 @@ new ForkTsCheckerWebpackPlugin({
   vue: true
 });
 ```
+Optionally change default [vue-template-compiler](https://github.com/nativescript-vue/nativescript-vue/tree/master/packages/nativescript-vue-template-compiler) to [nativescript-vue-template-compiler](https://github.com/nativescript-vue/nativescript-vue/tree/master/packages/nativescript-vue-template-compiler) if you use [nativescript-vue](https://github.com/nativescript-vue/nativescript-vue)
```
johnnyreilly: Is the first link correct?

README.md
```diff
@@ -206,8 +206,8 @@ new ForkTsCheckerWebpackPlugin({
   node doesn't share memory between workers - keep in mind that memory usage will increase. Be aware that in some scenarios increasing workers
   number **can increase checking time**. Default: `ForkTsCheckerWebpackPlugin.ONE_CPU`.
 
-- **vue** `boolean`:
-  If `true`, the linter and compiler will process VueJs single-file-component (.vue) files. See the
+- **vue** `boolean | object`:
```
johnnyreilly: Could we include the schema of the expected object here please?

src/IncrementalChecker.ts
```diff
@@ -253,7 +254,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
 
   public nextIteration() {
     if (!this.watcher) {
-      const watchExtensions = this.vue
+      const watchExtensions = this.vue.enabled
```
johnnyreilly, line 257: Will this work when people supply `true` instead of `{ enabled: true }`?

Or does `ForkTsCheckerWebpackPlugin.prepareVueOptions(options.vue)` take care of this?

johnnyreilly (changes requested):

johnnyreilly:
Looks pretty good to me; though I'm mindful I'm not too knowledgeable about Vue. cc @yyx990803 in case there's any views on this? (I think this plugin is used in the Vue CLI so thought I'd give a heads up)

README.md
cainrus: That's why code review matters.  Fixed.

README.md
cainrus: Sure.

src/IncrementalChecker.ts
cainrus: ```true``` is enough, these feature is not a breaking change (prev api is the same).
I have covered it with updated [unit test](https://github.com/cainrus/fork-ts-checker-webpack-plugin/blob/feat/support-nativescript-vue-template-compiler/test/unit/VueProgram.spec.js#L120-L136) and [integration tests](https://github.com/cainrus/fork-ts-checker-webpack-plugin/blob/feat/support-nativescript-vue-template-compiler/test/integration/vue.spec.ts)

johnnyreilly (dismissed):
I'm happy with this. If @piotr-oles / @phryneas are cool with this I'm fine with this shipping.  

It would be great to get some Vue user opinion in the mix too

phryneas:
Code-wise, this looks good to me. But I'm not 100% happy with the integration tests, as this doesn't test if the actual compilation happens with `nativescript-vue-template-compiler` at all, if I'm reading it right.

Maybe duplicate the whole test suite with `describe.each` [like in line 25](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blame/aafe8cd6b166e39080ac3c1278bd84de37ee2276/test/integration/vue.spec.ts#L25) and mock the respective "other" compiler into one that throws an Error when called to make sure it's the right one being used? (Or spy on the "right one" to make sure it's being used - or a combination of both.)

src/index.ts
```diff
@@ -311,6 +312,20 @@ class ForkTsCheckerWebpackPlugin {
...
+  ): VueOptions {
+    const defaultVueOptions: VueOptions = {
+      compiler: 'vue-template-compiler',
+      enabled: false
+    };
+    return typeof vueOptions === 'boolean'
+      ? Object.assign(defaultVueOptions, { enabled: vueOptions })
+      : typeof vueOptions === 'object' && vueOptions !== null
+      ? Object.assign(defaultVueOptions, vueOptions)
+      : defaultVueOptions;
```
piotr-oles: Could you use `if` statements instead of `ternary` operator in this function? It would make the code more readable :)

test/integration/vue.spec.ts
```diff
@@ -21,63 +31,103 @@ describe.each([/*[true], */ [false]])(
...
+    const vueEnabledOptions = [true, { enabled: true }];
 
-      const fileNames = await getKnownFileNames();
+    const vueDisabledOptions = [void 0, false, {}, { enabled: false }];
 
-      let fileFound;
-      let fileWeWant = unixify(files['example.vue']);
-      fileFound = fileNames.some(filename => unixify(filename) === fileWeWant);
-      expect(fileFound).toBe(true);
+    vueEnabledOptions.forEach(option => {
```
piotr-oles: Can you change it to use `it.each()` from `jest` instead of `forEach`?

test/integration/vue.spec.ts
```diff
@@ -12,6 +12,16 @@ interface Error {
...
+const printValue = (value: any) => {
+  if (typeof value === 'undefined') {
+    return 'undefined';
+  } else if (typeof value === 'object') {
+    return JSON.stringify(value);
+  } else {
+    return String(value);
+  }
+};
+
```
piotr-oles: I see this code duplicated in the next test - can we move it to the test helpers?

piotr-oles (changes requested):

test/integration/vue.spec.ts
cainrus: It seems that we do not need to use these helper, jest has builtin```%p``` placeholder

src/index.ts
cainrus: Agree. fixed

test/integration/vue.spec.ts
cainrus: Hmm... done, but I don't sure if I need to update old test code to unify code style.

cainrus:
> Maybe duplicate the whole test suite with `describe.each`

Good idea. in progress right now

test/integration/vue.spec.ts
piotr-oles: It would be great if you could. Is this a lot of work?

cainrus: I don't think so, i will do it.

cainrus: Done

test/integration/vue.spec.ts
cainrus: No more ```printValue``` in the code

cainrus:
@johnnyreilly, @phryneas, @piotr-oles 
I think i'm done with review fixes. Do you want to review it again?

test/integration/vue.spec.ts
```diff
@@ -12,6 +12,23 @@ interface Error {
...
+  compiler: typeof templateCompilers[number];
+}
+
+const vueEnabledOptions = templateCompilers.reduce(
+  (list, templateCompiler) => {
+    return list.concat({ enabled: true, compiler: templateCompiler });
+  },
+  [] as VueEnabledOption[]
+);
+
```
phryneas: This isn't very readable, please use something along the lines of
```typescript
const vueEnabledOptions = templateCompilers.map(compiler => ({ enabled: true, compiler}))
```

test/integration/vue.spec.ts
```diff
@@ -21,22 +38,59 @@ describe.each([/*[true], */ [false]])(
         pluginOptions: { ...options.pluginOptions, useTypescriptIncrementalApi }
       });
 
-    it('should create a Vue program config if vue=true', async () => {
-      const { getKnownFileNames, files } = await createCompiler({
-        pluginOptions: { vue: true }
-      });
+    it.each(templateCompilers)(
```
phryneas: Please do not wrap every single test in `it.each` - rather wrap the whole suite in `describe.each`.If a test really should run in only one of those two cases, you can still use something similar to skipIfIncremental from the general test helpers.

phryneas (changes requested):
Just some things I saw at first glance - have to give this a better review later, or tomorrow.

test/integration/vue.spec.ts
cainrus: My point was that we need to reduce the number of unnecessary test runs, but I got your point.
```skipIfIncremental``` is interesting solution, I will use it.

test/integration/vue.spec.ts
cainrus: Code block was removed

test/integration/vue.spec.ts
cainrus: Done.
I did not applied prettier for ```.test/integration/vue.spec.ts``` to simplify code review.
I will apply it later.
```// eslint-disable-next-line @typescript-eslint/array-type``` was added, because current prettier version(or current prettier config) can't parse syntax like ```<T>(a: readonly T[] ) => void;```

phryneas (approved):
looks good to me

piotr-oles (approved):

PR #364, opened 2019-12-11
Refactor normalized message and formatters

piotr-oles:
Hi everyone :)

I would like to make an attempt to rewrite this plugin part by part in order to improve code quality (as there is a big technical debt already).

This is the first (and easiest) part - refactor of `NormalizedMessage` and `formatters`.

The next part will be dropping support for multi-process mode and webpack 2 and 3. 

This will be released as `beta` as I would like to introduce a few more breaking changes later and also it would be nice to test new code on larger projects before going into master :)


johnnyreilly:
Nice! Great to see your plans!

I also think we should drop TSLint support now that it's been deprecated for quite some time and ESLint support is established.

Question: it looks like you've rewritten and renamed `NormalizedMessage` as `Diagnostic`. I wonder if we might want to consider a different name just so it's clear we aren't using TypeScripts own `Diagnostic`? I found myself a little confused as I looked at the API and read the code; thinking that TSLint / ESLint messages were being forced into TypeScript `Diagnostic`; I realize that isn't the case now but I think others will be confused too.

package.json
```diff
@@ -146,7 +146,7 @@
     "vue-class-component": "^6.1.1",
     "vue-loader": "^15.2.4",
     "vue-template-compiler": "^2.5.16",
-    "webpack": "^5.0.0-alpha.5"
```
johnnyreilly, line 149: I'm hesitant about moving to webpack 5 until it's properly out of beta; it's been in alpha / beta for a year now and so it may not be soon

src/IncrementalChecker.ts
```diff
@@ -42,7 +48,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
     eslints: []
   }));
 
-  private linter?: Linter;
+  private linter?: tslint.Linter;
```
johnnyreilly, line 51: Maybe this should be renamed `tslinter` to be specific

src/diagnostic/DiagnosticOrigin.ts
```diff
@@ -0,0 +1,33 @@
+enum DiagnosticOrigin {
```
johnnyreilly: Is there any reason we couldn't just use an object literal here? I have a preference for them over `enum`s unless there's a compelling reason 😉

johnnyreilly (changes requested):
Looks like a good start! I've got some comments - also the tests don't seem to be passing?

package.json
piotr-oles: I did move **from** webpack 5.0.0 **to** webpack 4.0.0 :) I think it's ok, right?

src/IncrementalChecker.ts
piotr-oles: I would keep this change for later refactoring - there are a lot more places like this and I don't want this PR to grow too much :)

src/diagnostic/DiagnosticOrigin.ts
piotr-oles: :+1:

package.json
johnnyreilly: I may have misread that 😉

Yeah webpack 4 is fine for now. Webpack 5 still has a ways to go I think

piotr-oles:
Regarding `Diagnostic` name - do you have any suggestions on how we could name it? I can't find any good name for it other than `Diagnostic` :/

johnnyreilly:
Naming things is hard 😄

Perhaps `Issue`? That's simple and would seem to cover it. Or if you still like having "diagnostic" in the name then perhaps `DiagnosticInfo` to distinguish from TS's own `Diagnostic`?

piotr-oles:
I like the `Issue` name - will change to that :)

README.md
```diff
@@ -178,15 +178,12 @@ new ForkTsCheckerWebpackPlugin({
...
 ```
 
-- **colors** `boolean`:
-  If `false`, disables built-in colors in logger messages. Default: `true`.
-
 - **logger** `object`:
   Logger instance. It should be object that implements method: `error`, `warn`, `info`. Default: `console`.
 
-- **formatter** `'default' | 'codeframe' | ((message: NormalizedMessage, useColors: boolean) => string)`:
+- **formatter** `'default' | 'codeframe' | (diagnostic: Diagnostic) => string)`:
```
johnnyreilly: Should this be `Issue` now?

README.md
```diff
@@ -178,15 +178,12 @@ new ForkTsCheckerWebpackPlugin({
...
   Formatter for diagnostics and lints. By default uses `default` formatter. You can also pass your own formatter as a function
-  (see `src/NormalizedMessage.js` and `src/formatter/` for api reference).
+  (see `src/diagnostic/` and `src/formatter/` for API reference).
```
johnnyreilly: Should this be `issue` also?

src/IncrementalChecker.ts
```diff
@@ -439,19 +434,18 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
...
-    return NormalizedMessage.deduplicate(
-      lints.map(this.createNormalizedMessageFromRuleFailure)
-    );
+    return createIssuesFromTsLintRuleFailures(lints);
   }
 
-  public getEsLints(cancellationToken: CancellationToken) {
+  public async getEsLintIssues(
+    cancellationToken: CancellationToken
+  ): Promise<Issue[]> {
```
johnnyreilly, line 448: I missed this first time round, what's the motivation behind making the method `async`? It doesn't look like it consumes any `async` methods? (Though I am reviewing on a phone so might have missed it)

johnnyreilly (changes requested):
Looks pretty good! - couple of minor points and a question

README.md
piotr-oles: :+1:

README.md
piotr-oles: :+1:

src/IncrementalChecker.ts
piotr-oles: I wanted to make the interface consistent - we have async method for `getTypeScriptIssues` so we should have the same approach to other integrations. Thanks to that we are not leaking internal implementation to the public interface - we assume the "worst" scenario that every integration is potentially asynchronous.

johnnyreilly (approved):
Looks good!

johnnyreilly:
Cool - so should the beta branch now get published by semantic release?

PR #365, opened 2019-12-12
Set-up GitHub Actions

piotr-oles:
I would like to test GitHub Actions as Travis CI is very slow and flaky last time


johnnyreilly:
This looks cool! Looks like we still have some failing tests?

piotr-oles:
The tests were failing on Windows - we didn't run them on the Travis :) I will comment Windows for now as an error occurs on the multi-process mode and we want to drop support for this anyway.

johnnyreilly:
@piotr-oles - did you want to swap the badges in the readme as well? The Travis badge is still there and there isn't a badge in place for the GitHub actions build.

piotr-oles:
Good catch - I will change badges :)

PR #366, opened 2019-12-14
Switch from Travis CI to GitHub CI

piotr-oles:
I did test all jobs on the `beta` branch (including the release job) and it seems to work better and faster :)


johnnyreilly (approved):
Looks good! I need to learn some more GitHub actions stuff. I'm using it for ts-loader but I don't know all the funky syntax you're using.

.github/workflows/main.yml
```diff
@@ -0,0 +1,127 @@
...
+        with:
+          node-version: 12
+
+      - name: Get yarn cache
+        id: yarn-cache
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
```
johnnyreilly, line 20: What's happening here?

.github/workflows/main.yml
```diff
@@ -0,0 +1,127 @@
...
+            ${{ runner.os }}-yarn-
+      
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build project
+        run: yarn build
+
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v1
```
johnnyreilly, line 32: What does this do?

johnnyreilly (approved):

.github/workflows/main.yml
piotr-oles: We upload yarn cache on the cloud to speed up `yarn install` command :)

.github/workflows/main.yml
piotr-oles: We upload build artifacts (which is `lib` directory) to the cloud to re-use them in other jobs.
So we build **only once**, and then we test artifacts on different environments and then release these tested artifacts to the npm.

johnnyreilly: Oh nice!

johnnyreilly: BTW - I didn't notice a build being pushed out after you merged https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/365

Is our semantic release still working?

piotr-oles: It should work. The beta branch was only 3 _chore_ commits ahead so semantic-release didn’t run real release, only did dry run to ensure if setup is correct (it really test it because I had to set new tokens for it to work)

johnnyreilly: Cool. It would be good to test it for real as you're making changes so we know it's all still good. I've kind of given up on npm giving me publishing rights on fork-ts-checker-webpack-plugin. Despite all your efforts as well as my own it still hasn't happened. There's something deeply broken with npm's tech support

It's super frustrating but as long as we've got an automated release process that works we can cope. Hence I'm slightly paranoid about it working because I'm going to be under powered to fix it if it's not 😄

piotr-oles: Yes, the npm tech support wasn't very helpful 🙁 We will test semantic-release after merging #364 into the beta branch 🙂

johnnyreilly: Excellent!

PR #368, opened 2019-12-17
feat: drop support for multi-process mode

piotr-oles:
We drop multi-process mode to simplify code-base and focus more on
optimizing single-process mode.

BREAKING CHANGE: 🧨 Dropped support for multi-process mode (removed `workers` option from
plugin configuration, `workersNumber` argument from serviceStart hook
and pre-computed consts: ONE_CPU, ALL_CPUS, ONE_CPU_FREE, TWO_CPUS_FREE)

PR #369, opened 2019-12-17
Drop `watch` paths option

piotr-oles:
We drop watch paths option to simplify code-base and to switch to
TypeScript API watcher

BREAKING CHANGE: 🧨 Dropped support for `watch` paths option (removed `watch` option from
the plugin-configuration and `watchPaths` argument from the `serviceStart`
hook)

To be merged after #368 :)


piotr-oles:
The PR is ready for merge. @johnnyreilly and @phryneas please take a look :)

johnnyreilly:
Hey @piotr-oles,

As is often the case, I'm doing this from my phone and so may be misreading things. I'm not clear what this PR does? 

My initial reading was that it drops support for the non-APIIncrementalChecker approach. i.e. the original mechanism that powered the plugin. However, we still need that approach to support Vue. But looking again I'm not sure if I got that right. Could you elaborate more on what this does please? So sorry that I haven't followed 😄

What does "TypeScript API watcher` mean in this case?

piotr-oles:
@johnnyreilly , Sorry for misleading description - dropping support for the `IncrementalChecker` is a final goal but this PR is just a small step towards this goal - removing `watch` option from the plugin (which is not supported by TypeScript API as it has its own internal watcher) :) I know that we need IncrementalChecker for Vue support so I will not remove it until we add support for Vue to the ApiIncrementalChecker.

johnnyreilly (approved):
Nice work!

phryneas:
So this removes the watcher, but I'm not 100% sure what the consequence is.
Am I interpreting that correctly that this leaves the `files` organized by the CompilerHost (just implicitly adding new items on `getSourceFile`) and that that array is only used by the linter anyways?

If so, thumbs up from my side. I believe the dependency to `chokidar` would be obsolete then as well.

piotr-oles:
The watcher was used only by the `IncrementalChecker`, it was not used by the `ApiIncrementalChecker`. The usage of the watcher was optional - so if you didn't set the `watch` option, the plugin didn't use watcher at all. The idea of the watcher was to limit I/O operations like checking `mtime` :)

phryneas (approved):
Oh well, then this really can go :)

PR #372, opened 2019-12-28
Drop TsLint support

piotr-oles:
We drop support for the TsLint as it's deprecated and we want to simplify our code-base.

BREAKING CHANGE: 🧨 Dropped support for TsLint
BREAKING CHANGE: 🧨 Removed "tslint" and "tslintAutoFix" option
BREAKING CHANGE: 🧨 Removed "hasTslint" argument for the "waiting" plugin hook
BREAKING CHANGE: 🧨 Removed "tslintPath" argument for the "serviceStart" plugin hook

This PR contains also changes required to add Node.js 6 to our CI/CD matrix.


src/index.ts
```diff
@@ -542,24 +487,18 @@ class ForkTsCheckerWebpackPlugin {
...
-    );
+    forkTsCheckerHooks.serviceStart.call(this.tsconfigPath, this.memoryLimit);
 
     if (!this.silent && this.logger) {
-      this.logger.info(
-        'Starting type checking' +
-          (this.tslint ? ' and linting' : '') +
-          ' service...'
-      );
+      this.logger.info('Starting type checking service...');
```
johnnyreilly, line 501: Do we want a comment here to indicate when linting is happening too?

src/service.ts
```diff
@@ -17,19 +15,22 @@ import { createIssueFromInternalError, Issue } from './issue';
 
 const rpc = new RpcProvider(message => {
   try {
-    process.send!(message, undefined, undefined, error => {
-      if (error) {
-        process.exit();
-      }
-    });
+    if (process.send) {
```
johnnyreilly, line 18: Can this not exist?

johnnyreilly (approved):
Amazing work!

So much less code 🥰

phryneas:
Whoa, that's loads of work. I'm a big fan of this!
I could only review this on my phone, looks great from what I can see. 👍

johnnyreilly:
> ! I could only review this on my phone

Same, same. My review was submitted from a bathroom in Mallorca whilst my family slept 😉

phryneas:
Oh, have a great time there!  For me, it's a hotel room in Leipzig, visiting the chaos communication congress 😄

src/service.ts
piotr-oles: From the Node.js documentation: 

> If Node.js was not spawned with an IPC channel, process.send will be undefined.

I added this check to avoid the not-null assertion - maybe we should throw an Error instead of failing silently.

src/index.ts
piotr-oles: I would leave it as it is in this PR as I plan to refactor plugin logs anyway :)

src/service.ts
johnnyreilly: Hmmmm.... Not sure. Maybe this is fine. We could `console.error` in the case of it being `undefined` but I'm not sure if there's value in that or not.

PR #373, opened 2020-01-08
Chore upgrade micromatch

LoicPernod:
Micromatch 4 has been released with a massive reduction in dependencies:
https://github.com/micromatch/micromatch/issues/137

This PR upgrades micromatch as a dependency in fork-ts-checker-webpack-plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "babel-code-frame": "^6.22.0",
     "chalk": "^2.4.1",
     "chokidar": "^3.3.0",
-    "micromatch": "^3.1.10",
+    "micromatch": "^4.0.2",
     "minimatch": "^3.0.4",
     "semver": "^5.6.0",
     "tapable": "^1.0.0",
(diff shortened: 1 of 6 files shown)


johnnyreilly:
Cool - looks like the tests fail though?

@piotr-oles would it be better for this to be a PR against the beta branch? Incidentally I've been using the beta on some work projects and it may be safe to release that fully. What are your thoughts?

LoicPernod:
I'm going to close this PR and create a new one on beta. However, beta has added Node 6 to the testing matrix and micromatch 4 is no longer compatible with node 6.

Per https://github.com/nodejs/release#end-of-life-releases, Node 6, and Node 8 are end-of-life'd.

Could dropping Node 6 (and 8) be a part of the breaking 4.x release ?

johnnyreilly:
I'd be fine with that - very little supports node 6 these days. @piotr-oles would that be fine with you?

piotr-oles:
My idea for deciding whether we should support given Node.js version is to follow webpack's and TypeScript's support for Node. Currently, the `beta` branch supports webpack 4 which supports Node.js 6. It's just a plugin for the webpack, so I think it should obey webpack's decision. 😄

johnnyreilly:
Okay - I guess we stay with micromatch as is for now. Sorry @lPadier

jimmywarting:
would have like to seen a update to use newer micromatch@4 😞 
our dep. tree is massive... something i'm trying to reduce/speed up

piotr-oles:
Feel free to open a PR, things changed in 2 years, and I think we can upgrade to newer version :)

jimmywarting:
hmm, lol, you don't even depend on micromatch anymore
i'm using a old version of fork-ts-checker-webpack-plugin ^^

PR #385, opened 2020-02-12
fix: prevent null-pointer error after killing child process

### LINKED ISSUE #363: Errors outside of the fork-ts-checker-webpack-plugin cause it to error
## Current behavior

Error in fork-ts-checker-webpack-plugin after an error occurs outside of fork-ts-checker-webpack-plugin. I had a legitimate error in my postcss config file, which was causing the build to bail early. When this happens, I get the following error.

```
/Users/jonathan/my-app/node_modules/fork-ts-checker-webpack-plugin/lib/index.js:459
            return this.serviceRpc.dispatch(message);
                                   ^
TypeError: Cannot read property 'dispatch' of undefined
    at ChildProcess.service.on.message (/Users/jonathan/my-app/node_modules/fork-ts-checker-webpack-plugin/lib/index.js:459:36)
    at ChildProcess.emit (events.js:189:13)
    at emit (internal/child_process.js:820:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

I'm not familiar with fork-ts-checker-webpack-plugin, but it appears a spawned service is receiving a message after shutting down.

## Expected behavior

Don't fail when a message is received after the service is killed.

Forcing an error in postcss.config.js reproduces the error for me, but I imagine forcing an error in any other loader may also reproduce the error.

## Note

The fix likely entails adding a null check to [the `this.serviceRpc!.dispatch` call](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/index.ts#L708), since the RPC provider dispatches in `promise.resolve` callbacks and doesn't appear to have a way to tear itself down.

## Environment
 - **fork-ts-checker-webpack-plugin**: 3.1.0
 - **typescript**: 3.6.4
 - **tslint**: 5.16.0
 - **webpack**: 4.30.0
 - **os**: macOS 10.15.1

piotr-oles:
✅ Closes: #363

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -489,8 +489,12 @@ class ForkTsCheckerWebpackPlugin {
 
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     this.serviceRpc = new RpcProvider(message => this.service!.send(message));
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    this.service.on('message', message => this.serviceRpc!.dispatch(message));
+    this.service.on('message', message => {
+      if (this.serviceRpc) {
+        // ensure that serviceRpc is defined to avoid race-conditions
+        this.serviceRpc.dispatch(message);
+      }
+    });
 
     const forkTsCheckerHooks = ForkTsCheckerWebpackPlugin.getCompilerHooks(
       this.compiler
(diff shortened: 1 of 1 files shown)


johnnyreilly (approved):
Nice one!

PR #389, opened 2020-02-14
feat(vue): add vue 3 support

jods4:
I added support for the new Vue 3 compiler, it seems to work on my project.

How I kept v2 compatibility:
- I changed the `compiler` option to `string | undefined` (it's undefined by default);
- When `undefined`, I try to load `@vue/compiler-sfc` (v3) and if it doesn't work I try `vue-template-compiler` (v2). If the option is set to a string, I just go for that string;
- Then I had to adapt compiler usage to account for slight changes in the public API. Basically I'm testing the presence of each api that differs (everything is fully typed).

⚠️ I have one unrelated error in `TypeScriptIssueFactory.ts`, with the `DiagnosticMessageChain.next`.  being an array rather than just the next instance. Maybe this field type changed in a recent release (I was using TS 3.7.5, which is not the one in `yarn.lock`). Just a heads up, you may have breakage coming your way.


jods4:
I see the CI fails because `yarn.lock` was not updated (I added `@vue/compiler-sfc` to the list of dev deps)... sorry I don't use yarn, if you can update it.

johnnyreilly:
Thanks for the PR @jods4!

What are your thoughts @piotr-oles? Supporting Vue 3 is great. The alpha-ness of it makes me wonder if this might be too soon. Not sure though...

jods4:
I would agree it's to early for an official release, but maybe add Vue 3 support to preview releases for now? Like a `4.1-alpha.1`?

I doubt this API will change in major ways before release but we can adapt if required.

The major snag right now is precisely that the ecosystem is lacking v3 support, so it'd be nice to have this. It will provide tooling so that others (e.g. libs) can get onboard and prepare for the v3 release.

piotr-oles:
To update `yarn.lock`, please just run `yarn install` command - it will update the file automatically :) We can release it on the alpha channel.

src/VueProgram.ts
```diff
@@ -334,8 +353,9 @@ export class VueProgram {
 
     // There is src attribute
     if (script.attrs.src) {
-      // import path cannot be end with '.ts[x]'
-      const src = script.attrs.src.replace(/\.tsx?$/i, '');
+      // Import path cannot end with '.ts[x]'
+      // In v3, `attrs` can be `true` (attribute without a value), but `src` should be a string
+      const src = (<string>script.attrs.src).replace(/\.tsx?$/i, '');
```
piotr-oles: What if someone will add the `src` attribute without value and then will save it? Wouldn't it throw an error inside the `fork-ts-checker-webpack-plugin`? (calling undefined `replace` method)

src/types/vue-options.ts
```diff
@@ -1,4 +1,4 @@
 export interface VueOptions {
   enabled: boolean;
-  compiler: string;
+  compiler?: string;
```
piotr-oles, line 3: could you also update the `README.md`?

piotr-oles (changes requested):

src/VueProgram.ts
jods4: Yes, it would. I was trying to fixup the api changes quickly and it's a bit sloppy.
I agree it would be better not to throw inside `fork-ts-checker-webpack-plugin`, even when user input makes no sense.
Instead of casting we could use the good old trick `src + ""` that turns anything into a string. If that string makes no sense (i.e. is not a real file)... see comment on lines 362-364.

jods4:
@piotr-oles I don't have `yarn` on my machine, just using the plain ol' `npm`, could you do it please?

Also, I'm afraid of what's gonna happen next time you install dependencies: I think there was a breaking change in TS apis.

johnnyreilly:
> @piotr-oles I don't have yarn on my machine, just using the plain ol' npm

It's not high cost to install; instructions are here: https://classic.yarnpkg.com/en/docs/install

It's literally seconds of work 🥰

piotr-oles (approved):

jods4:
Vue 3 is now in beta!
Would love to see this merged.

piotr-oles:
Solved in the #423

PR #390, opened 2020-02-15
fix: properly output chained Typescript errors

connorjsmith:
Updates `flattenDiagnosticMessageText` to properly traverse a `ts.DiagnosticMessageChain.next`, which is an array of `ts.DiagnosticMessageChain` rather than a single object

### DIFF
--- a/src/ApiIncrementalChecker.ts
+++ b/src/ApiIncrementalChecker.ts
@@ -1,3 +1,4 @@
+import * as ts from 'typescript'; // used for types only
 import {
   IncrementalCheckerInterface,
   IncrementalCheckerParams
@@ -13,6 +14,7 @@ import { LintReport } from './types/eslint';
 
 export class ApiIncrementalChecker implements IncrementalCheckerInterface {
   protected readonly tsIncrementalCompiler: CompilerHost;
+  protected readonly typescript: typeof ts;
 
   private currentEsLintErrors = new Map<string, LintReport>();
   private lastUpdatedFiles: string[] = [];
@@ -41,6 +43,8 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
       resolveModuleName,
       resolveTypeReferenceDirective
     );
+
+    this.typescript = typescript;
   }
 
   public hasEsLinter(): boolean {
@@ -60,7 +64,10 @@ export class ApiIncrementalChecker implements IncrementalCheckerInterface {
     this.lastUpdatedFiles = tsDiagnostics.updatedFiles;
     this.lastRemovedFiles = tsDiagnostics.removedFiles;
 
-    return createIssuesFromTsDiagnostics(tsDiagnostics.results);
+    return createIssuesFromTsDiagnostics(
+      tsDiagnostics.results,
+      this.typescript
+    );
   }
 
   public async getEsLintIssues(cancellationToken: CancellationToken) {
--- a/src/IncrementalChecker.ts
+++ b/src/IncrementalChecker.ts
@@ -246,7 +246,7 @@ export class IncrementalChecker implements IncrementalCheckerInterface {
       tsDiagnostics.push(...tsDiagnosticsToRegister);
     });
 
-    return createIssuesFromTsDiagnostics(tsDiagnostics);
+    return createIssuesFromTsDiagnostics(tsDiagnostics, this.typescript);
   }
 
   public async getEsLintIssues(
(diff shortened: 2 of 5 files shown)


johnnyreilly:
Could you provide some more explanation around this please? What's the issue you've experienced and how does this resolve it?

Thanks so much for contributing by the way! (PS the tests are failing if you could take a look)

connorjsmith:
I think the issue stems from using a newer version of typescript.

TS 3.5.1:
```
ERROR in title.tsx
ERROR in title.tsx(181,13):
TS2322: Type 'Element | Text' is not assignable to type 'HTMLElement'.
  Type 'Element' is missing the following properties from type 'HTMLElement': accessKey, accessKeyLabel, autocapitalize, dir, and 110 more.
```
TS 3.7.5:
```
ERROR in title.tsx
ERROR in title.tsx(181,13):
TS2322: Type 'Element | Text' is not assignable to type 'HTMLElement'.
  undefined
```

I think this stems from the definition of `DiagnosticMessageChain.next` becoming an array instead of a single object somewhere along the lines

You can see in the latest typescript.d.ts that this is the case
https://github.com/microsoft/TypeScript/blob/master/lib/typescript.d.ts#L2565

connorjsmith:
Ah yes, our yarn.lock is pinned to TS 3.5.2, hence the CI errors that I didn't see locally 😛
```
typescript@^3.0.1:
  version "3.5.2"
  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.5.2.tgz#a09e1dc69bc9551cadf17dba10ee42cf55e5d56c"
  integrity sha512-7KxJovlYhTX5RaRbUdkAXN1KUZ8PwWlTzQdHV6xNqvuFOs7+WBo10TQUqT19Q/Jz2hk5v9TQDIhyLhhJY4p5AA==
```

johnnyreilly:
I'm afraid I'm still wondering what the issue you're experiencing is? 😄

Is something not working on your machine?

connorjsmith:
In TS 3.5.2, we're getting the proper "extended" error text (the last line in the code block I sent above)

In TS 3.7.5, the last line is being output as `undefined` due changes in the info object being provided by typescript.

The fact that an error appears at all is unrelated to the library 😛

johnnyreilly:
Gotcha - looks like the build is erroring? Also, did this API change over time? If so then we need to cater for both versions of the API

connorjsmith:
@johnnyreilly Build seems to be green, added a path for both newer and older TS versions.

edit: FWIW, we might want to spin up another CI check that tests using the latest version of Typescript

johnnyreilly:
Thanks @connorjsmith! @piotr-oles what do you think?

piotr-oles:
I think it would be better to revert the change that I made on the last refactoring and just use `flattenDiagnosticMessageText` provided by the TypeScript instance. We would need to pass a typescript instance as an argument to the IssueFactory. It's how it was implemented in version 3.

@connorjsmith  Would you have time to take that approach? It would be a safer approach in my opinion.

connorjsmith:
Sure, I'll take a look :)

connorjsmith:
FYI I tried all minor versions of TS from 3.5 to 3.8, and we currently have a failing test unrelated to these changes if you use v3.8.2 (using Powershell on Windows 10)
```
● [INTEGRATION] specific tests for useTypescriptIncrementalApi: true › should find global errors even when checkSyntacticErrors is false (can only be tested on case-insensitive file systems)

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      82 |           )
      83 |         );
    > 84 |         expect(globalErrorFoundInStats).toBe(true);
         |                                         ^
      85 |         callback();
      86 |       });
      87 |     }
```

piotr-oles (approved):

piotr-oles:
> FYI I tried all minor versions of TS from 3.5 to 3.8, and we currently have a failing test unrelated to these changes if you use v3.8.2 (using Powershell on Windows 10)
> 
> ```
> ● [INTEGRATION] specific tests for useTypescriptIncrementalApi: true › should find global errors even when checkSyntacticErrors is false (can only be tested on case-insensitive file systems)
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: true
>     Received: false
> 
>       82 |           )
>       83 |         );
>     > 84 |         expect(globalErrorFoundInStats).toBe(true);
>          |                                         ^
>       85 |         callback();
>       86 |       });
>       87 |     }
> ```

Thanks for finding this out, I will address that in the v5.0.0 release :)

johnnyreilly (approved):

johnnyreilly:
Nice work!

PR #395, opened 2020-03-01
feat: support eslint fix (#323)

### LINKED ISSUE #323: How can add the --fix option to eslint?

Methuselah96:
Reopening https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/342 without the dependency on @types/eslint.

Resolves https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/323.

### DIFF
--- a/src/types/eslint.ts
+++ b/src/types/eslint.ts
@@ -34,3 +34,65 @@ export interface LintReport {
   fixableErrorCount: number;
   fixableWarningCount: number;
 }
+
+export interface Options {
+  allowInlineConfig?: boolean;
+  baseConfig?: false | { [name: string]: any };
+  cache?: boolean;
+  cacheFile?: string;
+  cacheLocation?: string;
+  configFile?: string;
+  cwd?: string;
+  envs?: string[];
+  errorOnUnmatchedPattern?: boolean;
+  extensions?: string[];
+  fix?: boolean;
+  globals?: string[];
+  ignore?: boolean;
+  ignorePath?: string;
+  ignorePattern?: string | string[];
+  useEslintrc?: boolean;
+  parser?: string;
+  parserOptions?: ParserOptions;
+  plugins?: string[];
+  resolvePluginsRelativeTo?: string;
+  rules?: {
+    [name: string]: RuleLevel | RuleLevelAndOptions;
+  };
+  rulePaths?: string[];
+  reportUnusedDisableDirectives?: boolean;
+}
+
+interface ParserOptions {
+  ecmaVersion?:
+    | 3
+    | 5
+    | 6
+    | 7
+    | 8
+    | 9
+    | 10
+    | 11
+    | 2015
+    | 2016
+    | 2017
+    | 2018
+    | 2019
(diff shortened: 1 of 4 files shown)


src/types/eslint.ts
```diff
@@ -34,3 +34,65 @@ export interface LintReport {
   fixableErrorCount: number;
   fixableWarningCount: number;
 }
+
+export interface Options {
```
johnnyreilly, line 38: I've got mixed feelings about duplicating ESLint options inside fork-ts-checker-webpack-plugin. There's then a need to keep them up to date. Incidentally; where specifically were these copied from please?

johnnyreilly (commented):
What are your thoughts @piotr-oles?

src/types/eslint.ts
Methuselah96: Yeah, I was just copying the style that was used in https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/380 by copying the types since it seems as if we're opposed to making a dependency on `@types/eslint`. They're copied from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/eslint/index.d.ts.

What would you think about just having the `Options` interface be:
```
export interface EslintOptions {
    fix?: boolean;
    // the rest of the properties are not specified here because they are not directly used by this package and are instead just passed to eslint
    // we do this in order to avoid a dependency on @types/eslint (since the use of eslint is optional) and to avoid copying types from @types/eslint
}
```

I personally don't see an issue with having a dependency on `@types/eslint`, but that's up to you guys. `@types/eslint` is fairly small, it doesn't pollute the types like `@types/node` does, and it's not going to be bundled into a web app, so I don't think the negative effects of having it as a dependency are very large. That way the user of this package actually gets accurate type-checking when they're using this package.

The final option would be to have part of the build script copy the types out of `@types/eslint` and into this package, but that sounds harder than I (or anybody else) has time to look into.

Methuselah96: @johnnyreilly I care more about it getting done than how we do it. Would your preference be to go without any types and just do:
```
if (eslintOptions && (eslintOptions as any).fix) {
    eslinter.outputFixes(lintReport);
}
```
I'll do it that way if that's what you prefer.

johnnyreilly: I think I'd prefer the `any` (can't believe I typed that!) but I'm not certain.  @piotr-oles what do you think?  I'm super snowed so I'm sorry I'm giving this less attention than I might otherwise.

piotr-oles: Maybe we could create a basic version of EsLintOptions?
```typescript
export interface Options {
  fix?: boolean;
  // and any other option that we rely on

  // open for extension, closed for modification :D
  [key: string]: any;
}
```

Methuselah96: Done.

piotr-oles (approved):

PR #405, opened 2020-04-26
Don't inherit --inspect flag from the host process

Janpot:
To avoid conflicting ports like https://github.com/zeit/next.js/issues/11821

`fork-ts-checker-webpack-plugin` passes all env variables from the host process to its subprocess. If the host process was started with `NODE_OPTIONS` containing an `--inspect` flag, it will try to start a debugger on the same port as the host process is trying to start a debugger. One of the processes will fail to start since the port is already in use. This code filters out this CLI flag from the `NODE_OPTIONS` env variable if it exists for the host process.


piotr-oles (approved):

Janpot:
Before this can be merged, we should probably look into https://github.com/zeit/next.js/pull/12555 as I think it applies to this PR as well

piotr-oles:
@Janpot 
Could you check if this issue was maybe solved with the [v5.0.0-alpha.1](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/tree/v5.0.0-alpha.1) release? :)

Janpot:
I don't think it's solved, but next doesn't enable this plugin anymore. Will reopen when/if I run into this again

PR #409, opened 2020-05-16
refactor(dependency): update babel-code-frame to @babel/code-frame

### LINKED ISSUE #408: Bump babel dependency to @babel/code-frame
## Feature motivation

I have a large project that's having issues with conflicting versions of `chalk` being installed. Wrong versions are being picked up in places (which is the individual package's fault).

`fork-ts-checker-webpack-plugin` has a dependency on `babel-code-frame` which in turn pulls in `chalk@1`. For me this is the only package that uses `chalk@1` so I'd like to be able to update it.

Looks like the babel package has now been renamed to `@babel/code-frame` which is where any updates are being published. That depends on `chalk@2` so allows me to update and hopefully drop the conflict.

## Feature description

Reference the new `@babel/code-frame` package in package.json to allow users to upgrade it

andyrooger:
Update code-frame dependency from v6 to v7 so that we can drop the chalk v1 dependency.

Closes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ new ForkTsCheckerWebpackPlugin({
   (see `src/issue/` and `src/formatter/` for API reference).
 
 - **formatterOptions** `object`:
-  Options passed to formatters (currently only `codeframe` - see [available options](https://www.npmjs.com/package/babel-code-frame#options))
+  Options passed to formatters (currently only `codeframe` - see [available options](https://babeljs.io/docs/en/next/babel-code-frame.html#options))
 
 - **silent** `boolean`:
   If `true`, logger will not be used. Default: `false`.
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     }
   },
   "dependencies": {
-    "babel-code-frame": "^6.22.0",
+    "@babel/code-frame": "^7.5.5",
     "chalk": "^2.4.1",
     "micromatch": "^3.1.10",
     "minimatch": "^3.0.4",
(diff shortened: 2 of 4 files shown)


andyrooger:
I've set the dependency to ^7.5.5 because ^7.0.0 has issues with importing @babel/highlight. I can specify latest instead if you prefer though.

johnnyreilly (dismissed):

johnnyreilly:
We seem to have a failing test?

andyrooger:
We do... It's failing to clean up after the tests by the look of it.

Each time I run these actions it seems to fail on a different OS using webpack 5, node 8. Can see if it does the same this time, but I'm wondering if this is just an issue with webpack alpha finishing when we expect.

Is there a good way to try these tests on latest master to see if they're still passing repeatably?

andyrooger:
So they're all passing now... all I've changed is the commit author.

johnnyreilly:
Flaky tests are the worst! @piotr-oles are you happy for us to merge?

piotr-oles (approved):

piotr-oles:
Looks good. I think we should update the semantic-release config to trigger patch release for `refactor` commits. Give me a few minutes, I will send you what to change.

piotr-oles:
I have created a new PR: #410 to update the semantic-release configuration. We can merge this PR later.

andyrooger:
In case you hadn't already noticed, it looks like refactors don't appear in the release notes even though they do trigger releases. The github release looks blank for 4.1.4. Release itself is working great for me though. :)

Issue #412, opened 2020-05-23
Error: Cannot find module './ForkTsCheckerWebpackPlugin' on @alpha


IgnusG:
## Current behavior
Running webpack crashes immediately with:

<details>
<summary>Stacktrace</summary>
<pre>
/.../node_modules/webpack-cli/bin/cli.js:93
                                throw err;
                                ^

Error: Cannot find module './ForkTsCheckerWebpackPlugin'
Require stack:
- /.../node_modules/fork-ts-checker-webpack-plugin/lib/index.js
- /.../webpack.config.js
- /.../node_modules/webpack-cli/bin/utils/convert-argv.js
- /.../node_modules/webpack-cli/bin/cli.js
- /.../node_modules/webpack/bin/webpack.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1020:15)
    at Function.Module._load (internal/modules/cjs/loader.js:890:27)
    at Module.require (internal/modules/cjs/loader.js:1080:19)
    at require (/.../node_modules/v8-compile-cache/v8-compile-cache.js:161:20)
    at Object.<anonymous> (/.../node_modules/fork-ts-checker-webpack-plugin/lib/index.js:2:38)
    at Module._compile (/.../node_modules/v8-compile-cache/v8-compile-cache.js:192:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1196:10)
    at Module.load (internal/modules/cjs/loader.js:1040:32)
    at Function.Module._load (internal/modules/cjs/loader.js:929:14)
    at Module.require (internal/modules/cjs/loader.js:1080:19)
    at require (/.../node_modules/v8-compile-cache/v8-compile-cache.js:161:20)
    at Object.<anonymous> (/.../webpack.config.js:6:29)
    at Module._compile (/.../node_modules/v8-compile-cache/v8-compile-cache.js:192:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1196:10)
    at Module.load (internal/modules/cjs/loader.js:1040:32)
    at Function.Module._load (internal/modules/cjs/loader.js:929:14)
    at Module.require (internal/modules/cjs/loader.js:1080:19)
    at require (/.../node_modules/v8-compile-cache/v8-compile-cache.js:161:20)
    at WEBPACK_OPTIONS (/.../node_modules/webpack-cli/bin/utils/convert-argv.js:114:13)
[...]

PR #413, opened 2020-05-24
fix: resolve issue with publishing empty package on the npm

piotr-oles:
Changed "files" field in the package.json from "./lib" back to "lib". It seems that there is different behavior between npm and yarn.

Closes: #412


johnnyreilly:
nice work!

PR #415, opened 2020-05-25
fix: close service on the `failed` hook

### LINKED ISSUE #411: [bug]webpack will hang out and not break progress if set bail=true
"webpack": "^4.41.5"
"fork-ts-checker-webpack-plugin": "^4.1.3"

```js
module.exports = {
    ...
    bail: true,
    plugins: [
            async: true,
        }),
        ...
    ],
    ...
}
```
image

piotr-oles:
When a compilation is not successful and webpack is not watching, the plugin should close the service.

✅ Closes: #411


johnnyreilly (approved):
Nice work! BTW I've tweeted about the v5 alpha to see if we can get some more testers 🌻🤗

https://www.twitter.com/johnny_reilly/status/1264862046975336450

PR #418, opened 2020-05-25
feat: remove Yarn PnP extension as it's redundant

piotr-oles:
Thanks to yarn pnpify, we don't need to patch TypeScript anymore.

BREAKING CHANGE: 🧨 Removed Yarn PnP extension (and related options) as it became
redundant


arcanis (approved):
To be exact, PnPify isn't what makes Yarn works with TypeScript. Instead, Yarn ships with a limited set of patchfiles (that we maintain ourselves) that are applied on known "broken" packages.

TypeScript is one of them, and its patchfile applies an automatic merge of https://github.com/microsoft/TypeScript/pull/35206. Same result, except that you don't even have to use PnPify to have TypeScript support 😃

piotr-oles:
@arcanis Thanks for the clarification, I will update the commit message on merge :)

johnnyreilly (approved):

Issue #414, opened 2020-05-24
5.0.0-alpha.2: ESLint support only seems to work on initial compile, not on changes


eamodio:
## Current behavior

ESLint seems to only run on the initial compile, not on recompile (watch) after changes

## Expected behavior

ESLint should run in both scenarios

## Steps to reproduce the issue

Enable ESlint, have no ESLint issues on initial watch, then add an ESLint error

## Environment
 - **fork-ts-checker-webpack-plugin**: [5.0.0-alpha.2]
 - **typescript**: [3.9.3]
 - **eslint**: [7.1.0]
 - **@typescript-eslint/eslint-plugin**: [3.0.0]
 - **webpack**: [4.43.0]
 - **os**: [Windows 10]

piotr-oles:
Could you send a `webpack.config.js` file?

eamodio:
https://github.com/eamodio/vscode-gitlens/blob/develop/webpack.config.js

eamodio:
Looks great -- thanks!

PR #419, opened 2020-05-25
fix: handle absolute paths for the eslint.files option

piotr-oles:
✅ Closes: #414

### DIFF
--- a/src/eslint-reporter/EsLintReporterConfiguration.ts
+++ b/src/eslint-reporter/EsLintReporterConfiguration.ts
@@ -1,4 +1,5 @@
 import webpack from 'webpack';
+import { isAbsolute, join } from 'path';
 import { EsLintReporterOptions } from './EsLintReporterOptions';
 import { CLIEngineOptions } from './types/eslint';
 
@@ -7,7 +8,6 @@ interface EsLintReporterConfiguration {
   memoryLimit: number;
   options: CLIEngineOptions;
   files: string[];
-  cwd: string;
 }
 
 function castToArray<T>(value: T | T[] | undefined): T[] {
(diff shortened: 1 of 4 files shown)


src/eslint-reporter/EsLintReporterConfiguration.ts
```diff
@@ -28,8 +28,10 @@ function createEsLintReporterConfiguration(
     enabled: !!(options && options.enabled === true),
     memoryLimit: 2048,
     ...(typeof options === 'object' ? options : {}),
-    files: typeof options === 'object' ? castToArray(options.files) : [],
-    cwd: compiler.options.context || process.cwd(),
+    files: (typeof options === 'object' ? castToArray(options.files) : []).map((files) =>
```
johnnyreilly, line 31:
```suggestion
    files: (typeof options === 'object' ? castToArray(options.files) : []).map((file) =>
    cwd: compiler.options.context || process.cwd(),	      // ensure that `file` is an absolute path
      isAbsolute(file) ? file : join(compiler.options.context || process.cwd(), file)
    ),
```

src/eslint-reporter/EsLintReporterConfiguration.ts
```diff
@@ -28,8 +28,10 @@ function createEsLintReporterConfiguration(
...
+      // ensure that `files` is an absolute path
+      isAbsolute(files) ? files : join(compiler.options.context || process.cwd(), files)
+    ),
```
johnnyreilly: Should `files` be `file` here? i.e. single rather than plural?  It's always a single file in this context I think?

```suggestion
    files: (typeof options === 'object' ? castToArray(options.files) : []).map((file) =>
    cwd: compiler.options.context || process.cwd(),	      // ensure that `file` is an absolute path
      isAbsolute(file) ? file : join(compiler.options.context || process.cwd(), file)
    ),
```

johnnyreilly (approved):

src/eslint-reporter/EsLintReporterConfiguration.ts
piotr-oles: It's a glob pattern, so it could represent multiple files. Maybe `filesPattern` or `pattern` would be better?

Issue #339, opened 2019-10-07
Could not find a declaration file for module 'webpack' / 'tapable'


IlCallo:
## Current behavior
```
ERROR in /.../node_modules/fork-ts-checker-webpack-plugin/lib/hooks.d.ts(1,26):
TS7016: Could not find a declaration file for module 'webpack'. '/.../node_modules/webpack/lib/webpack.js' implicitly has an 'any' type.
  Try `npm install @types/webpack` if it exists or add a new declaration (.d.ts) file containing `declare module 'webpack';`
ERROR in /.../node_modules/fork-ts-checker-webpack-plugin/lib/hooks.d.ts(2,43):
TS7016: Could not find a declaration file for module 'tapable'. '/.../node_modules/tapable/lib/index.js' implicitly has an 'any' type.
  Try `npm install @types/tapable` if it exists or add a new declaration (.d.ts) file containing `declare module 'tapable';`
Version: typescript 3.6.3
Time: 5433ms
```

## Expected behavior
No errors by default, or a `peerDependencies` warning fired off, or a disclaimer in setup section of the README.

## Steps to reproduce the issue
Not really sure, I just updated `fork-ts-checker-webpack-plugin` to `1.5.0` from `0.5.2` and the error popped up.
The error doesn't show up in `1.3.1` because of the typings dependencies being added, then removed again in `1.3.2`.
The problematic declaration seems to be into `lib/hooks.d.ts`, where `webpack` and `tapable` are types are imported.

```ts
import * as webpack from 'webpack';
import { AsyncSeriesHook, SyncHook } from 'tapable';
export declare type ForkTsCheckerHooks = 'serviceBeforeStart' | 'cancel' | 'serviceStartError' | 'waiting' | 'serviceStart' | 'receive' | 'serviceOutOfMemory' | 'emit' | 'done';
declare type ForkTsCheckerLegacyHookMap = Record<ForkTsCheckerHooks, string>;
export declare const legacyHookMap: ForkTsCheckerLegacyHookMap;
export declare function getForkTsCheckerWebpackPluginHooks(compiler: webpack.Compiler): Record<ForkTsCheckerHooks, SyncHook<any, any, any> | AsyncSeriesHook<any, any, any>>;
export {};
```

Manually adding the dependencies fix the errors.

Related:
- https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/192
[...]

PR #422, opened 2020-05-25
fix: remove references to external typings

piotr-oles:
Fixes E2E test to ensure that the package typings can be consumed without need to install external packages except `@types/webpack`

Solves: #339


README.md
```diff
@@ -349,6 +349,20 @@ hooks.waiting.tap('yourListenerName', () => {
 });
 ```
 
+## Typings
+
+To use the plugin typings, you have to install `@types/webpack`. It's not included by default to not collide with your
+existing typings (`@types/webpack` imports `@types/node`). [It's an old TypeScript issue](https://github.com/microsoft/TypeScript/issues/18588), 
+the alternative is to set `skipLibCheck: true` in the `compilerOptions` 😉
```
johnnyreilly, line 356: Did you know that `skipLibCheck: true` is the default as of TypeScript 3.9?

https://www.twitter.com/orta/status/1260583041740812288

johnnyreilly (approved):

README.md
piotr-oles: Good to know 😃

Issue #269, opened 2019-04-23
Add an example `tasks.json` to README?


phryneas:
This is just something I was missing and that kept me personally from using fork-ts-checker-webpack-plugin in a few projects: integration with with the "Problems" tab in vscode.

But as I'm starting to dogfood now, too, I did a bit of googling and found https://code.visualstudio.com/docs/editor/tasks#_defining-a-multiline-problem-matcher which describes the process of defining `yarn watch` as a vscode task with a custom problem matcher.

That left me with the following configuration:

`.vscode/tasks.json`
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "watch",
      "command": "yarn",
      "args": ["watch"],
      "isBackground": true,
      "problemMatcher": [
        {
          "owner": "ts-loader",
          "fileLocation": "absolute",
          "pattern": [
            {
              "regexp": "^\\[tsl\\] (ERROR|WARNING) in (.*)\\((\\d+,\\d+)\\)$",
              "file": 2,
              "location": 3,
              "severity": 1
            },
            { "regexp": "^\\s+(.+): (.*)$", "code": 1, "message": 2 }
          ],
          "background": {
            "activeOnStart": true,
            "beginsPattern": "Type checking and linting in progress\\.\\.\\.",
            "endsPattern": "^Time: \\d+ms"
          }
        },
        {
          "owner": "fork-ts-checker-webpack-plugin",
          "fileLocation": "absolute",
          "pattern": [
            {
              "regexp": "^(ERROR|WARNING) in (.*)\\((\\d+,\\d+)\\):$",
              "file": 2,
              "location": 3,
              "severity": 1
            },
            { "regexp": "^(.+): (.*)$", "code": 1, "message": 2 }
          ],
          "background": {
            "activeOnStart": true,
            "beginsPattern": "Type checking and linting in progress\\.\\.\\.",
            "endsPattern": "^Time: \\d+ms"
          }
        }
      ]
    }
  ]
}
```

Which raises the question: do we want to add that to the README? While it might be very useful for many people, the README is also getting very long and this is very tool-specific information.

As an alternative: what do you think about splitting some parts of the README out into additional markdown files and link those from a FAQ section in the README?

CODE_OF_CONDUCT.md and CONTRIBUTING.md are essentially first steps into that direction.

johnnyreilly:
> Which raises the question: do we want to add that to the README? 

[...]

Issue #171, opened 2018-10-21
Add example for new Babel 7 TypeScript Integration


fabb:
Babel 7 supports TypeScript without type checking, so basically it just strips all the types away and makes the code valid JS code. It still would be preferrable to do type checking with `ts` in a separate process during the build. It would ve nice if in the readme of this project there was an example on how to configure this correctly (if it is possible).

antch:
I think the same pattern could be applied for linting along with `AngularCompilerPlugin` as well. I use `ForkTsCheckerWebpackPlugin` for non-Angular TypeScript files but I'm stuck.

kenjiru:
Does this plugin even work with Babel 7? `webpack --watch` doesn't work anymore when enabling this plugin.

johnnyreilly:
I believe that create-react-app are using it - so could be worth ejecting one of their apps to get config.

huy-nguyen:
I do use babel 7's typesceript transpilation and this is my settings for development. For production, I run `tsc --noEmit` instead of using the plugin.
```javascript
import ForkTSCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin';
// ...
export default () => {
  return {
    module: {
      rules: [
        {
          test: /\.(ts|tsx)$/,
          use: [
            loader: 'babel-loader',
            options: {
              presets: [
                '@babel/preset-typescript',
              ]
            }
          ],
          exclude: /node_modules/,
        } 
      ],
      // ... other rules
    },
    plugins: [
      new ForkTSCheckerWebpackPlugin({
        checkSyntacticErrors: true,    
      }),
      // ... other plugins
    ]
  }
}
```

bestseob93:
oops.. this was typescript problem.
typescript version: 3.4.1 is very slow.

PR #427, opened 2020-05-27
Add examples

piotr-oles:
* docs: add vscode-tasks example 
* docs: add ts-loader example
* docs: add babel-loader example 
* fix: respect eslint extensions in incremental check

Closes: #269
Closes: #171


johnnyreilly (approved):
Nice work!

PR #428, opened 2020-05-27
feat: export `version` const for external tools

piotr-oles:
The difference between version 4 and 5 is big, so we need to provide some API to allow external tools to detect which version of the plugin is used.


johnnyreilly (approved):
Very clever!

PR #429, opened 2020-05-27
feat: add "mode" for typescript reporter

piotr-oles:
To improve initial compilation time on babel-loader, we can now specify "write-tsbuildinfo" and "write-references" mode.


piotr-oles:
I'm merging this as it's only an alpha release. If you have any comments, I will address them in next PRs :)

Issue #297, opened 2019-06-04
Equivalent for ts-loader's context


stavalfi:
## Feature motivation
https://github.com/TypeStrong/ts-loader#context-string-defaultundefined
This is a killer feature in complex projects.

In my specific use-case, I moved all my build related files (including tsconfig.json) to a different package (something like create-react-app is doing but in much more flexible way) in my mono-repo and everything is working great. it's just that i'm fighting to use this KILLER plugin!

-----
* as a workaround, I could override all the relevant tsconfig.json proeprties inside this plugin's options, but I can't override include and exclude properties :(
---

Thanks again for you amazing work!!!

johnnyreilly:
Heya! Thanks for the complements! Maintainer of both ts-loader and fork-ts-checker-webpack-plugin here. It would be great to understand more about your use case. Could you go into some more details please?

Also, would you consider working on a PR?

stavalfi:
hihihi!

I'm working on a super Lerna-starter that pretty much includes everything there is to include. I'm not even close to finishing but anyways, here is a taste: https://github.com/stavalfi/lerna-starter  (just run `yarn` and you are good to go).

as part of my work, I found out that a lot of the sub packages (workspaces in yarn) has 90% of their build strategy in common. for example, packages that generate js to be used by other packages, like react/angular/... and packages that generate webapps/websites/... for each group of packages, most of the build files are the same so I moved **all** of them to a dedicated sub-package under "build". you can look it out your self, it's working great with all the 3rd party packages that I'm using like webpack, eslint and more. 

actually, I lied... I moved all of them except a single build file: "tsconfig.json". I need that in each package from any type for 2 reasons: webstorm-typescript integration AND this plugin.

for the first reason, I can use a tsconfig.json that is in the main folder (where .git folder is). for the second reason, I need your help so I can move it from there and I will get a complete 100% free-build-files packages.

johnnyreilly:
Sorry I think I missed this reply! This sounds like a potentially interesting feature; do you want to submit a PR?

stavalfi:
I have no idea where to start lol.. let me investigate a little bit..

u8sand:
I would also be interested in this capability--my use case is building opinionated reusable webpack build scripts--tsconfig.json is defined in the build scripts project but it will be utilized to build typescript in a separate project. When context is not set, the include paths don't check in the right location (instead it checks in the same place as tsconfig.json).

[...]

PR #430, opened 2020-05-28
feat: add `context` option for typescript reporter

piotr-oles:
Add the `context` option to support the usage of external tsconfig.json files.

✅ Closes: #297

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 lib
 
 # Package archive used by e2e tests
-fork-ts-checker-webpack-plugin.tgz
+fork-ts-checker-webpack-plugin-0.0.0-semantic-release.tgz
 
 # Coverage directory used by tools like istanbul
 coverage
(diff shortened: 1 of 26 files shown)


johnnyreilly (approved):

shayke:
Thanks for the quick fix!

I've tested this and it works great, one thing that is still different, i get:
`TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.`

Now, my setup is such that I potentially use webpack without ts files at all (only scss for example).
I understand that `tsc -p .` on a project with no ts files gives this error but when using `ts-loader` alone (`transpileOnly: false`) I don't get such an error and webpack builds successfully.

Any ideas? (other than disabling this plugin if no ts files, I'd rather it silently continue like ts-loader with no errors maybe?).

Thanks!

piotr-oles:
If you want to skip this error, you can configure it in the plugin:
```js
    issues: {
      exclude: [{ code: 'TS18003' }]
    }
  })
```

PR #436, opened 2020-06-03
feat: rename tsconfig option to configFile for consistency

piotr-oles:
As [ts-loader](https://github.com/TypeStrong/ts-loader) and [tsconfig-paths-webpack-plugin](https://github.com/dividab/tsconfig-paths-webpack-plugin) uses `configFile` option,
the plugin should use the same for consistency.

BREAKING CHANGE: 🧨 `typescript.tsconfig` option renamed to the `tsconfig.configFile`


johnnyreilly (approved):
Nice work 👌!

PR #440, opened 2020-06-05
feat: overwrite other fields in tsconfig.json

piotr-oles:
As we have full control over the configFile content, we can easily overwrite other fields like `include` or `exclude`. It can be useful, for example, if you don't want to type-check storybook files in your webpack compilation. Now you don't have to create a separate tsconfig.json file to accomplish that.

BREAKING CHANGE: 🧨 `typescript.compilerOptions` option changed to the `typescript.configOverwrite.compilerOptions` option


johnnyreilly (approved):
Nice work!

PR #442, opened 2020-06-08
feat: add issue.scope option

piotr-oles:
The plugin uses TypeScript settings regarding files that are included in the compilation. It can be a different set of files comparing to the webpack compilation. To not display issues that are related to files outside webpack compilation, an `issue.scope` option has been added.

BREAKING CHANGE: 🧨 Issues outside webpack compilation will not be reported by default. See
`issue.scope` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ Options for the issues filtering (`issues` option object).
 | --------- | --------------------------------- | ------------- | ----------- |
 | `include` | `object` or `function` or `array` | `undefined`   | If `object`, defines issue properties that should be [matched](./src/issue/IssueMatch.ts). If `function`, acts as a predicate where `issue` is an argument. |
 | `exclude` | `object` or `function` or `array` | `undefined`   | Same as `include` but issues that match this predicate will be excluded. |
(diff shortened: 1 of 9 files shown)


piotr-oles:
@johnnyreilly would you have time to review this?

README.md
```diff
@@ -195,6 +195,7 @@ Options for the issues filtering (`issues` option object).
 | --------- | --------------------------------- | ------------- | ----------- |
 | `include` | `object` or `function` or `array` | `undefined`   | If `object`, defines issue properties that should be [matched](./src/issue/IssueMatch.ts). If `function`, acts as a predicate where `issue` is an argument. |
 | `exclude` | `object` or `function` or `array` | `undefined`   | Same as `include` but issues that match this predicate will be excluded. |
+| `scope`   | `'all'` or `'webpack'`            | `'webpack'`   | Defines issues scope to be reported. If `'webpack'`, reports errors only related to the webpack compilation. Reports all errors otherwise. |
```
johnnyreilly: Could be worth being specific about what "all errors" means in this context? I imagine this is what `tsc` would do. Is that right?

johnnyreilly (approved):

README.md
piotr-oles: Yes, it works like `tsc` and `eslint` command when the scope is `all`.  I've updated the README.md :)

PR #443, opened 2020-06-10
fix: always pass compilation to issues hook

arthurdandrea:
Currently on a watch run of webpack this plugin is not passing the compilation argument to the issues hook. This PR fixes that.

Related: the type of the issues hook explicitly defines compilation a possible undefined, maybe if this is approved it could be changed as well.


piotr-oles:
Thank you for the PR, looks good :)

PR #445, opened 2020-06-12
Release 5.0.0

piotr-oles:
As there are no new issues reported to the beta release, I think we can release it as a stable version. I was testing it also on my private projects and seems to work well :)


johnnyreilly:
This is the big one!

johnnyreilly (approved):

PR #454, opened 2020-06-18
fix: use typescript watcher for directories to prevent TS6307

piotr-oles:
Webpack watcher notifies only about changes of files that are in the compilation dependencies graph. This means that it will not notice if we create a new file that is not missing. This is a difference between webpack and typescript watch mechanism that can lead to errors like
TS6307. This fix uses typescript watcher for directories.

### DIFF
--- a/src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
+++ b/src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
@@ -6,9 +6,10 @@ import { createRealFileSystem } from '../file-system/RealFileSystem';
 
 interface ControlledTypeScriptSystem extends ts.System {
   // control watcher
+  invokeFileCreated(path: string): void;
   invokeFileChanged(path: string): void;
   invokeFileDeleted(path: string): void;
-  invokeQueuedChanged(): void;
+  pullAndInvokeCreatedOrDeleted(): void;
   // control cache
   clearCache(): void;
   // mark these methods as defined - not optional
@@ -236,6 +237,14 @@ function createControlledTypeScriptSystem(
         await new Promise((resolve) => setImmediate(resolve));
       }
     },
+    invokeFileCreated(path: string) {
+      const normalizedPath = realFileSystem.normalizePath(path);
+
+      invokeFileWatchers(path, ts.FileWatcherEventKind.Created);
+      invokeDirectoryWatchers(normalizedPath);
+
+      deletedFiles.set(normalizedPath, false);
+    },
     invokeFileChanged(path: string) {
       const normalizedPath = realFileSystem.normalizePath(path);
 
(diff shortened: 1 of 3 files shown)


piotr-oles:
@johnnyreilly Please review :)

src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
```diff
@@ -97,24 +102,42 @@ function createControlledTypeScriptSystem(
...
+            recursiveDirectoryWatcherCallback(forwardSlash(normalizedPath))
+          );
+        }
       }
-    });
+    );
+  }
+
+  function updateDirectorySnapshot(path: string, recursive = false) {
+    const dirents = passiveFileSystem.readDir(path);
```
johnnyreilly, line 128: What is a `dirent`? Is that directory?

johnnyreilly (dismissed):

johnnyreilly:
> This means that it will not notice if we create a new file that is not missing

I didn't quite understand what this means - could you provide an example please?

piotr-oles:
Let say we have the following files:
```typescript
// src/index.ts
import message from './message';

console.log(message);
```
```typescript
// src/message.ts
const message = "Hello world";

export default message;
```
Let's run `webpack-dev-server`. The dependency graph will be:
```
src/index.ts <- src/message.ts
```
Then let's create a new file:
```typescript
// src/toUpperCase.ts
function toUpperCase(message: string) {
  return message.toUpperCase();
}

export default toUpperCase;
```
webpack-dev-server will not re-run compilation, because `toUpperCase.ts` is not in the dependency graph. Also, we don't run watcher on the typescript reporter side, because we rely on the changes reported by the webpack (see `ControlledTypeScriptSystem.ts`)

Then, let's edit `src/index.ts`:
```typescript
import message from './message';
import toUpperCase from './toUpperCase';

console.log(toUpperCase(message));
```

Webpack will notify the typescript reporter that `src/index.ts` has been changed. Unfortunately, it will not notify, that `src/toUpperCase.ts` has been created. And that's a problem because typescript assumes that it will be notified if such a file will be created (it calls `watchDirectory('src', callback, true)`, where true means `recursive` - so it watches for any change in the `src` and nested directories)

To overcome this issue, we poll directories on each compilation and compare them to detect if any other file has been created or deleted (that's how polling watcher works except we don't have fixed interval)

src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
piotr-oles: It's a name from the Node.js API (there is a `Dirent` type) which seems to be borrowed from the POSIX API. It's a structure that represents directory children (can be another directory, file, device, etc.)

PR #471, opened 2020-07-01
fix: include .vue files as the project root files

### LINKED ISSUE #452: Vue SFC support not working
## Current behavior
`createTypeScriptReporter` from `TypeScriptReporter`, which seems to be the only file that calls `createTypeScriptVueExtension` (https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/typescript-reporter/reporter/TypeScriptReporter.ts#L37), which itself should provide the `.vue` (Vue SFC) support, **is never imported or called**.

I think ForkTsChecker uses a `TypeScriptRpcClient` reporter over the one with Vue SFC support.

## Expected behavior
`createTypeScriptVueExtension` is called and provides `.vue` (Vue SFC) support.

- Setup a Vue project with README `Vue` section instructions
- Intentionally have TS errors in `.vue` files (wrong types ex: `const x: string = 2`)
- Run a project which uses ForkTsChecker

## Environment
 - **fork-ts-checker-webpack-plugin**: 5.0.1
 - **typescript**: 3.8.3 
 - **eslint**: 7.1.0
 - **webpack**: 4.43.0
 - **os**: Ubuntu 20.04

piotr-oles:
✅ Closes: #452

### DIFF
--- a/src/typescript-reporter/extension/TypeScriptEmbeddedExtension.ts
+++ b/src/typescript-reporter/extension/TypeScriptEmbeddedExtension.ts
@@ -130,6 +130,40 @@ function createTypeScriptEmbeddedExtension({
         fileExists: createEmbeddedFileExists(host.fileExists),
       };
     },
+    extendParseConfigFileHost<THost extends ts.ParseConfigFileHost>(host: THost): THost {
+      return {
+        ...host,
+        readDirectory(
+          rootDir: string,
+          extensions: readonly string[],
+          excludes: readonly string[] | undefined,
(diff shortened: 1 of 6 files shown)


piotr-oles:
The issue is that this plugin doesn't add `.vue` files to the root files of the project. So in order for it to work, you need `.ts` or `.tsx` entry file which imports `.vue` file. This PR fixes it by extending config parse host to include `.vue` files.

johnnyreilly (approved):

PR #472, opened 2020-07-04
Accept custom logger object

tessro:
The runtime validation of the options schema in v5 doesn't seem to like custom logger objects. I believe these are still intended to work, since they're accepted by the TypeScript interface and are invoked correctly.

This PR updates the schema to accept logger objects and adds a test case. Apologies for not opening an issue in advance! Since it was a small change I figured it was quick enough to pull this together all at once. :)

### DIFF
--- a/src/ForkTsCheckerWebpackPluginOptions.json
+++ b/src/ForkTsCheckerWebpackPluginOptions.json
@@ -94,7 +94,18 @@
       "enum": ["console", "webpack-infrastructure", "silent"]
     },
     "Logger": {
-      "instanceof": "Function"
+      "type": "object",
+      "properties": {
+        "error": {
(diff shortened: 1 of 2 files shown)


tessro:
Not sure why CI failed, looks like an unrelated timeout? Tests passed on my machine. :( If there's something I can do to debug, let me know!

piotr-oles (approved):

piotr-oles:
The e2e tests are not 100% deterministic. Sometimes it's a problem on the GitHub Actions (especially for MacOS), sometimes it's a race condition. I think we will never have deterministic e2e tests, but we should improve stability because currently, e2e tests fail very often.

tessro:
Ah, that makes sense. Thanks!

PR #478, opened 2020-07-09
feat: add typescriptPath option

### LINKED ISSUE #474: Add "typescript" option from v4 back in some form
## Feature motivation
Hi, In v4 there was a "typescript" option which did

> typescript string: If supplied this is a custom path where typescript can be found. Defaults to require.resolve('typescript').

This option was especially useful in the case of multi-module projects or projects that bundle webpack and its configuration in a separate dependency for reusability (I'm the maintainer of https://github.com/swissquote/crafty, one of such tools).

Would you accept a Pull Request that adds this feature back ?

## Feature description

Add an option `typescriptPath` ( I'm open to propositions for the name :) )
That defaults to `require.resolve("typescript")` and allows to be changed with the path to another TypeScript module if needed.

## Feature implementation

* Add option to schema
* Load instance on configuration init
* Change all references to the import in `TypeScriptReporterConfiguration`, `TypeScriptIssueFactory`, `TypeScriptPerformance`, `ControlledTypeScriptSystem`, `ControlledWatchCompilerHost`, `TypeScriptConfigurationParser` and `TypeScriptReporter` with references to the instance loaded from the configuration

onigoetz:
Fixes #474 

Hi, as per the issue I opened the other day, I would love to see the `typescript` option from V4 back.
I tried a "naive" implementation of it, I changed all references to variables/ function call to a `typescript` variable instead of `ts` and this variable is meant to be passed to the function.

I added the option `typescriptPath` that defaults to `require.resolve('typescript')` and initialized TypeScript in 
* `ForkTsCheckerWebpackPlugin.ts`
* `typescript-reporter/reporter/TypeScriptReporter.ts`

As I said, this is a very naive implementation, it definitely adds the argument "typescript" in a lot of functions, if needed I can refactor it to something else if you prefer another style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ Options for the TypeScript checker (`typescript` option object).
 | `diagnosticOptions` | `object`  | `{ syntactic: false, semantic: true, declaration: false, global: false }`                                      | Settings to select which diagnostics do we want to perform. |
 | `extensions`         | `object`  | `{}`                                                                                                           | See [TypeScript extensions options](#typescript-extensions-options). |
 | `profile`            | `boolean` | `false`                                                                                                        | Measures and prints timings related to the TypeScript performance. |
+| `typescriptPath`     | `string`  | `require.resolve('typescript')`                                                                                | If supplied this is a custom path where TypeScript can be found. |
 
 #### TypeScript extensions options
 
--- a/src/ForkTsCheckerWebpackPluginOptions.json
+++ b/src/ForkTsCheckerWebpackPluginOptions.json
@@ -175,6 +175,10 @@
             "profile": {
               "type": "boolean",
               "description": "Measures and prints timings related to the TypeScript performance."
+            },
+            "typescriptPath": {
+              "type": "string",
+              "description": "If supplied this is a custom path where TypeScript can be found."
             }
           }
         }
--- a/src/typescript-reporter/TypeScriptReporterConfiguration.ts
+++ b/src/typescript-reporter/TypeScriptReporterConfiguration.ts
@@ -23,6 +23,7 @@ interface TypeScriptReporterConfiguration {
     vue: TypeScriptVueExtensionConfiguration;
   };
   profile: boolean;
+  typescriptPath: string;
 }
 
 function createTypeScriptReporterConfiguration(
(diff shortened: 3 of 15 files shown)


README.md
```diff
@@ -144,6 +144,7 @@ Options passed to the plugin constructor will overwrite options from the cosmico
 | ----------------- | --------------------- | ------------------------------------------------ | ----------- |
 | `async`           | `boolean`             | `compiler.options.mode === 'development'`        | If `true`, reports issues **after** webpack's compilation is done. Thanks to that it doesn't block the compilation. Used only in the `watch` mode. | 
 | `typescript`      | `object` or `boolean` | `true`                                           | If a `boolean`, it enables/disables TypeScript checker. If an `object`, see [TypeScript options](#typescript-options). |
+| `typescriptPath`  | `string`              | `require.resolve('typescript')`                  | If supplied this is a custom path where TypeScript can be found. |
```
piotr-oles: please move typescriptPath to the typescript options and use the same interface as other typescript options :)

src/ForkTsCheckerWebpackPluginOptions.ts
```diff
@@ -7,6 +7,7 @@ import LoggerOptions from './logger/LoggerOptions';
 interface ForkTsCheckerWebpackPluginOptions {
   async?: boolean;
   typescript?: TypeScriptReporterOptions;
+  typescriptPath?: string;
```
piotr-oles: please move it to the TypeScriptReporterOptions instead of top-level options :)

piotr-oles (changes requested):

README.md
onigoetz: I moved it as requested, the integration seems cleaner now :)

src/ForkTsCheckerWebpackPluginOptions.ts
onigoetz: done

piotr-oles (dismissed):

onigoetz:
Hi @piotr-oles 
Thank you for the approval,
I see that tests still don't pass and I don't exactly understand what I did wrong, what can I do to fix them ?
Also CodeClimate says I added too many arguments on `parseTypeScriptConfiguration` do you know what approach I could take to fix it ?

Thanks in advance

piotr-oles:
Don't worry about code climate - it's ok that there is more arguments in my opinion. Please try to run e2e tests locally (`yarn test:e2e`) - it seems that your changes broke "out of memory" test :)

onigoetz:
okay sure, I'll try that.
How can I see what uses memory to reduce it ?

piotr-oles:
The test actually simulates out of memory case (by spawning process with 10MB memory) to test if we are providing some feedback message to the user :)

onigoetz:
@piotr-oles Hi, I've been experimenting with this to reproduce the errors on my machine and now the tests pass fine.
I can't understand why it fails in CI

Do you have an idea ?

piotr-oles (approved):

PR #486, opened 2020-07-22
docs: update babel-loader docs for type-only modules watch

piotr-oles:
This update should help with the `babel-loader` issues described in the #36 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,11 +62,14 @@ module.exports = {
 };
 ```
 
-If you are using **TypeScript >= 3.8.0**, it's recommended to set `"importsNotUsedAsValues": "preserve"` [compiler option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) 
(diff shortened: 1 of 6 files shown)


dmarkow:
I'm using babel-typescript. This update says "If you use **TypeScript >=3.8.0**, you can fix it:" however I just tested these recommendations (setting `onlyRemoveTypeImports` to `true`) and updating a types-only file is still not triggering a build. I need to fully restart webpack to get it to recognize my updates.

piotr-oles:
Thanks for the feedback. Is this file imported somehow by another file or is it just a reference / ambient context? This is not an ultimate solution, it should fix the same cases as `"importsNotUsedAsValues": "preserve"` which doesn't mean it will fix all cases. I'm thinking about a better solution but I don't have time to implement it.

johnnyreilly (approved):

johnnyreilly:
Nice! 👍

PR #489, opened 2020-07-26
fix: use `:line:col` format in webpack errors

### LINKED ISSUE #481: TS error in file can no longer be opened at the error location (v4 -> v5)
Hi, I'm upgrading from v4 to v5, and noticed the issue location in the code frame display is different. It used to be displayed with brackets around the location, and without the range, e.g: `/customer.ts(86,16)`. But now it has a space and no brackets e.g: `/customer.ts 86:16-24`.

This is an issue for me as in iTerm I could cmd+click the file name to open it in my editor (Sublime Text), and it would open with the cursor on the line where the error occurred. With v5 however, the new formatting means only the file will open - not the location where the error happened. This is a big productivity issue for me.

Is there any way to restore some of the old behaviour or customise the output file location to display it without range and in brackets? Either in this library or something I can do to my Webpack config. Thanks!

## Current behavior

In v5, the location with range is given after a space.

```ts
ERROR in /.../customer.ts 86:16-24
TS2551: Property 'giveName' does not exist on type 'Customer'. Did you mean 'givenName'?
    84 |     }
    85 |     if (c.givenName) {
  > 86 |       return c.giveName;
       |                ^^^^^^^^
    87 |     }
    88 |     if (c.familyName) {
    89 |       return c.familyName;
```

## Expected behavior

In v4, the location is given without range, in brackets next to the file.

In order for editors like Sublime and VSCode to open at the location, I think the format needs to be `FILE_NAME(LINE,COL)` or `FILE_NAME:LINE:COL`.

```ts
ERROR in /.../customer.ts(86,16):
86:16 Property 'giveName' does not exist on type 'Customer'. Did you mean 'givenName'?
    84 |     }
    85 |     if (c.givenName) {
  > 86 |       return c.giveName;
       |                ^
    87 |     }
    88 |     if (c.familyName) {
    89 |       return c.familyName;
```

Get a TypeScript error.

None provided.

## Environment
 - **fork-ts-checker-webpack-plugin**: 5.0.7
 - **typescript**: 3.9.7
 - **eslint**: n/a
 - **webpack**: 4.41.6
 - **os**: MacOS

robcrocombe:
Fixes #481

As discussed in #481, changes the format of TypeScript error file locations to use `file:line:col`. This means files can be opened from the terminal at the location the error occurred (e.g. cmd+click on Mac).  Like this:

```ts
ERROR in /Users/.../customer.ts:86:16
TS2551: Property 'giveName' does not exist on type 'Customer'. Did you mean 'givenName'?
    84 |     }
    85 |     if (c.givenName) {
  > 86 |       return c.giveName;
       |                ^^^^^^^^
    87 |     }
    88 |     if (c.familyName) {
    89 |       return c.familyName;
```

The location end is still used in the code frame.

### DIFF
--- a/test/e2e/EsLint.spec.ts
+++ b/test/e2e/EsLint.spec.ts
@@ -6,6 +6,7 @@ import {
   WEBPACK_CLI_VERSION,
   WEBPACK_DEV_SERVER_VERSION,
 } from './sandbox/WebpackDevServerDriver';
+import { fileLocation } from './fileLocationHelper';
 
 describe('EsLint', () => {
   let sandbox: Sandbox;
@@ -59,7 +60,7 @@ describe('EsLint', () => {
     errors = await driver.waitForErrors();
     expect(errors).toEqual([
       [
-        'WARNING in src/authenticate.ts 14:34-37',
+        fileLocation('WARNING in src/authenticate.ts', '14:34', async),
         '@typescript-eslint/no-explicit-any: Unexpected any. Specify a different type.',
         '    12 | }',
         '    13 | ',
@@ -70,7 +71,7 @@ describe('EsLint', () => {
         '    17 |     {',
       ].join('\n'),
       [
-        'WARNING in src/index.ts 31:44-49',
+        fileLocation('WARNING in src/index.ts', '31:44', async),
         "@typescript-eslint/no-unused-vars: 'event' is defined but never used.",
         '    29 |   }',
         '    30 | });',
@@ -126,7 +127,7 @@ describe('EsLint', () => {
     errors = await driver.waitForErrors();
     expect(errors).toEqual([
       [
-        'WARNING in src/model/User.ts 11:5-19',
+        fileLocation('WARNING in src/model/User.ts', '11:5', async),
         "@typescript-eslint/no-unused-vars: 'temporary' is defined but never used.",
         '     9 | }',
         '    10 | ',
@@ -137,7 +138,7 @@ describe('EsLint', () => {
         '    14 | function getUserName(user: User): string {',
       ].join('\n'),
       [
-        'WARNING in src/model/User.ts 11:16-19',
+        fileLocation('WARNING in src/model/User.ts', '11:16', async),
         '@typescript-eslint/no-explicit-any: Unexpected any. Specify a different type.',
         '     9 | }',
         '    10 | ',
--- a/test/e2e/EsLint.spec.ts
+++ b/test/e2e/EsLint.spec.ts
@@ -6,7 +6,6 @@ import {
   WEBPACK_CLI_VERSION,
   WEBPACK_DEV_SERVER_VERSION,
 } from './sandbox/WebpackDevServerDriver';
-import { fileLocation } from './fileLocationHelper';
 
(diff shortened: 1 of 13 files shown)


robcrocombe:
The E2E tests are broken, I will take a look when I can.

johnnyreilly:
Hey @robcrocombe

I just noticed you're a fellow Londoner. Hello fellow Londoner! Whereabouts do you work when you're in the office? I'm pretty central; though like everyone else I've been homebased since March.

robcrocombe:
Hey @johnnyreilly yep, I'm based in Old Street :) Except for now I'm also working from home.

I've updated the tests to handle both formats, since I don't think the `async: false` option is being changed.

johnnyreilly:
> Hey @johnnyreilly yep, I'm based in Old Street :) Except for now I'm also working from home.

Haha nice! I'm near St Paul's Cathedral.... Daring to consider cycling into town sometime. Just to have my own personal 28 days later experience 😉

test/e2e/WebpackProductionBuild.spec.ts
```diff
@@ -129,7 +129,7 @@ describe('Webpack Production Build', () => {
             '    4 |   const response = await fetch(',
           ].join('\n'),
           [
-            'ERROR in src/index.ts 2:29-43',
+            'ERROR in src/index.ts 2:29',
```
johnnyreilly: Should this read `src/index.ts:2:29` instead? i.e. with a colon

johnnyreilly (commented):
See my one question; looking good otherwise!

test/e2e/EsLint.spec.ts
```diff
@@ -59,7 +60,7 @@ describe('EsLint', () => {
     errors = await driver.waitForErrors();
     expect(errors).toEqual([
       [
-        'WARNING in src/authenticate.ts 14:34-37',
+        fileLocation('WARNING in src/authenticate.ts', '14:34', async),
```
piotr-oles: I don't think we should have different format depending on the async mode. Could you try to modify WebpackError.ts file to make it consistent?

piotr-oles (changes requested):
Good job! I hope we can make it consistent across async and sync mode ;)

test/e2e/EsLint.spec.ts
robcrocombe: Aaah, `IssueWebpackError.ts`, I missed that. Much better!

test/e2e/WebpackProductionBuild.spec.ts
robcrocombe: Yes, I missed out `IssueWebpackError.ts`, thanks

piotr-oles (approved):
Nice! :+1:

robcrocombe:
Thanks @piotr-oles! Some E2E tests are timing out, doesn't seem related to my change. Can we re-run them without pushing a new commit?

johnnyreilly (approved):

johnnyreilly:
I *think* I've requeued the tests....

robcrocombe:
Thanks guys, looking forward to using it :)

PR #491, opened 2020-07-29
fix(eslint): don't lint ignored files

### LINKED ISSUE #485: eslint: File ignored because of a matching ignore pattern. Use "--no-ignore" to override.
## Current behavior

Ignored files in .eslintrc.js are ignored successfully on the initial build. After that if they are changed and a watch build picks them up you get the error:

> WARNING in path/file.ts
> [unknown]: File ignored because of a matching ignore pattern. Use "--no-ignore" to override.

The files are successfully ignored still - it's just a distracting warning.

## Expected behavior

No warnings

```js
// .eslintrc.js
module.exports = {
    ...
    ignorePatterns: ['*/*.ignoreme.ts']
};
```

```js
// webpack.config.js
modules.exports = {
    plugins: [
        new ForkTSCheckerWebpackPlugin({
            eslint: {
                files: '.',
            },
        }),
    ]
}
```

Run a watched webpack build. The initial build will run with no warnings. Now update and save test.ignoreme.ts.
You see the warning referred to above.

## Environment
 - **fork-ts-checker-webpack-plugin**: 5.0.6
 - **typescript**: 3.9.2
 - **eslint**: 6.8.0
 - **webpack**: 4.43.0
 - **os**: Windows 10

## Notes

This could possibly be worked around by not including the eslint ignored files in the include patterns for `fork-ts-checker-webpack-plugin` (or at least ignoring the warnings from the file as per #450).

This seems to be the same issue as #316 which was previously fixed but seems to have been partially reverted in the recent restructuring. If I update the [eslint reporter](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/5ba683960e871d387685379d5a78943e4a5dc679/src/eslint-reporter/reporter/EsLintReporter.ts#L36) locally like so, I no longer see the error

```diff
const changedAndIncludedFiles = changedFiles.filter(
  (changedFile) =>
    includedFilesPatterns.some((includedFilesPattern) =>
      minimatch(changedFile, includedFilesPattern)
    ) &&
    (configuration.options.extensions || []).some((extension) =>
      changedFile.endsWith(extension)
-   )
+   ) &&
+  (!engine.isPathIgnored(changedFile )
);
```

Happy to drop in a PR if you think this looks reasonable.

andyrooger:
Filter out files ignored by the eslint config when linting on watched builds. This happened already for initial and non-watched builds.

This should stop the warnings every time a file ignored by eslintrc is changed that look like

> [unknown]: File ignored because of a matching ignore pattern. Use "--no-ignore" to override.

Closes #485

### DIFF
--- a/src/eslint-reporter/reporter/EsLintReporter.ts
+++ b/src/eslint-reporter/reporter/EsLintReporter.ts
@@ -40,7 +40,8 @@ function createEsLintReporter(configuration: EsLintReporterConfiguration): Repor
             ) &&
             (configuration.options.extensions || []).some((extension) =>
               changedFile.endsWith(extension)
-            )
+            ) &&
+            !engine.isPathIgnored(changedFile)
         );
 
         if (changedAndIncludedFiles.length) {


piotr-oles (approved):

piotr-oles:
Thanks a lot for this fix! :)

Issue #463, opened 2020-06-22
Slow Performance on Incremental Runs (Project References)


nathanforce:
## Current behavior
Incremental typechecks are very slow.

We have a project that leverages Project References in Typescript.  As the latest release of your plugin supports references we tried to upgrade.  We are seeing an unusable drop in performance when type-checking this code.

![image](https://user-images.githubusercontent.com/6694194/85316820-05d85300-b483-11ea-8da5-5b6ff578f06c.png)

For comparison, here is the same code change ran with `tsc --build --incremental`

![image](https://user-images.githubusercontent.com/6694194/85317017-4f28a280-b483-11ea-9be3-700f44c5e5f4.png)

Here is our config:

```
      new ForkTsCheckerWebpackPlugin({
        async: true,
        typescript: {
          mode: 'write-references',
          configFile: path.resolve(APP_ROOT, 'tsconfig.json'),
          configOverwrite: {
            compilerOptions: {
              incremental: true
            }
          },
          profile: true,
          build: true
        }
      }),
```

## Expected behavior
Incremental typechecks should be quick.

## Environment
 - **fork-ts-checker-webpack-plugin**: 5.0.0
 - **typescript**: 3.7.2
 - **eslint**: n/a
 - **webpack**: 4.39.3
 - **os**: macos

piotr-oles:
Hi,
Thanks for reporting this issue. Could you check timings using `tsc --build --incremental --watch`? This is the API we use and the performance drop is not in the initial compilation but in the next one.

nathanforce:
Sure thing.  I've attached the comparison.

`tsc --build --incremental --watch`
<img width="567" alt="Screen Shot 2020-06-23 at 2 49 46 PM" src="https://user-images.githubusercontent.com/6694194/85453006-0c3cfc80-b561-11ea-819c-afc5c3429ee3.png">

Webpack, showing both initial compilation and then after a file change.
<img width="607" alt="Screen Shot 2020-06-23 at 2 48 02 PM" src="https://user-images.githubusercontent.com/6694194/85453133-2971cb00-b561-11ea-914f-255cb7b05bf2.png">


One interesting thing I see is that `tsc` reports 1318 files while webpack reports 2111 modules.  Not sure if that mismatch indicates anything.

piotr-oles:
Is there any chance of getting a reproduction repository?

nathanforce:
[...]

PR #494, opened 2020-08-01
perf: improve solution builder performance by skipLibCheck

piotr-oles:
I did some profiling regarding #463 (thanks @nathanforce for the repo) and found that the difference between the plugin and  `tsc` is that `tsc` doesn't check standard library files in referenced projects. I implemented an optional `getParsedCommandLine` method on the watch solution builder host to automatically use `skipLibCheck` in referenced projects.


nathanforce:
🙌 I'll take a look today.  Thanks!

Issue #495, opened 2020-08-03
Errors reported in async mode trigger error overlays in v5


ark120202:
## Current behavior

Starting from the v5 this plugin [seems to add all reported issues to the `compilation.errors` list](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/83bc29c5260f9bd8180700d9d5d538612fd76692/src/hooks/tapDoneToAsyncGetIssues.ts#L76-L92) when [webpack-dev-server is used](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/83bc29c5260f9bd8180700d9d5d538612fd76692/src/hooks/interceptDoneToGetWebpackDevServerTap.ts#L13), apparently to resolve #180.

However this change has an (unintended?) side effect - it makes error overlay plugins consider reported issues as build errors and block the app view. This behavior can be observed with [`devServer.overlay`](https://webpack.js.org/configuration/dev-server/#devserveroverlay), [error-overlay-webpack-plugin](https://github.com/smooth-code/error-overlay-webpack-plugin), [react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin), and I think it might be the case with create-react-app's builtin overlay once it upgrades this plugin to v5.

## Expected behavior

Personally I find the new behavior problematic. I'd like error overlay to show errors that actually break the application, for the same reason why TypeScript's `noEmitOnError` option is `false` by default.

### Option 1

Add an option to enable (or disable?) the new behavior. This would be easy to implement and to configure, however it might be a little hacky (these issues are valid errors, so it makes sense for them to be included in `compilation.errors`?).

### Option 2

Add an option to override severity of all issues reported by the plugin. This also would be easy to implement, however it removes the visual distinction between errors and warnings.

### Option 3 

Don't change anything in the plugin, but document this behavior and document way to ignore these errors on overlay level. It sounds like the best solution conceptually, however AFAIK none of the overlays have an option to ignore certain errors, and that might make configuration more complex.

## Steps to reproduce the issue

```js
// webpack.config.js
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');

module.exports = {
  resolve: { extensions: ['.ts', '.js'] },
  module: {
    rules: [
      {
        test: /\.ts$/,
        use: [{ loader: 'ts-loader', options: { transpileOnly: true } }],
      },
    ],
  },

  devServer: {
[...]

PR #496, opened 2020-08-05
feat: add option to disable reporting to Webpack Dev Server

edhollinghurst:
Following the discussion in #495, looking to add option to enabling/disabling issues being displayed in Webpack Dev Server's error overlay.

### DIFF
--- a/src/ForkTsCheckerWebpackPluginConfiguration.ts
+++ b/src/ForkTsCheckerWebpackPluginConfiguration.ts
@@ -19,6 +19,7 @@ interface ForkTsCheckerWebpackPluginConfiguration {
   issue: IssueConfiguration;
   formatter: FormatterConfiguration;
   logger: LoggerConfiguration;
+  enableOverlay: boolean;
 }
 
 function createForkTsCheckerWebpackPluginConfiguration(
@@ -32,6 +33,7 @@ function createForkTsCheckerWebpackPluginConfiguration(
     issue: createIssueConfiguration(compiler, options.issue),
     formatter: createFormatterConfiguration(options.formatter),
     logger: createLoggerConfiguration(compiler, options.logger),
+    enableOverlay: options.enableOverlay !== false,
   };
 }
 
--- a/src/ForkTsCheckerWebpackPluginConfiguration.ts
+++ b/src/ForkTsCheckerWebpackPluginConfiguration.ts
@@ -19,7 +19,6 @@ interface ForkTsCheckerWebpackPluginConfiguration {
   issue: IssueConfiguration;
   formatter: FormatterConfiguration;
   logger: LoggerConfiguration;
-  enableOverlay: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


src/ForkTsCheckerWebpackPluginConfiguration.ts
```diff
@@ -32,6 +33,7 @@ function createForkTsCheckerWebpackPluginConfiguration(
     issue: createIssueConfiguration(compiler, options.issue),
     formatter: createFormatterConfiguration(options.formatter),
     logger: createLoggerConfiguration(compiler, options.logger),
+    enableOverlay: options.enableOverlay !== false,
```
Sayvai: You can probably just cast to a Boolean, to guarantee a boolean value is set? And wouldn't need to check specifically check for false and just take the value as-is (with casting)?

```suggestion
    enableOverlay: Boolean(options.enableOverlay),
```

src/ForkTsCheckerWebpackPluginConfiguration.ts
```diff
@@ -19,6 +19,7 @@ interface ForkTsCheckerWebpackPluginConfiguration {
   issue: IssueConfiguration;
   formatter: FormatterConfiguration;
   logger: LoggerConfiguration;
+  enableOverlay: boolean;
```
piotr-oles: Generally it looks ok, but I would put it as a part of the logger configuration - we could have `logger.devServer: boolean` option. I would not use `overlay` name because it's about reporting to the webpack-dev-server in general - the overlay is an option in webpack-dev-server :)

src/ForkTsCheckerWebpackPluginConfiguration.ts
piotr-oles: I think it's intentional - by default (for undefined) it will be true :)

src/ForkTsCheckerWebpackPluginOptions.json
```diff
@@ -20,6 +20,10 @@
     },
     "logger": {
       "$ref": "#/definitions/LoggerOptions"
+    },
+    "enableOverlay": {
+      "type": "boolean",
+      "description": "When false, plugin will not display issues in Webpack Dev Server's overlay"
```
piotr-oles: Please update README.md as well :)

piotr-oles (changes requested):
Looks good, please change option name and we are good to go

src/ForkTsCheckerWebpackPluginConfiguration.ts
edhollinghurst: :+1:

src/ForkTsCheckerWebpackPluginOptions.json
edhollinghurst: :+1:

piotr-oles (approved):
Thanks! :+1:

PR #507, opened 2020-08-30
fix: add missing peer dependencies

merceyz:
**What's the problem this PR addresses?**

`fork-ts-checker-webpack-plugin` depends on some peer dependencies without declaring them as such, this makes it rely on hoisting to place the dependencies in an accessible location which is not guaranteed to happen.

**How did you fix it?**

Declare `typescript` and `webpack` as peer dependencies and `vue-template-compiler` and `eslint` as an optional peer dependency


slavafomin:
@piotr-oles could you please look into this issue? I've stumbled upon the fact that `ForkTsCheckerWebpackPlugin` isn't listing it's peer dependencies correctly. This leads to some serious problems when incorrect version of webpack is getting resolved by package manager. Which in turn leads to typing errors (i.e. webpack ≠ webpack).

slavafomin:
# pnpm workaround

If you are using pnpm, you can temporarily fix this problem using the following `pnpmfile.js`:

```javascript
module.exports = {
  hooks: { readPackage },
};

function readPackage($package, context) {
  if ($package.name === 'fork-ts-checker-webpack-plugin') {
    $package.peerDependencies.webpack = '^5.0.0';
  }
  return $package;
}
```

Methuselah96:
@piotr-oles What's status of this PR? The [latest comment](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/497#issuecomment-679016683) was that it would produce unwanted warnings, however this is not the case since all major package managers support `peerDependenciesMeta` and these peer dependencies would be marked as optional.

I would be happy to create a new PR that specifically targets `alpha` so we get the added benefit of releasing it in a major version in case it causes issues.

piotr-oles:
I think we can merge this to the main branch if it's well supported :)

Methuselah96:
@piotr-oles Awesome! Looks like it just needs to be approved.

piotr-oles (approved):

PR #512, opened 2020-09-16
file path for webpack  base on process.cwd

lerit:
What's the problem this PR addresses?

fork-ts-checker-webpack-plugin output error file path for webpack base on complie context, not base on root.when error in vscode console,ite can opened the file because wrong relative file path.

How did you fix it?
output formatter error file path for webpack base on process.cwd()
relate issue:https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/509


piotr-oles:
We are still using context for `async: false` mode. Please take a look at https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/issue/IssueWebpackError.ts. That's done that way because, in `async: false` mode, we add errors to the compilation instead of outputing it directly to the console. Would be nice to modify one e2e test to use different path than webpack's context (I would not add new one as it already takes long time to run all tests).

lerit:
@piotr-oles 
I'm sorry I' m inexperienced in testing,I really don't know  what to do,would you please give me some help.

piotr-oles:
Sorry for the late response. In the tests/e2e directory, we have test files for end-to-end tests.
They use jest as a test runner and our custom ’sandbox’ code. 

The workflow that the sandbox code executes is:
 1. Creates a temporary directory (sandbox) and copy files described in the fixture directory.
 2. Runs ’npm install’ command
 3. Runs ’webpack-dev-server’ and checks the output of the command

What you have to do is modify one of these tests to have that case when the context is different than process.cwd :) Let me know if you have more questions :)

piotr-oles:
I'm closing this PR - will be merged from #524

PR #516, opened 2020-10-03
webpack side file watching

piotr-oles:
This PR moves all watching logic to the webpack side. 

### Motivation
This change allows us to watch for files changes in the same way as `tsc` does. It means that we can watch for "type-only" files because they become a part of the webpack's compilation (they are added to the `compilation.fileDependencies`). 

### Implementation
Each reporter (typescript, eslint) is asked to provide a list of dependencies (files and dirs) which are then watched by webpack. Unfortunately, webpack's directory watching is not very flexible and we are not able to handle the "add new file" case. The solution is hacky and experimental (see InclusiveNodeWatchFileSystem) but it seems to work. I'm not happy with the fact that we re-load SolutionBuilder when we add or remove files. It seems highly inefficient but it works for now and I don't have a better idea.

### Notes
I suspect that there will be some bugs/issues with this code as it hooks into some webpack internals. That's why I would like to publish it as alpha and test it in real life :)


johnnyreilly (approved):

johnnyreilly:
Looking amazing - a lot of failing tests though!

piotr-oles:
Yes, it's nice because it proves that they really work :D There are some quirks regarding case-sensitiveness in different operating systems and how it's handled by libraries (typescript, fsevents, webpack, chokidar...). I'm trying to make it work :D

piotr-oles:
Tests are passing, I'm going to ship this experimental version :)

PR #523, opened 2020-10-23
feat: add support for "generateTrace" tsconfig option

piotr-oles:
In response to https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/453#issuecomment-706683525
Related TypeScript's PR: https://github.com/microsoft/TypeScript/pull/40063


src/typescript-reporter/reporter/TypeScriptReporter.ts
```diff
@@ -18,6 +18,13 @@ import {
 import { createPerformance } from '../../profile/Performance';
 import { connectTypeScriptPerformance } from '../profile/TypeScriptPerformance';
 
+// write this type as it's available only in the newest TypeScript versions (^4.1.0)
```
johnnyreilly, line 21: So when typescript 4.1 ships we can lose this?

johnnyreilly (approved):

src/typescript-reporter/reporter/TypeScriptReporter.ts
piotr-oles: We would have to upgrade typescript in dev dependencies, and then probably this type will be available (if it will be in the public API)

PR #544, opened 2020-12-01
Fix NPEs where the watchFileSystem does not have optional properties

mjgp2:


piotr-oles:
Thank you for the PR! I will close it as we need the changed/removed files set for the plugin to work correctly, so this change would hide the real issue. I solved the issue in a slightly different way in #546. But again, thanks a lot for the contribution ;)

PR #547, opened 2020-12-04
fix: only invoke WDS tap when there are issues

tessro:
👋🏼 Hi! We were seeing a strange behavior in our app where it would spontaneously perform a full-page HMR about 15s after booting. I traced it to `fork-ts-checker-webpack-plugin`'s `async` mode invoking `webpack-dev-server`'s `sync` tap, which was causing an HMR even if no issues were reported.

I was able to resolve the issue by modifying `fork-ts-checker-webpack-plugin` to only invoke `webpack-dev-server`'s tap if any issues were found. This eliminates the unneeded reload. I believe this change is safe based on my testing, but this is my first time diving into the codebase so let me know if I'm thinking about this the wrong way! Thanks for considering this PR. 🍻


piotr-oles:
Thanks for the PR, looks good 👍

piotr-oles (approved):

tessro:
Hi, is there any chance this can be merged and released soon? Thanks!

piotr-oles:
Ah sorry, I was waiting for other release to finish and forgot about it :D

tessro:
No problem! Thank you so much. 🙏🏼

PR #556, opened 2021-01-05
refactor: update webpack to v5

### LINKED ISSUE #534: Update @types/webpack to ^5.0.0
## Feature motivation

I am trying to run fork-ts-checker-webpack-plugin with webpack v5, and am getting errors originating from the `hooks/pluginHooks.ts` file. Looking at that file and the project's package.json file, I see that the types are those of webpack v4, and that `@types/webpack` is indeed a dev dependency.
Does the plugin support webpack v5? If so, I believe the types need to be updated (and am willing to open a PR with the needed changes).

Note that updating the types to support webpack v5 will probably break compatibility with webpack v4. From what I've seen with other plugins, it's an either-or situation...

## Feature description

If the plugin already supports webpack v5 in theory, no functional changes will need to be made. If it isn't, well that's a whole different story, and this ticket is irrelevant.

## Feature implementation

Assuming the plugin already supports webpack v5, and from what I've seen, only changes required are using webpack v5 as a dev dependency (instead of `@types/webpack` of v4), and updating webpack types used throughout this repository.

roikoren755:
Closes #534


roikoren755:
If I'm not mistaken, one of the changes I did (having `IssueWebpackError` extend `WebpackError` from `webpack`, instead of the general `Error`) is breaking, as it requires `webpack@^5.11.0`, since it is only exported since then.

roikoren755:
Some e2e tests are still failing when running only with `webpack@^5.11.0`. It looks like the error messages printed are different, and may contain output from webpack itself?
I'm not able to run these tests locally - the script is unable to install node_modules. Any idea about this?

roikoren755:
Managed to get e2e tests to run locally, will work on fixing them ✌️

roikoren755:
* `TypeScriptWatchApi.spec.ts` - Line 329, there are 2 errors reported, not 1, the first showing info about the path to the files, which changes between test runs. I can add an `expect.stringMatchin(/.../)` with a regex to catch the error properly. In line 404 (before any further changes are made), `async: false` tests return a space less at the end of the line, while `async: true` tests succeed.
* `TypeScriptPnpSupport.spec.ts` - Same issues as in `TypeScriptWatchApi.spec.ts` - one where there are 2 errors instead of 1, with the new error printing info about path to packed files, and another where there's a difference in whitespace output in the error between `async: false` and `async: true`.

Other tests pass fine :)

piotr-oles:
Thanks for your work! This PR introduce breaking change which is dropping support for webpack 4. I think it's a little bit too early to drop it because webpack 5 has been released only 2 months ago :) I think it will be ok to drop in April/May 2021 to give users more time to migrate. @johnnyreilly what do you think?

roikoren755:
> Thanks for your work! This PR introduce breaking change which is dropping support for webpack 4. I think it's a little bit too early to drop it because webpack 5 has been released only 2 months ago :) I think it will be ok to drop in April/May 2021 to give users more time to migrate. @johnnyreilly what do you think?

No problem :)
In projects I work on, this plugin is the only webpack plugin that doesn't fully support webpack 5 (even though it is only type-wise), and is preventing us from updating to v5. This is definitely a breaking change, so maybe a 3rd major version bump in 6 months is worth it?

piotr-oles:
Did you tried `skipLibCheck: true` setting in `tsconfig.json`? Or using `webpack.config.js` instead of `webpack.config.ts` as a temporary solution?

roikoren755:
> Did you tried `skipLibCheck: true` setting in `tsconfig.json`? Or using `webpack.config.js` instead of `webpack.config.ts` as a temporary solution?

I didn't try these options, actually...
We'll stick with webpack 4 in the meantime, not rushing or anything

keeganstreet:
Hi @piotr-oles, any chance you could release this with a major version bump? This is blocking us from upgrading to Webpack 5.

piotr-oles:
Why is it blocking you from upgrading to webpack 5? Did you try setting `skipLibCheck: true`? Or just having webpack config file in JS instead of TS?

keeganstreet:
We're not willing to disable type safety for all packages because one package has broken types. But a workaround we can use in the meantime is to delete `fork-ts-checker-webpack-plugin`'s types:

package.json:

```
"scripts": {
  "prepare": "rm ./node_modules/fork-ts-checker-webpack-plugin/lib/index.d.ts"
}
```

piotr-oles:
The ‘skipLibCheck’ option is enabled by default in newer versions of TypeScript. The issue with package typing compatibility is known for a long time. Enabling this option doesn't decrease type safety in your codebase - it will still use packages typings to check if you use them correctly.

keeganstreet:
I see. Thanks for the explanation @piotr-oles

qtiki:
As a workaround for those who do not wish to set `skipLibCheck: true` you can ignore the types for this specific package by changing from:
```ts
import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin';
```
To:
```ts
```
The latter will load the module into a variable of `any` type.

**Disclaimer:** This worked for us with our compiler options, it's possible that it's not compatible with all compiler/linter options.

xarety:
Any updates with this PR?

roikoren755:
> Any updates with this PR?

I'll merge main into my branch and resolve the conflicts later today :)

piotr-oles:
Released in #638

PR #559, opened 2021-01-08
feat: add built-in support for concurrency limiting

### LINKED ISSUE #441: Limit service concurrency
## Feature motivation
I'm running this plugin in a serverless framework project. Now that I have multiple lambda functions this plugin is now taking up quite a lot of memory

## Feature description
Is there a way to limit the amount of concurrent services it runs? So maybe an option saying process only 10 services at the same time, the others can wait

piotr-oles:
Inspired by the `fork-ts-checker-webpack-plugin-limiter` plugin - thanks @vicary. This implementation uses a global pool to limit plugin concurrency. No configuration needed from the user.

✅ Closes: #441

### DIFF
--- a/src/ForkTsCheckerWebpackPlugin.ts
+++ b/src/ForkTsCheckerWebpackPlugin.ts
@@ -20,9 +20,18 @@ import { tapAfterCompileToAddDependencies } from './hooks/tapAfterCompileToAddDe
 import { tapErrorToLogMessage } from './hooks/tapErrorToLogMessage';
 import { getForkTsCheckerWebpackPluginHooks } from './hooks/pluginHooks';
 import { tapAfterEnvironmentToPatchWatching } from './hooks/tapAfterEnvironmentToPatchWatching';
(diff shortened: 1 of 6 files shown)


johnnyreilly (dismissed):

kim00425:
@piotr-oles how to use this option in webpack.config.js can you show me any example?

piotr-oles:
To change pool size (the default one is number of CPUs), you can set `ForkTsCheckerWebpackPlugin.issuesPool.size = x` and `ForkTsCheckerWebpackPlugin.dependenciesPool.size = x`. This is shared static variable, you can't set it in plugin constructor. So an example would be:
```
// webpack.config.js

ForkTsCheckerWebpackPlugin.issuesPool.size = 10;
ForkTsCheckerWebpackPlugin.dependenciesPool.size = 10;

module.exports = {
  // ...
  plugins: [
    new ForkTsCheckerWebpackPlugin()
  ]
}
```

PR #573, opened 2021-03-04
Make logging readable in Windows Command Prompt

joshuacc:
Logging text in blue is unfortunately not readable on the Windows Command Prompt with default color settings. Cyan, however, is.

Before: 

After:


piotr-oles (approved):
Thanks! :)

PR #578, opened 2021-03-15
feat: enhance the issue formatter to get a customized function

### LINKED ISSUE #564: Custom formatter
Is there a way to implement a custom formatter?

thanks!

vanpipy:
close #564


src/formatter/FormatterFactory.ts
```diff
@@ -25,17 +25,21 @@ function createFormatter<T extends ConfigurableFormatterType>(
...
+  }
 
-    default:
-      throw new Error(`Unknown "${type}" formatter. Available types are: basic, codeframe.`);
+  if (type === 'codeframe') {
+    return createCodeFrameFormatter(options);
   }
+
+  throw new Error(
+    `Unknown "${type}" formatter. Available types are: basic, codeframe or only a function.`
```
piotr-oles:
```suggestion
    `Unknown "${type}" formatter. Available types are: "basic", "codeframe" or a custom function.`
```

piotr-oles (changes requested):
Thanks for this PR - it looks good :) Please update `README.md` and `ForkTsCheckerWebpackPluginOptions.json`

src/formatter/FormatterFactory.ts
vanpipy: Update.

vanpipy:
> Thanks for this PR - it looks good :) Please update `README.md` and `ForkTsCheckerWebpackPluginOptions.json`

I am the first time to use the [json schema](http://json-schema.org/specification.html), but I found that the `ForkTsCheckerWebpackPluginOptions.json` don't need to change? Because follow the definition, 

```
formatter -> #/definitions/FormatterOptions

#/definitions/FormatterOptions:

1. #/definitions/FormatterType
2. #/definitions/ComplexFormatterPreferences
3. #/definitions/Formatter

#/definitions/Formatter -> {
  "type": "string", -> "type": ["string", "undefined", Formatter] // but the undefined and the Formatter cannnot used in the json schema
  "enum": ["basic", "codeframe"]
}
```

Is the right way to understand this? Or something I missed, please tell me and thanks :D.

---

Emmm.. I try to submit a new `ForkTsCheckerWebpackPluginOptions.json`, please review it later.

src/ForkTsCheckerWebpackPluginOptions.json
```diff
@@ -41,8 +41,15 @@
...
     "FormatterType":  {
-      "type": "string",
-      "enum": ["basic", "codeframe"]
+      "oneOf": [
+        {
+          "type": "string",
+          "enum": ["basic", "codeframe"]
+        },
+        {
+          "$ref": "#/definitions/FormatterType"
```
piotr-oles: you can use `"instanceof": "Function"` instead - it's not a standard json schema, but it's supported by schema-utils package that is used for validating plugin options: https://github.com/webpack/schema-utils#examples

vanpipy: Update. I did not get the idea of the _it's not a standard json schema_.

The _it's not a standard json schema_ points on `"$ref": "#/definitions/FormatterType"` and `"instanceof": "Function"` or just `"$ref": "#/definitions/FormatterType"`? Im interested on the usage of the `josn schema`. Thanks :D.

piotr-oles: it's related to the `"instanceof"` - there is no instanceof in json schema spec as far as I know. And it makes sense, because you can't have function or custom class in a JSON, so it's more like a JavaScript related extension of the json schema spec :)

piotr-oles (changes requested):
Thanks for applying theses changes! :) I have 1 last request - please add E2E tests for this.
You can update `TypeScriptContextOption.spec.ts` file - rename it to `TypeScriptContextAndFormatterOption.spec.ts` and update it to use custom function as a formatter :) You can use `sandbox.patch` to modify webpack config (defined in `fixtures/environment/typescript-basic.fixture` file). When it's done, we are ready to ship.

vanpipy:
Hi, @piotr-oles , I drop the change of the `ForkTsCheckerWebpackPluginOptions.json`, because the `Function` for the `Formatter` existed already at:

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/fe34ce1d746d421bbf01251307f84ae500a013ac/src/ForkTsCheckerWebpackPluginOptions.json#L27

```
    "Formatter": {
      "instanceof": "Function"
    },
```

piotr-oles (approved):
great! :)

PR #613, opened 2021-05-11
fix: avoid loading configs from node_modules dir

### LINKED ISSUE #606: files from `node_modules` getting resloved despite being in `.eslintignore`
Hi 👋

After updating to 6.2.6 which includes https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/602 I've got this error https://github.com/eslint/eslint/issues/14533. My config is: 
```
eslint: {
    files: ['**/*.{js,jsx,ts,tsx}'],
},
```

But I also have `node_modules` ignored and when running from CLI `yarn eslint --fix '**/*.{js,jsx,ts,tsx}'` it works just fine.

Do you think it can be fixed? Or maybe I can specify this blob differently to not include node_modules in the first place?

ratherblue:
✅ Closes: #606

### DIFF
--- a/src/eslint-reporter/reporter/EsLintReporter.ts
+++ b/src/eslint-reporter/reporter/EsLintReporter.ts
@@ -21,8 +21,8 @@ function createEsLintReporter(configuration: EsLintReporterConfiguration): Repor
 
   function isFileIncluded(path: string) {
     return (
-      includedGlobPatterns.some((pattern) => minimatch(path, pattern)) &&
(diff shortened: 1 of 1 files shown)


ratherblue:
I figure there are a few ways to fix this, but ignoring files in the `node_modules` directory seems like the simplest approach as those files should never be linted.

piotr-oles (approved):
Thanks!

Issue #611, opened 2021-05-11
Incorrect `webpack --watch` behavior for type-only imports when using yarn workspaces


jtbandes:
## Current behavior
Type-only imports ([`import type X`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export), or `import X` where X is only used as a type) result in incorrect `webpack --watch` behavior. When files residing in workspace packages are modified, the importing code does not always get recompiled.

This is a similar issue to https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/36 and https://github.com/TypeStrong/ts-loader/issues/1138, but that issue has been fixed and the recommendation to use `"importsNotUsedAsValues": "preserve"` has been removed in https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/516. However, issues still arise when using yarn workspaces.

It seems like enabling `"importsNotUsedAsValues": "preserve"` and avoiding `import type` might be a viable workaround for this issue, but that's not great, because `import type` is a core TS feature that should ideally be supported properly.

## Expected behavior
Modifying a file in a workspace package should cause webpack to recompile the bundle, even if it's only used via type-only imports.

## Steps to reproduce the issue
0. Check out https://github.com/jtbandes/ts-workspaces-repro. This repo uses [Yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) and has two sub-packages under `packages/pkg1` and `packages/pkg2`. The main file `src/index.ts` imports from these packages.

1. Run `yarn install && yarn webpack --watch --progress --mode=development`

2. Modify `type Example1` in [packages/pkg1/index.ts](https://github.com/jtbandes/ts-workspaces-repro/blob/main/packages/pkg1/index.ts). Observe that webpack **does not** recompile `src/index.ts` nor display new type errors. ❌

3. Modify `function Example2` in [packages/pkg2/index.ts](https://github.com/jtbandes/ts-workspaces-repro/blob/main/packages/pkg2/index.ts). Observe that webpack **does** recompile `src/index.ts` and display new type errors (this is not a type-only import). ✅

4. Add
    `"compilerOptions": { "importsNotUsedAsValues": "preserve" }` to [tsconfig.json](https://github.com/jtbandes/ts-workspaces-repro/blob/main/tsconfig.json). Now, observe that changing `pkg1/index.ts` **does** cause webpack to recompile. ✅
[...]

PR #617, opened 2021-05-17
fix: use memfs for ts artifacts to prevent mtime conflict

piotr-oles:
When ts-loader uses project references mode, it writes down .js and .d.ts files. If it does it faster than the plugin, the typescript instance in the plugin is confused and doesn't work properly. To fix that, let's use memfs for dirs and files that are detected as project artifacts.

Related: #611

### DIFF
--- a/src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
+++ b/src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
@@ -3,6 +3,8 @@ import { dirname } from 'path';
 import { createPassiveFileSystem } from '../file-system/PassiveFileSystem';
 import forwardSlash from '../../utils/path/forwardSlash';
 import { createRealFileSystem } from '../file-system/RealFileSystem';
+import { createMemFileSystem } from '../file-system/MemFileSystem';
+import { FilesMatch } from '../../reporter';
 
 interface ControlledTypeScriptSystem extends ts.System {
   // control watcher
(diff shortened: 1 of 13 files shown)


johnnyreilly (approved):

src/typescript-reporter/reporter/ControlledTypeScriptSystem.ts
```diff
@@ -34,6 +36,7 @@ interface ControlledTypeScriptSystem extends ts.System {
   clearTimeout(timeoutId: any): void;
   // detect when all tasks scheduled by `setTimeout` finished
   waitForQueued(): Promise<void>;
+  setArtifacts(dependencies: FilesMatch): void;
```
piotr-oles:
```suggestion
  setArtifacts(artifacts: FilesMatch): void;
```

src/typescript-reporter/reporter/TypeScriptReporter.ts
```diff
@@ -178,6 +178,19 @@ function createTypeScriptReporter(configuration: TypeScriptReporterConfiguration
...
+
+    const [parseConfigFileHost] = getParseConfigFileHost();
+
+    return getArtifactsFromTypeScriptConfiguration(
+      typescript,
+      parsedConfiguration,
+      configuration.context,
+      parseConfigFileHost
+    );
+  }
```
piotr-oles, line 193: maybe we should cache it to avoid parsing all config files for each compilation...

PR #632, opened 2021-07-20
fix: ensure typescript is imported only as type-only

### LINKED ISSUE #631: getTsBuildInfoEmitOutputFilePath/getParsedCommandLineOfConfigFile does not use the ts version specified by typescriptPath
## Current behavior
With the below config
image

Both the specified version of typescript (via typescriptPath) and the "default" require("typescript") version are being imported, if they differ
image

## Expected behavior
I would expect only typescript-next (as defined by my typescriptPath value) is the only version of typescript loaded

Specify some other path to typescript, open chrome inspect and note there are multiple versions. I believe this is due to two references to typescript posted below. Importing typescript in most cases is just type-only, but these two cases actually utilize ts code. Can we change these typescript imports to type-only after fixing these two cases?

image

1. https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/baa3a37f18e7764cf7b1e47f4473f7b068d194d5/src/typescript-reporter/reporter/ControlledWatchSolutionBuilderHost.ts#L62
2. https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/8d8125c94d40d16dc2cc5182416d73d15d336392/src/typescript-reporter/reporter/TypeScriptConfigurationParser.ts#L118

I think there is enough here to not require a repro repo. lmk if you need one though

## Environment
 - **fork-ts-checker-webpack-plugin**: 6.2.12
 - **typescript**: 4.3.2 / typescript-next: 4.4.0-beta
 - **eslint**: N/A
 - **webpack**: 5.45.1
 - **os**: Windows 10

markjm:
Fixes #631

### DIFF
--- a/src/typescript-reporter/reporter/TypeScriptConfigurationParser.ts
+++ b/src/typescript-reporter/reporter/TypeScriptConfigurationParser.ts
@@ -1,4 +1,4 @@
-import * as ts from 'typescript';
+import type * as ts from 'typescript';
 import { normalize, dirname, basename, resolve, relative } from 'path';
 import { TypeScriptConfigurationOverwrite } from '../TypeScriptConfigurationOverwrite';
 import { FilesMatch } from '../../reporter';
@@ -114,10 +114,10 @@ function removeJsonExtension(path: string) {
   }
 }
 
-function getTsBuildInfoEmitOutputFilePath(options: ts.CompilerOptions) {
-  if (typeof ts.getTsBuildInfoEmitOutputFilePath === 'function') {
+function getTsBuildInfoEmitOutputFilePath(typescript: typeof ts, options: ts.CompilerOptions) {
+  if (typeof typescript.getTsBuildInfoEmitOutputFilePath === 'function') {
     // old TypeScript version doesn't provides this method
-    return ts.getTsBuildInfoEmitOutputFilePath(options);
+    return typescript.getTsBuildInfoEmitOutputFilePath(options);
   }
 
   // based on the implementation from typescript
@@ -159,7 +159,10 @@ function getArtifactsFromTypeScriptConfiguration(
     if (parsedConfiguration.options.outFile) {
       files.add(resolve(configFileContext, parsedConfiguration.options.outFile));
     }
-    const tsBuildInfoPath = getTsBuildInfoEmitOutputFilePath(parsedConfiguration.options);
+    const tsBuildInfoPath = getTsBuildInfoEmitOutputFilePath(
+      typescript,
+      parsedConfiguration.options
+    );
     if (tsBuildInfoPath) {
       files.add(resolve(configFileContext, tsBuildInfoPath));
     }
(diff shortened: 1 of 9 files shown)


johnnyreilly:
This looks pretty promising!  I've approved the tests to run.

src/typescript-reporter/reporter/TypeScriptConfigurationParser.ts
```diff
@@ -114,10 +114,10 @@ function removeJsonExtension(path: string) {
   }
 }
 
-function getTsBuildInfoEmitOutputFilePath(options: ts.CompilerOptions) {
-  if (typeof ts.getTsBuildInfoEmitOutputFilePath === 'function') {
+function getTsBuildInfoEmitOutputFilePath(typescript: typeof ts, options: ts.CompilerOptions) {
+  if (typeof typescript.getTsBuildInfoEmitOutputFilePath === 'function') {
```
markjm, line 118: hm, there does seem to be some issue with getTsBuildInfoEmitOutputFilePath causing failures when using the passed in version. Seems something may be swallowing it, as this is how the issue is surfacing

```
Error: No handler found for procedure getReport.
Error: No handler found for procedure getReport.
    at Object.<anonymous> (/tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/RpcService.js:21:27)
    at Generator.next (<anonymous>)
    at /tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/RpcService.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/RpcService.js:4:12)
    at callListener (/tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/RpcService.js:16:39)
    at messageListeners.forEach (/tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:31:13)
    at Set.forEach (<anonymous>)
    at process.handleMessage (/tmp/fork-ts-checker-sandbox-7rVqZk/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:30:26)
    at process.emit (events.js:198:13)
```

Any thoughts @piotr-oles ?

A suprising result, given it seems to be a common pattern across the whole file

markjm: It almost seems like just how things are being imported is causing the memory issue to be hit a little sooner (before RPC handlers are set up)

piotr-oles: It seems like another bug that has been revealed by this change - we should close the reporter on an unsuccessful `configure` call. I will send a fix

piotr-oles:
Closing in favor of #633

Issue #634, opened 2021-07-27
`async` setting is not respected for tapAfterCompileToAddDependencies


markjm:
## Current behavior
When `async` is set to `true`, issues will be surfaced asynchronously via `tapDoneToAsyncGetIssues`. However, the _first_ recompilation gets blocked in `tapAfterCompileToAddDependencies` due to the webpack compilation waiting on `state.dependencies` promise. When TS compilation is very slow or broken (as there seemed to be in 4.3.2 for us), this causes webpack compilation to hang with no indication of what is going wrong.

## Expected behavior
When `async` is set, triggering a recompilation via file save should happen immediately, instead of being stuck behind initial TS project compilation to add dependencies.

## Steps to reproduce the issue
It is tough to create a clear repro because this only happens when repo is so big or having issues that this step takes a long time. To emulate it, you can set `state.dependenciesPromise` to just be a very long sleep promise and see that recompilations after the first can not start until that is complete.

## Proposed fix
I have attached a proposed fix patch that is working for us. I worry I may be missing some potential leak by not waiting for this promise, but it has been working will in our large repo for about a week now

```
diff --git a/node_modules/fork-ts-checker-webpack-plugin/lib/hooks/tapAfterCompileToAddDependencies.js b/node_modules/fork-ts-checker-webpack-plugin/lib/hooks/tapAfterCompileToAddDependencies.js
index f5155a2..589846a 100644
--- a/node_modules/fork-ts-checker-webpack-plugin/lib/hooks/tapAfterCompileToAddDependencies.js
+++ b/node_modules/fork-ts-checker-webpack-plugin/lib/hooks/tapAfterCompileToAddDependencies.js
@@ -10,7 +10,11 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 };
 Object.defineProperty(exports, "__esModule", { value: true });
 function tapAfterCompileToAddDependencies(compiler, configuration, state) {
-    compiler.hooks.afterCompile.tapPromise('ForkTsCheckerWebpackPlugin', (compilation) => __awaiter(this, void 0, void 0, function* () {
+    const tapFn = configuration.async
+        ? compiler.hooks.afterCompile.tap
+        : compiler.hooks.afterCompile.tapPromise;
+        
+    tapFn.call(compiler.hooks.afterCompile, 'ForkTsCheckerWebpackPlugin', (compilation) => __awaiter(this, void 0, void 0, function* () {
[...]

PR #636, opened 2021-08-01
feat: use separate workers for dependencies and issues

### LINKED ISSUE #612: ESLint checking from the 1st compilation blocks 2nd webpack compilation (tapAfterCompileToAddDependencies)
## Current behavior
1. During the initial compilation, fork-ts-checker starts doing its thing, and one of these things is `getDependencies`.
2. Dependencies resolution will be written to `dependenciesPromise`
3. Next webpack compilation will stumble upon `tapAfterCompileToAddDependencies` hook.
4. If dependencies weren't resolved yet, this compilation will be blocked until fork-ts-checker finishes its job.

- My first initial compilation takes ~22 seconds.
- Initial TS+ESLint checks take something like 2 minutes.
- Without this plugin an HMR recompilation takes 1-2 seconds.
- **Issue:** With the plugin, an HMR recompilation will have to wait for `dependenciesPromise` to resolve which means I can't use the app for the next 2 minutes while plugin is doing all the checks.

Illustration of the issue at hand, hope it's not too messy (I added a couple of logs to show how much time  `tapAfterCompileToAddDependencies` takes).

screenshot

**Note:** I can disable ESLint and this doesn't happen. `getDependencies` takes so long for ESLint specifically.

## Expected behavior
Not sure how this should be fixed, but maybe this can work?
- Don't tap into afterCompile hook.
- Wait until `getDependencies` resolves and only **after that** tap some hook and add resolved dependencies into next webpack compilation?

I couldn't really make this work locally for me, but if you can point me to the right solution, I can make a PR.

- Have a project big enough for ESLint's getDependencies to take considerable time.
- Compile with webpack and immediately do an HMR recompilation - easiest use-case is when `lazyCompilation` is enabled.
- Notice how 2nd webpack compilation will only finish when fork-ts-checker is finished.

I can try to create one if description above isn't enough.

## Environment
 - **fork-ts-checker-webpack-plugin**: 6.2.6
 - **typescript**: 4.1.2 (irrelevant I think)
 - **eslint**: 7.8.1
 - **webpack**: 5.35.0
 - **os**: win10

piotr-oles:
In order to not block the next iteration on the `getDependencies` call, we use separate worker for dependencies calculations (so `getIssues` from previous compilation will not block the next `getDependencies` call)

✅ Closes: #612, #634

It's not a perfect implementation, but the one that requires minimum code change. I will improve this in the next major version.

### DIFF
--- a/src/ForkTsCheckerWebpackPlugin.ts
+++ b/src/ForkTsCheckerWebpackPlugin.ts
@@ -29,9 +29,18 @@ class ForkTsCheckerWebpackPlugin implements webpack.Plugin {
    */
   static readonly version: string = '{{VERSION}}'; // will be replaced by the @semantic-release/exec
   /**
-   * Default pool for the plugin concurrency limit
+   * Default pools for the plugin concurrency limit
    */
-  static readonly pool: Pool = createPool(Math.max(1, os.cpus().length));
+  static readonly issuesPool: Pool = createPool(Math.max(1, os.cpus().length));
+  static readonly dependenciesPool: Pool = createPool(Math.max(1, os.cpus().length));
+
+  /**
+   * @deprecated Use ForkTsCheckerWebpackPlugin.issuesPool instead
+   */
+  static get pool(): Pool {
+    // for backward compatibility
+    return ForkTsCheckerWebpackPlugin.issuesPool;
+  }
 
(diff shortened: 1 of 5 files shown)


johnnyreilly (approved):
Nice work!

PR #638, opened 2021-08-01
feat: drop support for webpack 4

piotr-oles:
BREAKING CHANGE: 🧨 Webpack 4 is no longer supported. Please upgrade to Webpack ^5.11.0 or use an older version of the plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 ## Installation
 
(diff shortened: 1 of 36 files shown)


johnnyreilly (approved):
Nice work - I moved ts-loader to run against webpack 5 earlier in the year. It was very satisfying being able to consume the type definitions directly from webpack itself.

PR #662, opened 2021-10-06
fix: make all peer dependencies optional

Methuselah96:
This is a followup to https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/507. I'm guessing CRA would want at least the `typescript` peerDependency to be optional and maybe `webpack` as well (although I'm not sure the reason for `webpack`). I could be mistaken, but I don't want the previous PR to cause any problems for CRA.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,13 @@
     "eslint": {
       "optional": true
     },
+    "typescript": {
+      "optional": true
+    },
     "vue-template-compiler": {
+      "optional": true
+    },
+    "webpack": {
       "optional": true
     }
   },


Methuselah96:
@piotr-oles It's probably worth merging this fairly quickly so that the previous PR doesn't cause problems for CRA.

merceyz:
They are both required so marking them as optional is not correct, if `create-react-app` uses this package in a way that makes them optional it's up to them to mark their peer dependencies as such.

> It's probably worth merging this fairly quickly so that the previous PR doesn't cause problems for CRA.

Their current stable version (https://registry.yarnpkg.com/react-dev-utils/11.0.4) is locked to `4.1.6` so there is no rush

Methuselah96:
Glad to know it's locked so there's no rush. Although it looks like it will be unlocked with v5 (or at least it currently is in the main branch).

I also think it makes sense for them to be required. However, the maintainers of this package have voiced that they want to support CRA's requirement that they be optional to fit CRA's needs and therefore I just want to make sure it's understood that this is going against that requirement and avoid any unnecessary friction.

piotr-oles:
Let's wait for CRA response when they upgrade to the newest version :)

Issue #668, opened 2021-10-20
`Module has no default export` error with Vue 3 `script setup` components in v6.3.5


cexbrayat:
## Current behavior
Thank you for your work for a better support of `script setup`

When upgrading an application to use v6.3.5, we encounter the following error with Vue 3 components using `script setup`:

```
ERROR in src/App.vue:7:24
TS2306: File '/Users/ced-pro/Code/test/vue-cli-tests/no-default-export/src/components/HelloWorld.vue.ts' is not a module.
     5 |
     6 | <script setup lang="ts">
  >  7 | import HelloWorld from './components/HelloWorld.vue';
       |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     8 | </script>
     9 |
    10 | <style>

ERROR in src/main.ts:2:8
TS1192: Module '"/Users/ced-pro/Code/test/vue-cli-tests/no-default-export/src/App.vue"' has no default export.
    1 | import { createApp } from 'vue'
  > 2 | import App from './App.vue'
      |        ^^^
    3 |
    4 | createApp(App).mount('#app')
```

## Expected behavior

Previously, with v6.3.4, there was no error.

## Steps to reproduce the issue

This occurs in new projects using Vue CLI v5, with Vue 3 and TypeScript, blocking new CLI users.
```
npx -p @vue/cli@next vue create no-default-export --inlinePreset '{"useConfigFiles": true,"plugins": {"@vue/cli-plugin-typescript": {"classComponent": false}},"vueVersion": "3"}'
cd no-default-export
# switch components to `script setup`
yarn serve
```

## Issue reproduction repository
A simple reproduction is available at https://github.com/cexbrayat/no-default-export
```
yarn
yarn serve
```

If you change the resolution in package.json to use v6.3.4, then `yarn serve` works again.

## Environment
 - **fork-ts-checker-webpack-plugin**: 6.3.5
 - **typescript**: 4.1.5 (same issue with 4.3 and 4.4)
 - **webpack**: 5.59.0
 - **os**: MacOS

cexbrayat:
Reading through commit https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/commit/2cfe45f567e30a56dea4cb5907a0d47e62dae644, I realize that your example of script setup component is inaccurate: `script setup` component have a completely different syntax, with no export.
[...]

Issue #665, opened 2021-10-18
Not support Vue3 setup-script block


curly210102:
## Current behavior

Config fork-ts-checker-webpack-plugin with 
``` js
{
  typescript: {
          extensions: {
            vue: {
              enabled: true,
              compiler: isVue3
                ? require.resolve('@vue/compiler-sfc')
                : require.resolve('vue-template-compiler')
            }
          },
          diagnosticOptions: {
            semantic: true,
            // https://github.com/TypeStrong/ts-loader#happypackmode
            syntactic: useThreads
          }
      }
   }
}
```
the ts-checker is not working with vue setup-script block.

``` vue
<script setup lang="ts">
const b:string = 1;
</script>
```

## Expected behavior

ts-checker works with setup-script block, also works when <script setup> alongside normal <script>
Reference to Vue3 document <https://v3.vuejs.org/api/sfc-script-setup.html>

## Steps to reproduce the issue

https://github.com/vuejs/vue-cli/issues/6716
https://github.com/vuejs/vue-cli/issues/6736

``` bash
npm install -g @vue/cli
vue create my-vue-app
```
Mannually select features: add TypeScript

``` bash
echo '<script setup lang="ts">const x: string = 42;</script>' >> src/App.vue
```
run Build

```
npm run build
```

## Issue reproduction repository

https://github.com/curly210102/webpack-typescript-vue/tree/main

``` bash
yarn
yarn build
```

[Open in StackBlitz](https://stackblitz.com/github/curly210102/webpack-typescript-vue/tree/main)

## Issue Tracking

@vue/compiler-sfc parse methods return result
``` js
export interface SFCDescriptor {
  filename: string
  source: string
  template: SFCTemplateBlock | null
  script: SFCScriptBlock | null
  scriptSetup: SFCScriptBlock | null
  styles: SFCStyleBlock[]
  customBlocks: SFCBlock[]
  cssVars: string[]
  // whether the SFC uses :slotted() modifier.
  // this is used as a compiler optimization hint.
  slotted: boolean
}
```
The ts-checker only handle 'script' property, ignored 'scriptSetup'.
https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/ad466dfd6c124fb624c3ce409c8fba4300a33255/src/typescript-reporter/extension/vue/TypeScriptVueExtension.ts#L112-L127

## Environment
- "fork-ts-checker-webpack-plugin": "^6.3.4",
[...]

PR #676, opened 2021-11-21
Support <script setup /> in vue 3

piotr-oles:
In Vue 3, when <script> tag has "setup" attribute, SFC parser assignes script block to a separate field called "scriptSetup"

It's a second attempt to handle this syntax, the previous one was reverted due to incomplete feature support (#668). 

See #668, #665

### DIFF
--- a/src/typescript-reporter/extension/vue/TypeScriptVueExtension.ts
+++ b/src/typescript-reporter/extension/vue/TypeScriptVueExtension.ts
@@ -10,8 +10,7 @@ import { VueTemplateCompilerV3 } from './types/vue__compiler-sfc';
 
 interface GenericScriptSFCBlock {
   content: string;
-  attrs: Record<string, string | true>;
-  lang?: string;
+  attrs: Record<string, string | true | undefined>;
   src?: string;
 }
 
@@ -92,6 +91,22 @@ function createTypeScriptVueExtension(
     };
   }
 
+  function mergeVueScriptsContent(
+    scriptContent: string | undefined,
+    scriptSetupContent: string | undefined
+  ): string {
+    const scriptLines = scriptContent?.split(/\r?\n/) ?? [];
+    const scriptSetupLines = scriptSetupContent?.split(/\r?\n/) ?? [];
+    const maxScriptLines = Math.max(scriptLines.length, scriptSetupLines.length);
+    const mergedScriptLines: string[] = [];
+
+    for (let line = 0; line < maxScriptLines; ++line) {
+      mergedScriptLines.push(scriptLines[line] || scriptSetupLines[line]);
+    }
+
+    return mergedScriptLines.join('\n');
+  }
+
   function getVueEmbeddedSource(fileName: string): TypeScriptEmbeddedSource | undefined {
     if (!fs.existsSync(fileName)) {
       return undefined;
@@ -113,26 +128,32 @@ function createTypeScriptVueExtension(
       });
 
       if (parsed.descriptor) {
-        const scriptV3 = parsed.descriptor.scriptSetup || parsed.descriptor.script;
-        const isScriptSetup = Boolean(parsed.descriptor.scriptSetup);
+        const parsedScript = parsed.descriptor.script;
+        const parsedScriptSetup = parsed.descriptor.scriptSetup;
+        let parsedContent = mergeVueScriptsContent(
+          parsedScript?.content,
+          parsedScriptSetup?.content
+        );
 
-        if (scriptV3) {
-          const scriptV3Content =
-            isScriptSetup && scriptV3.content
-              ? scriptV3.content +
-                [
-                  'export default {};',
-                  // fallback to "any" in case we can't import from '@vue/runtime-core';
-                  '// @ts-ignore',
-                  "import { defineProps, defineEmits, defineExpose, withDefaults } from '@vue/runtime-core';",
-                ].join('\n')
-              : scriptV3.content;
+        if (parsedScriptSetup) {
+          // a little bit naive, but should work in 99.9% cases without need for parsing script
+          const alreadyHasExportDefault = /export\s+default[\s|{]/gm.test(parsedContent);
+
+          if (!alreadyHasExportDefault) {
+            parsedContent += '\nexport default {};';
+          }
+          // add script setup lines at the end
+          parsedContent +=
+            "\n// @ts-ignore\nimport { defineProps, defineEmits, defineExpose, withDefaults } from '@vue/runtime-core';";
+        }
+
+        if (parsedScript || parsedScriptSetup) {
           // map newer version of SFCScriptBlock to the generic one
           script = {
-            content: scriptV3Content,
-            attrs: scriptV3.attrs,
-            lang: scriptV3.lang,
-            src: scriptV3.src,
+            content: parsedContent,
+            attrs: {
+              lang: parsedScript?.lang || parsedScriptSetup?.lang,
+            },
           };
         }
       }
--- a/src/typescript-reporter/extension/vue/TypeScriptVueExtension.ts
+++ b/src/typescript-reporter/extension/vue/TypeScriptVueExtension.ts
@@ -11,8 +11,6 @@ import { VueTemplateCompilerV3 } from './types/vue__compiler-sfc';
 interface GenericScriptSFCBlock {
   content: string;
   attrs: Record<string, string | true>;
-  start?: number;
-  end?: number;
   lang?: string;
   src?: string;
 }
@@ -105,25 +103,38 @@ function createTypeScriptVueExtension(
     let script: GenericScriptSFCBlock | undefined;
     if (isVueTemplateCompilerV2(compiler)) {
       const parsed = compiler.parseComponent(vueSourceText, {
-        pad: 'space',
+        pad: 'line',
       });
 
       script = parsed.script;
     } else if (isVueTemplateCompilerV3(compiler)) {
-      const parsed = compiler.parse(vueSourceText);
+      const parsed = compiler.parse(vueSourceText, {
+        pad: 'line',
+      });
 
-      if (parsed.descriptor && parsed.descriptor.script) {
-        const scriptV3 = parsed.descriptor.script;
+      if (parsed.descriptor) {
+        const scriptV3 = parsed.descriptor.scriptSetup || parsed.descriptor.script;
+        const isScriptSetup = Boolean(parsed.descriptor.scriptSetup);
 
-        // map newer version of SFCScriptBlock to the generic one
-        script = {
-          content: scriptV3.content,
-          attrs: scriptV3.attrs,
-          start: scriptV3.loc.start.offset,
-          end: scriptV3.loc.end.offset,
-          lang: scriptV3.lang,
-          src: scriptV3.src,
-        };
+        if (scriptV3) {
+          const scriptV3Content =
+            isScriptSetup && scriptV3.content
+              ? scriptV3.content +
+                [
+                  'export default {};',
+                  // fallback to "any" in case we can't import from '@vue/runtime-core';
+                  '// @ts-ignore',
+                  "import { defineProps, defineEmits, defineExpose, withDefaults } from '@vue/runtime-core';",
+                ].join('\n')
+              : scriptV3.content;
+          // map newer version of SFCScriptBlock to the generic one
+          script = {
+            content: scriptV3Content,
+            attrs: scriptV3.attrs,
+            lang: scriptV3.lang,
+            src: scriptV3.src,
+          };
+        }
       }
     } else {
       throw new Error(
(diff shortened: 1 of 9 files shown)


filimon-danopoulos-stratsys:
I am not an expert but I need these updates so I gave it a shot at testing it locally. It seems that it does not recognize that I have `@vue/compiler-sfc` installed and throws. To get around this i just add `@vue/compiler-sfc` locally in `fork-ts-checker-webpack-plugin` (make sure the version matches the one in you project). I think this issue has to do with how `yarn link` works. 

The good news is that it now checks `<script setup lang="ts">`. I have tested the following two components: 

Test1.vue
```html
<script lang="ts">
import { defineComponent } from 'vue'
const test: boolean = "test"
export default defineComponent({
  data() {
    return {
      test
    }
  }
})
</script>
<template>
  <div>
    {{ test }}
  </div>
</template>
```

Test2.vue
```html
<script setup lang="ts">
const test: boolean = "true"
const props = defineProps({
  prop1: String,
})
</script>

<template>
  <div>
    {{ test }}
  </div>
</template>
 ```
 
 The bad news is that regular components are no longer handled correctly. `Test2.vue` is clearly typechecked  but `defineComponent` seems to not handle `data()` correctly but it did so previously.
 
 ```bash
 ERROR in src/Test1.vue:5:7
TS2322: Type '"test"' is not assignable to type 'boolean'.
    3 | <script lang="ts">
    4 | import { defineComponent } from 'vue'
  > 5 | const test: boolean = "test"
      |       ^^^^
    6 | export default defineComponent({
    7 |   data() {
    8 |     return {

ERROR in src/Test1.vue:7:3
TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '{ data(): { test: boolean; }; }' is not assignable to parameter ofown, unknown, {}, {}, ComponentOptionsMixin, ... 4 more ..., any>'.
      Object literal may only specify known properties, and 'data' does not exist in typ unknown, {}, {}, ComponentOptionsMixin, ... 4 more ..., any>'.
     5 | const test: boolean = "test"
     6 | export default defineComponent({
  >  7 |   data() {
       |   ^^^^
     8 |     return {
     9 |       test
    10 |     }

ERROR in src/Test2.vue:3:7
TS2322: Type '"true"' is not assignable to type 'boolean'.
    1 | 
    2 | <script setup lang="ts">
  > 3 | const test: boolean = "true"
      |       ^^^^
    4 | const props = defineProps({
    5 |   prop1: String,
    6 | })

webpack 5.64.1 compiled with 3 errors in 1249 ms
```

Compared to the output of `6.4.0` where `Test2.vue` is absent and `defineComponent` is handled correctly. 

```bash
ERROR in src/Test1.vue:5:7
    3 | <script lang="ts">
    4 | import { defineComponent } from 'vue'
  > 5 | const test: boolean = "test"
      |       ^^^^
    6 | export default defineComponent({
    7 |   data() {
    8 |     return {

webpack 5.64.1 compiled with 1 error in 1460 ms
```
**Edit** 
For the sake of completeness I removed `const test: boolean = "test"` for `const test: boolean = false` in `Test1.vue` and still get the same error

```bash
ERROR in src/Test1.vue:7:3
TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '{ data(): { test: boolean; }; }' is not assignable to parameter of type 'ComponentOptionsWithObjectProps<readonly string[] | Readonly<ComponentObjectPropsOptions<Record<string, unknown>>>, unknown, unknown, {}, {}, ComponentOptionsMixin, ... 4 more ..., any>'.
      Object literal may only specify known properties, and 'data' does not exist in type 'ComponentOptionsWithObjectProps<readonly string[] | Readonly<ComponentObjectPropsOptions<Record<string, unknown>>>, unknown, unknown, {}, {}, ComponentOptionsMixin, ... 4 more ..., any>'.
     5 | const test: boolean =  false
     6 | export default defineComponent({
  >  7 |   data() {
       |   ^^^^
     8 |     return {
     9 |       test
    10 |     }

ERROR in src/Test2.vue:3:7
TS2322: Type '"true"' is not assignable to type 'boolean'.
    1 | 
    2 | <script setup lang="ts">
  > 3 | const test: boolean = "true"
      |       ^^^^
    4 | const props = defineProps({
    5 |   prop1: String,
    6 | })

webpack 5.64.1 compiled with 2 errors in 1269 ms
```

piotr-oles:
@filimon-danopoulos-stratsys thanks for testing! 

I will add a test-case with `defineComponent`. You can try to run `npm pack` instead of `yarn link` in the plugin directory, and then `yarn add <path-to-plugin.tgz-file>` in the project directory - this should fix the issue with dependencies that you have with `yarn link` (and maybe it will fix the issue altogether - maybe there is some typings versions conflict 🤔 )

bodograumann:
I had to use `nlc` from `npm-link-better` to see some reasonable results. `yarn pack` and `npm install ….tgz` didn’t work, because it is actually installed as a sub-dependency via `@vue/cli-plugin-typescript`.

Luckily I don’t have any legacy component, but I do have the situation where I have both a `<script setup lang="ts">` as well as a regular `<script lang="ts">` block. E.g. besides the component there is:
```
<script lang="ts">
export interface ScrollAlignOptions {
…
}
</script>
```

Now in another component I try to import that interface, but get the error:
> TS2614: Module '"@/components/generic/ScrollContainer.vue"' has no exported member 'ScrollAlignOptions'. Did you mean to use 'import ScrollAlignOptions from "@/components/generic/ScrollContainer.vue"' instead?

filimon-danopoulos-stratsys:
@piotr-oles Packing and installing locally fixes the issue I encountered. Just another case of `yarn link` not really doing what you expect. Great work!

filimon-danopoulos-stratsys:
@bodograumann I can verify that this is indeed an issue. In addition, no type checking is performed in `<script lang="ts">` if a `.vue` file includes `<script setup lang="ts">`. I expanded my little test suite with 

Test3.vue
```html
<script setup lang="ts">
const test: boolean = 'test'
</script>

<script lang="ts">
const test: boolean = 'test'
export interface TestInterface {
  prop1: string
}
</script>

<template>
  <div>
    {{ test }}
  </div>
</template>
```

I simply import it and use it for `defineProps<TestInterface>()` in `Test2.vue` and I get the following output. 

```bash
ERROR in src/Test1.vue:5:7
TS2322: Type 'string' is not assignable to type 'boolean'.
    3 | <script lang="ts">
    4 | import { defineComponent } from 'vue'
  > 5 | const test: boolean = "test"
      |       ^^^^
    6 | export default defineComponent({
    7 |   data() {
    8 |     return {

ERROR in src/Test2.vue:3:10
TS2614: Module '"./Test3.vue"' has no exported member 'TestInterface'. Did you mean to use 'import TestInterface from "./Test3.vue"' instead?
    1 |
    2 | <script setup lang="ts">
  > 3 | import { TestInterface } from "./Test3.vue"
      |          ^^^^^^^^^^^^^
    4 |
    5 | const test: boolean = "true"
    6 | const props = defineProps<TestInterface>()

ERROR in src/Test2.vue:5:7
TS2322: Type 'string' is not assignable to type 'boolean'.
    3 | import { TestInterface } from "./Test3.vue"
    4 |
  > 5 | const test: boolean = "true"
      |       ^^^^
    6 | const props = defineProps<TestInterface>()
    7 | </script>
    8 |

ERROR in src/Test3.vue:12:7
TS2322: Type 'string' is not assignable to type 'boolean'.
    10 |
    11 | <script setup lang="ts">
  > 12 | const test: boolean = 'test'
       |       ^^^^
    13 | </script>
    14 |
    15 | <template>

webpack 5.64.1 compiled with 4 errors in 1569 ms
```

The `<script lang="ts">` block in `Test3.vue` clearly contains a typing error that is not caught. 

Note that I am not using it through `@vue/cli-service` but rather a custom webpack setup.

piotr-oles:
Is it correct to have `<script />` and `<script setup />` in the same `.vue` file? Currently, we support only single `<script />` tag

filimon-danopoulos-stratsys:
Yes it is valid, see https://v3.vuejs.org/api/sfc-script-setup.html#usage-alongside-normal-script

piotr-oles:
Thanks, I will try to support for this case :)

piotr-oles:
I pushed a new commit with support for `<script />` and `<script setup />` in the same file. Let me know if it works for you!

bodograumann:
Works nicely, thank you @piotr-oles. Hopefully we can get this into vue-cli soon.

filimon-danopoulos-stratsys:
Unfortunately it does not work 100%. The following does not typescheck `<script setup lang="ts">` properly again
```html
<script setup lang="ts">
const test1: boolean = 'test1'
</script>

<script lang="ts">
const test2: boolean = 'test2'
export interface TestInterface {
  prop1: string
}
</script>

<template>
  <div>
    {{ test }}
  </div>
</template>
```
 
 Result
 ```bash
 ERROR in src/Test2.vue:5:7
TS2322: Type 'string' is not assignable to type 'boolean'.
    3 | import { TestInterface } from "./Test3.vue"
    4 |
  > 5 | const test: boolean = "true"
      |       ^^^^
    6 | const props = defineProps<TestInterface>()
    7 | </script>
    8 |

ERROR in src/Test3.vue:7:7
TS2322: Type 'string' is not assignable to type 'boolean'.
     5 |
     6 | <script lang="ts">
  >  7 | const test2: boolean = 'test2'
       |       ^^^^^
     8 | export interface TestInterface {
     9 |   prop1: string
    10 | }

webpack 5.64.2 compiled with 3 errors in 1546 ms
 ```
 
 I would have expected to see an error for `test1` in the first `<script setup lang="ts">` .  I think the issue is found on [this line]( https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/676/commits/55a8525f1440942d6209ff48ec93b6c995e19e17#diff-cb661a5fe8782c8ac0ded6380496b24f307b563912a2f1e004ce5c03d325b8eeR104)
 I tried to debug it a little but I don't know the code enough to give very meaningful feedback. What I saw is that the lines are not empty but contain the string `//`. I logged the internal state to file and it looks like this: 
  ```json
{
  "script": [
    "//",
    "//",
    "//",
    "//",
    "//",
    "",
    "const test: boolean = 'test2'",
    "export interface TestInterface {",
    "  prop1: string",
    "}",
    ""
  ],
  "setup": [
    "//",
    "",
    "const test: boolean = 'test1'",
    ""
  ],
  "merge": [
    "//",
    "//",
    "//",
    "//",
    "//",
    null,
    "const test: boolean = 'test2'",
    "export interface TestInterface {",
    "  prop1: string",
    "}",
    null
  ]
}
 ```

 This means that you would never get the contents of the `<script setup>` block since the `<script>` block contains `//` and thus `scriptLines[line] || scriptSetupLines[line]` would always return the contents of `scriptLines[line]`. 
 
 Furthermore I am not certain that this approach will result in correct behavior. Type checking both these things together might produce type warnings that are not valid.  `<script setup>` in essence means that you omit some code that is automatically generated. 
 ```html
 <script setup>
   const test = "test"
 </script>
 ```
 Is in practice something like this (the actually generated code is different, this just illustrates the point)
 ```js
 import { defineComponent } from 'vue'
 export default defineComponent({
   setup() {
     const test = "test"
     return {
       test
     }
   }
 })
 ```
 Expanding that example with a regular `<script>` this code
 ```html
 <script>
 const test = "test"
 </script>
 <script setup>
   const test = "test"
 </script>
 ```
 Would generate something like this
 ```js
 import { defineComponent } from 'vue'
 const test = "test"
 export default defineComponent({
   setup() {
     const test = "test"
     return {
       test
     }
   }
 })
 ```
 Where the two constants both named `test` are defined in different closures and would not throw any type errors. However the merge strategy introduced here would generate (only considering the merge step)
```js
const test = "test"
const test = "test"
```
The two `test` constants are in the same block and this will and should result in `TS2451: Cannot redeclare block-scoped variable 'test'`.

Viggo95:
Has anyone encountered this error?

``` json
"vue": "2.6.11",
"vue-template-compiler": "2.6.11"
"@vue/composition-api": "1.2.4",
```

``` vue
<template>
  <div id="app">
    hello world
  </div>
</template>

<script setup lang="ts">
  const x: string = 42;
</script>

<style>
#app {
  font-family: Avenir, Helvetica, Arial, sans-serif;
}
</style>
```

**Error**
``` js
 File '/Users/xxx/cli5-test/src/App.vue.ts' is not a module.
    1 | import Vue from 'vue'
  > 2 | import App from './App.vue'
```

my  `shims-vue.d.ts ` 👇

``` ts
declare module '*.vue' {
  import Vue from 'vue'
  export default Vue
}
```

bodograumann:
@screetBloom You cannot use `<script setup>` with vue 2. It only works in vue 3.

filimon-danopoulos-stratsys:
I would suggest you use the provided API from @johnsoncodehk that he mentioned [here](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/668#issuecomment-962625500) as he seems to have solved these issues already.

haoqunjiang:
> @screetBloom You cannot use `<script setup>` with vue 2. It only works in vue 3.

Actually, you can, with https://github.com/antfu/unplugin-vue2-script-setup
But maybe not many people have been using it yet.

piotr-oles:
It's not that easy to integrate as it seems :) I will have to add source-map support to the plugin to properly map diagnostic positions.

johnsoncodehk:
> It's not that easy to integrate as it seems :) I will have to add source-map support to the plugin to properly map diagnostic positions.

If you need, I can expose volar internal source map api. It can transform between source range and generated range.

filimon-danopoulos-stratsys:
Sorry I don't think that I can be of much assistance in the implementation. I don't know the codebase well enough to provide meaningful feedback. Am happy to test it when there is something to look at again.

PR #677, opened 2021-11-23
Emit d.ts files when mode is write-tsbuildinfo

adriangodong:
This PR updates the write logic to also emit files ending with `d.ts` when mode is set to `write-tsbuildinfo`. This allows the plugin to work together with ts-loader or babel-loader and emits `d.ts` files without needing to run `tsc`.


piotr-oles:
Why don't you use `write-references` instead?

adriangodong:
> Why don't you use `write-references` instead?

Wouldn't that mode overwrite the output of ts-loader?

piotr-oles:
Ah, right. Could you add `write-dts` mode instead to keep it backward compatible?

adriangodong:
Closing this in favor of a new mode, write-dts, implemented in https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/679

Issue #658, opened 2021-09-28
Fork-ts-checker-webpack-plugin with TSC, How to avoid double type checking?


facinick:
I am using babel for transpilation,
fork-ts-checker-webpack-plugin for type checking,

However, I also want to generate .d.ts files. for which I am running tsc command, which again does type checking and i want to avoid that.

Is there a preferred combination of libraries for transpiling + typechecking + .d.ts generation?

PR #679, opened 2021-11-24
feat: implement write-dts mode

### LINKED ISSUE #663: How to emit .d.ts and .d.ts.map files?
I cannot use ts-loader to generate declaration files with this plugin as that'll cause double type checking.
How can I use this plugin and also have declaration files generated?

adriangodong:
This PR adds a new mode called `write-dts`. When selected, the plugin will emit tsbuildinfo, d.ts, and d.ts.map files. This allows the plugin to work together with` ts-loader` (with `transpileOnly` set to `true`) or `babel-loader` and emit type definition files.

Fixes #663, #658


piotr-oles (approved):
Code looks great, I will wait for the CI and if everything is green, we're good to merge :) Thanks for the contribution!

adriangodong:
@piotr-oles ping for merge 🙏

PierBJX:
@piotr-oles @onigoetz @adriangodong 

After investigation (look at e2e test), in order to make it work, you need to set the `build` typescript option to `true`.

See the working configuration:
```
module.exports = {
  entry: './foo.ts',
  module: {
    rules: [
      {
        test: /\.ts$/,
        exclude: /node_modules/,
        use: [
          {
            loader: 'ts-loader',
            options: {
              transpileOnly: true,
            },
          },
        ],
      },
    ],
  },
  plugins: [
    new ForkTsChecker({
      typescript: {
        build: true,
        mode: 'write-dts',
      },
    }),
  ],
};

```

I don't know if it was the expected behavior but with this configuration it works. Remove `build: true`, you will see it does not work anymore. Indeed, by default, the value is set to `false`.

PR #681, opened 2021-12-01
chore: fix README typo

BBosman:


piotr-oles:
thanks for the PR, I fixed it in #708

PR #707, opened 2022-02-04
feat: allow to use webpack infrastructure logger

piecyk:
I agree that logger options where overcomplicated, and good that it was simplified in #695 but still having an option to use webpack infrastructure logger would be nice, matching other plugins. 

Maybe we can meet in the middle, this PR just use already provided infrastructure-logger enabling it via `webpack-infrastructure` option, imho this should be the default 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ module.exports = {
 ## Modules resolution
 
 It's very important to be aware that **this plugin uses [TypeScript](https://github.com/Microsoft/TypeScript)'s, not
-[webpack](https://github.com/webpack/webpack)'s modules resolution**. It means that you have to setup `tsconfig.json` correctly. 
+[webpack](https://github.com/webpack/webpack)'s modules resolution**. It means that you have to setup `tsconfig.json` correctly.
 
 > It's because of the performance - with TypeScript's module resolution we don't have to wait for webpack to compile files.
 >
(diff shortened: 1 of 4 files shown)


piotr-oles:
You can configure webpack to show infrastructure logs for the plugin:
```
  infrastructureLogging: {
    level: 'log',
    debug: /ForkTsCheckerWebpackPlugin/,
  },
```
Does this solve your issue?

piecyk:
> You can configure webpack to show infrastructure logs for the plugin:

Yes, but i'm not interested in those logs when writing code.

> Does this solve your issue?

No, as it still will print via console, overall i want everything being logged via webpack infrastructure logger, making the output consistent.

piotr-oles:
So basically you want to have a `[ForkTsCheckerWebpackPlugin]` prefix for all logs, even those that report type-checking errors, right?

piecyk:
Yes, keeping the formatting consistent, but here it's not only about the `prefix` it will also queue the messages correctly, checkout the https://github.com/webpack/webpack/blob/main/lib/node/nodeConsole.js

piotr-oles (approved):
Code seems legit, I will test it manually and if everything is good, we can merge this

piecyk:
Thanks 🙌

PR #716, opened 2022-03-01
fix: bring back :line:col format in webpack errors

robcrocombe:
Hello! In 2020 I opened an issue (#481) and subsequent PR (#489) to make error file names clickable with their line location intact, when you use cmd+click in a terminal like iTerm2. E.g. before it was `filename.ts 12:5` but we changed it to `filename.ts:12:5` to make it one item without space delimiters. This is a huge win for productivity as you don't need to hunt down the line number after opening the file.

I noticed while upgrading to v7 that this behaviour has been reverted by #641. It was mentioned in the GH issue that developer experience was more important than following Webpack convention. Has the stance changed on this or was it just a mistake?

If so, I've made this PR to bring back the clickable location behaviour. If not, is there another way we can fix this? I noticed there was a custom `formatter` option but that does not allow you to format the file name, and I still want to use codeframe. Thanks

### DIFF
--- a/src/formatter/webpack-formatter.ts
+++ b/src/formatter/webpack-formatter.ts
@@ -17,7 +17,7 @@ function createWebpackFormatter(formatter: Formatter): Formatter {
     if (issue.file) {
       let location = chalk.whiteBright.bold(relativeToContext(issue.file, process.cwd()));
       if (issue.location) {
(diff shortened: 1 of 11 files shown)


piotr-oles (approved):
Thanks for the PR! These changes were removed accidentally :)

robcrocombe:
Thanks @piotr-oles!

PR #730, opened 2022-04-09
Fix `generateTrace` with newer TypeScript versions and `build: false`

### LINKED ISSUE #722: `generateTrace` does not work with newer TypeScript versions
## Current behavior

[The README describes how to profile slow type-checking by setting `generateTrace` in `tsconfig.json`](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#profiling-types-resolution). However this doesn't work with recent TypeScript versions.

## Expected behavior

When building with `"generateTrace": "./traces"` in `tsconfig.json`, trace files should be created in `./traces` as stated in the README.

1. Set up a project with a recent version of TypeScript (I'm using 4.6.2, but I'm nearly certain this problem exists as far back as 4.2.x).
2. Set  up a `tsconfig.json` with `"generateTrace": "{folderName}"` as [described in the README](](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#profiling-types-resolution)).
3. Do a build.
4. Observe that no trace files were created.

## Environment
 - **fork-ts-checker-webpack-plugin**: 6.2.12
 - **typescript**: 4.6.2
 - **eslint**: 7.32.0
 - **webpack**: 4.46.0
 - **os**: macOS 12.0.1

mkantor:
Closes #722 and also fixes another issue I noticed along the way: `generateTrace` did not work unless I was using `build: true` (I think that was introduced by #691).

Adds tests and updates documentation.

### DIFF
--- a/src/typescript/worker/lib/tracing.ts
+++ b/src/typescript/worker/lib/tracing.ts
@@ -4,22 +4,29 @@ import { getConfigFilePathFromCompilerOptions } from './config';
 import { typescript } from './typescript';
 import { config } from './worker-config';
 
-// write this type as it's available only starting from TypeScript 4.1.0
+// these types are internal in TypeScript, so reproduce them here
+type TracingMode = 'project' | 'build' | 'server';
 interface Tracing {
-  startTracing(configFilePath: string, traceDirPath: string, isBuildMode: boolean): void;
-  stopTracing(typeCatalog: unknown): void;
-  dumpLegend(): void;
+  startTracing?: (tracingMode: TracingMode, traceDir: string, configFilePath?: string) => void;
+
+  tracing?: {
+    stopTracing(): void;
+    dumpLegend(): void;
+  };
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const tracing: Tracing | undefined = (typescript as any).tracing;
+const traceableTypescript: Tracing = typescript as any;
 
 export function startTracingIfNeeded(compilerOptions: ts.CompilerOptions) {
-  if (compilerOptions.generateTrace && tracing) {
-    tracing.startTracing(
-      getConfigFilePathFromCompilerOptions(compilerOptions),
-      compilerOptions.generateTrace as string,
-      config.build
+  if (
+    typeof compilerOptions.generateTrace === 'string' &&
+    typeof traceableTypescript.startTracing === 'function'
+  ) {
+    traceableTypescript.startTracing(
+      config.build ? 'build' : 'project',
+      compilerOptions.generateTrace,
+      getConfigFilePathFromCompilerOptions(compilerOptions)
     );
   }
 }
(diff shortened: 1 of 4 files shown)


test/e2e/type-script-tracing.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import { extractWebpackErrors } from './driver/webpack-errors-extractor';
+
+describe('TypeScript Tracing', () => {
+  it.each([
+    { build: false, typescript: '~4.3.0' },
+    { build: true, typescript: '~4.3.0' },
+    { build: false, typescript: '~4.4.0' },
+    { build: true, typescript: '~4.4.0' },
+    { build: false, typescript: '~4.5.0' },
+    { build: true, typescript: '~4.5.0' },
```
piotr-oles: I appreciate that you add all these cases, but each case takes time so we have to keep some balance here. Otherwise, our test job will take hours :) I suggest keeping cases only for the min and max versions. I would assume that if it works for these versions, it will work for versions in-between :)

src/typescript/worker/lib/tracing.ts
```diff
@@ -4,37 +4,48 @@ import { getConfigFilePathFromCompilerOptions } from './config';
...
 import { config } from './worker-config';
 
-// write this type as it's available only starting from TypeScript 4.1.0
+// these types are internal in TypeScript, so reproduce them here
+type Mode = 'project' | 'build' | 'server';
 interface Tracing {
-  startTracing(configFilePath: string, traceDirPath: string, isBuildMode: boolean): void;
-  stopTracing(typeCatalog: unknown): void;
-  dumpLegend(): void;
+  startTracing:
```
piotr-oles: nit: I think that ‘?’ instead of ‘undefined’ would be easier to read :)

src/typescript/worker/lib/tracing.ts
```diff
@@ -4,37 +4,48 @@ import { getConfigFilePathFromCompilerOptions } from './config';
...
+    | undefined
+    | ((tracingMode: Mode, traceDir: string, configFilePath?: string) => void);
+
+  tracing:
```
piotr-oles: nit: same here :)

src/typescript/worker/lib/tracing.ts
```diff
@@ -4,37 +4,48 @@ import { getConfigFilePathFromCompilerOptions } from './config';
 import { typescript } from './typescript';
 import { config } from './worker-config';
 
-// write this type as it's available only starting from TypeScript 4.1.0
+// these types are internal in TypeScript, so reproduce them here
+type Mode = 'project' | 'build' | 'server';
```
piotr-oles: nit: could you name it ‘TracingMode’?

piotr-oles (commented):
Thanks for the PR! I left a few minor comments :)

mkantor:
@piotr-oles I addressed your feedback; thanks for the quick review. Let me know if you have any other suggestions.

piotr-oles (approved):
Thanks, looks good!

PR #753, opened 2022-06-22
feat: add 'ignored' property as option

pacop:
I've created this PR to facilitate the understanding of this [bug report](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/752). I'm not familiar enough with the source code, so this may have to be improved, but at least we have some starting point.

Let me know if I could help.


piotr-oles:
Thanks for the PR and bug report! I'm closing in favor of #757

Issue #765, opened 2022-07-19
Vue SFC TS check was broken on vue2.7


iceprosurface:
## Current behavior

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/main/src/typescript/extension/vue/type-script-vue-extension.ts#L113

vue2.7 hit `isVueTemplateCompilerV3`,but there is no `descriptor` on `parsed`

```typescript
const parsed = compiler.parse(vueSourceText);

if (parsed.descriptor && parsed.descriptor.script) {
  const scriptV3 = parsed.descriptor.script;
```
`parsed` shows like following codes
```
 {
  source: undefined,
  filename: 'anonymous.vue',
  template: null,
  script: null,
  scriptSetup: null,
  styles: [],
  customBlocks: [],
  cssVars: [],
  errors: [],
  shouldForceReload: [Function (anonymous)]
}

```

So that any ts check on vue will fail.

```
export { default as FilteredSearchToken } from './FilteredSearchToken.vue';
```

<img width="827" alt="image" src="https://user-images.githubusercontent.com/16058059/179655054-dcea7285-b0f4-4333-b8ec-47ca98ebcd3a.png">

Error shows:

TS2322: Type '{}' is not assignable to type 'VueConstructor<Vue<Record<string, any>, Record<string, any>, never, never, never, never, (event: string, ...args: any[]) => Vue<Record<string, any>, Record<string, any>, ... 4 more ..., ...>>>'.

## Expected behavior

Checker should work fine.

## Steps to reproduce the issue

```bash
yarn serve
```

## Issue reproduction repository

https://github.com/iceprosurface/vue2.7-tsx-error

## Environment
 - fork-ts-checker-webpack-plugin@7.2.12 & fork-ts-checker-webpack-plugin@6.5.2
 - typescript@4.7.4
 - eslint@8.20.0
 - webpack@5.69.1
 - osx 12.3.1

piotr-oles:
I'm dropping support for Vue.js. For the last few years, support for Vue.js has been broken, and the implementation of Vue.js integration was a hacky work-around. Unfortunately, TypeScript doesn't expose an API to do that properly, and while I created a https://github.com/microsoft/TypeScript/issues/38736 on the TypeScript repository, the TypeScript team has other priorities. As I won't have time to support this feature and am receiving justified bug reports, I believe it's better to make it clear that Vue.js is not supported by the plugin instead of pretending that it is.

PR #768, opened 2022-07-21
fix: vue2.7 SFC compiler.parse error for v7

iceprosurface:
see: #765


piotr-oles (approved):

piotr-oles:
could you rebase and force-push the branch - it seems like CI got stuck :/

iceprosurface:
Rebase and force-push was finished.

After using this solution for 2 weeks, i found this doesn't align with the errors we see in IDEs(especially in volar) or from vue-tsc.

piotr-oles:
I'm dropping support for Vue.js - there is no clean API in TypeScript to support it and I don't have time to maintain it as current workaround is hacky and buggy :/

PR #808, opened 2023-04-16
Clarify "No errors found." message

cben:
A really minor usability tweak, subjective, up to you if you like it.

In the context of webpack-dev-server, there are 2 async build processes, which may report success/failure:

- webpack bundle:
  - `webpack 5.72.0 compiled with 2 errors in 5492 ms`,
  - `webpack 5.72.0 compiled successfully in 55345 ms` are self-explaining.

There is _sometimes_ a cyan "Type-checking in progress..." between them, but not always.

- TypeScript checking, which may arrive later:
  - when typescript fails, it starts with non-specific red "ERROR in ./src/..." but next line start with TS error code e.g. `TS6133: `, so clear enough.
  - when typescript succeeds, the green "No errors found." message may arrive later, and doesn't explain it's specifically what kind of errors. :point_left: 

This PR tweaks the last case to say "No typescript errors found." for clarity.

Screenshot from 2023-10-29 15-42-45

### DIFF
--- a/src/hooks/tap-done-to-async-get-issues.ts
+++ b/src/hooks/tap-done-to-async-get-issues.ts
@@ -68,7 +68,7 @@ function tapDoneToAsyncGetIssues(
       // print stats of the compilation
       config.logger.log(statsFormatter(issues, stats));
     } else {
-      config.logger.log(chalk.green('No errors found.'));
+      config.logger.log(chalk.green('No typescript errors found.'));
     }
 
     // report issues to webpack-dev-server, if it's listening
--- a/test/e2e/driver/webpack-dev-server-driver.ts
+++ b/test/e2e/driver/webpack-dev-server-driver.ts
@@ -57,7 +57,7 @@ function createWebpackDevServerDriver(
     process.stdout.on('data', (data) => {
       const content = stripAnsi(data.toString());
 
-      if (async && content.includes('No errors found.')) {
+      if (async && content.includes('No typescript errors found.')) {
         noErrorsListener.resolve();
       }
 


piotr-oles:
Maybe it would make more sense to just say "No typescript errors found."?

cben:
> Maybe it would make more sense to just say "No typescript errors found."?

Done, PTAL.  

UPDATE: now also updated `waitForNoErrors` test helper.  needs approval to re-run CI?

piotr-oles (approved):

tedmx:
This is a good change, but I suggest clearly marking TS logs in case of errors as well.

For example, imagine having a few "usual" webpack linking errors and also a TS error:

```
( ... omitted long error output from webpack linker ... )
webpack 5.99.8 compiled with 2 errors and 6 warnings in 509 ms
ERROR in ./src/helloTS/Comp.tsx:11:5
TS2339: Property 'message' does not exist on type 'CompProps'.
     9 | const Comp = (props : CompProps) => {
    10 |   const {
  > 11 |     message,
       |     ^^^^^^^
    12 |     fontSize,
    13 |   } = props
    14 |

Found 1 error in 639 ms.
```

Because of this message, reader may think that there is one error in total, and will be wrong.

For this reason, I have configured the plugin in my project like this:
```
    typescript: {
      configFile: env.prod ? 'tsconfig.json' : 'tsconfig.dev.json',
    },
    logger: {
      ...console,
      log: message => { console.log('TS:', message) },
    },
  }),
```

That way, TS-related messages are always marked, and are less misleading.

PR #815, opened 2023-05-30
Update cosmiconfig package in package.json

muskankhedia:
We have encountered a component governance problem with the older versions of the yaml packages. These packages are transitively utilized by `fork-ts-checker-webpack-plugin` package.


piotr-oles:
Hi! You have to run `yarn install` to update lock file :)

PR #818, opened 2023-06-21
chore(package): update cosmiconfig

mkhraisha:
Similar to #815 except it updates yarn.lock.

cosmiconfig dropped dependency on the yarn npm package which has CVEs

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@babel/code-frame": "^7.16.7",
     "chalk": "^4.1.2",
     "chokidar": "^3.5.3",
-    "cosmiconfig": "^7.0.1",
+    "cosmiconfig": "^8.2.0",
     "deepmerge": "^4.2.2",
     "fs-extra": "^10.0.0",
     "memfs": "^3.4.1",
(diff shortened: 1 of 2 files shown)


mkhraisha:
@piotr-oles

LucianBuzzo:
This also fixes an issue related to the vulnerable v1 version of the `yaml` package - https://github.com/advisories/GHSA-f9xv-q969-pqx4
Unfortunately v8.0.0 of `cosmiconfig` dropped support for v12 of node, which is still supported by `fork-ts-checker-webpack-plugin` - see https://github.com/cosmiconfig/cosmiconfig/blob/main/CHANGELOG.md#800
This is a little bit of a headache, because you end up with CVE warnings for any installation of NestJS, since the `@nestjs/cli` package has a transitive dependency on `cosmiconfig` (via this package) and the older version of `cosmiconfig` has a dependency on the vulnerable `yaml` version!
I'm not sure on the best way to proceed, my personal preference would be for the maintainers to cut a new major version and drop support for node v12, since [security support for v12 ended over 1 year ago](https://endoflife.date/nodejs).

PR #823, opened 2023-09-07
chore(node): drop support for node v12

LucianBuzzo:
This change removes "official" support for node v12, which is now unmaintained and hasn't been supported since 30th April 2022. If users still need to use node v12 they can continue using the v8 version of this package.

BREAKING CHANGE: Drop support for Node.js v12

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [14, 16]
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
This would also mean that problems like the one described [here](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/818#issuecomment-1710157139) can be resolved.

LucianBuzzo:
@piotr-oles Could you take a look at this if you have a moment?

piotr-oles:
Thanks for the PR, sorry for the delay, I was on vacations 🌴

PR #825, opened 2023-10-04
fix: issue-webpack-error should not add colors to file string

sreucherand:
Introduced in #641, the issue location is now coloured in green in both the terminal console and in the error sent to webpack.

No problem at all in the terminal:

image

Having a look at the message sent through ws:

image

The file property contains the ANSI chalk chars corresponding to the colour applied to the issue location.

The thing is, in the webpack-dev-server overlay, the location does not come up properly:

image

In fact, the overlay does not seem to apply a ANSI to HTML conversion for the header of the message (file) (see. [overlay.js#L247](https://github.com/webpack/webpack-dev-server/blob/master/client-src/overlay.js#L247)) but only for the body of the message (message) (see. [overlay.js#L262](https://github.com/webpack/webpack-dev-server/blob/master/client-src/overlay.js#L262)).

I suggest that we just remove the coloured location in the error message sent to webpack.

Here is a minimum reproducible example:
https://github.com/sreucherand/webpack-ts-checker-minimal-reproducible-example

```sh
yarn
yarn start
```

### DIFF
--- a/src/issue/issue-webpack-error.ts
+++ b/src/issue/issue-webpack-error.ts
@@ -1,6 +1,5 @@
 import path from 'path';
 
-import chalk from 'chalk';
 import webpack from 'webpack';
 
 import type { FormatterPathType } from '../formatter';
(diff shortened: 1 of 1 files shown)


sreucherand:
Hi @piotr-oles, did you have the opportunity to look at this?

piotr-oles:
Hi! The change looks reasonable, let's wait for the CI to pass and we can merge it

PR #852, opened 2024-11-21
chore(deps): update chokidar, constrain engines

Fuzzyma:
This PR updates chokidar to v4 and reduces its dependency count to 1.

It also constrains the engine field in package.json to use node versions that are supported by the tests.
Unfortunately there is not a way to only constrain it for development (since I am sure this package also runs in newer node versions once bundled). So if you want me to remove the constraint let me know.

In that case I would add a note in the readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
   "dependencies": {
     "@babel/code-frame": "^7.16.7",
     "chalk": "^4.1.2",
-    "chokidar": "^3.5.3",
+    "chokidar": "^4.0.1",
     "cosmiconfig": "^8.2.0",
     "deepmerge": "^4.2.2",
     "fs-extra": "^10.0.0",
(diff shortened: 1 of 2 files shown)


Fuzzyma:
The CI fails because it cant install typescript because it tries to get the package from a url ending in ":". No idea where this is coming from tbh :D

piotr-oles:
Looks like we have broken E2E tests:
```
  TypeScript SolutionBuilder API
    ✕ reports semantic error for {"async": false, "mode": "readonly", "typescript": "~4.3.0"} (34358 ms)
    ✕ reports semantic error for {"async": true, "mode": "write-tsbuildinfo", "typescript": "~4.5.0"} (34252 ms)
    ✓ reports semantic error for {"async": false, "mode": "write-references", "typescript": "~4.7.0"} (12156 ms)
    ✕ reports semantic error for {"async": true, "mode": "write-dts", "typescript": "~4.3.0"} (34042 ms)
```
I will take a look at them later (I'm pretty busy these days, it might take a while)

piotr-oles:
Closing in favour of #858. Thanks @Fuzzyma for the PR!

Fuzzyma:
Awesome. Thank you!

